PR #2, opened 2017-03-20
easy test of lorenzwalthert

lorenzwalthert:
I changed four files instead of three since I added two tests in test-styler.R as well. Hope that is ok.

### DIFF
--- a/tests/testthat/test-styler.R
+++ b/tests/testthat/test-styler.R
@@ -20,3 +20,8 @@ test_that("removes space at EOL", {
 test_that("removes blank lines at EOF", {
   expect_equal(style_text(c("a() ", "", "")), "a()")
 })
+
+
+test_that("Space placed after 'if' and before '('", {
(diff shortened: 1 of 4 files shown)


tests/testthat/test-styler.R
```diff
@@ -20,3 +20,11 @@ test_that("removes space at EOL", {
 test_that("removes blank lines at EOF", {
   expect_equal(style_text(c("a() ", "", "")), "a()")
 })
+
+test_that("logical T and F are converted to TRUE and FALSE", {
+  expect_equal(style_text("T == F", c("TRUE = FALSE")))
```
krlmlr: Can we reliably detect that there is no symbol `T` or `F` defined in the scope?

lorenzwalthert: You are right, that actually might be rather tricky. Since this goes slightly beyond the scope of a simple test, I opened #3

krlmlr:
Thanks!

PR #4, opened 2017-03-24
lorenzwalthert_medium

lorenzwalthert:
In order to be consistent with `style_pkg()`, I used `withr::with_dir` to perform the styling. This, at the same time, means that the message printed (by `transform_lines_enc` internally) at the end of the styling to indicate which files were changed shows the paths relative to the *temporary* working directory used by `withr::with_dir` and not the directory that was the working directory before and after the execution of `style_src()`. If this is confusing, we have to do it without `withr::with_dir`, but then it would not be consistent anymore with `style_pkg()`, so I guess we can do it as proposed in this PR.

To increase the flexibility of the function regarding files in sub directories, `style_src()` has an argument `recursive` that will be passed to `dir` internally, so the user can choose whether to style files in just the selected directory or also in all sub-directories.


krlmlr:
Thanks. We should eventually add a test, but that will be part of getting close to full test coverage. Mocking can be done with `stubthat`, `mockery`, or `mockr`.

PR #5, opened 2017-03-24
lorenzwalthert_hard

lorenzwalthert:
By adding dependencies to the DESCRIPTION file and documenting exported functions via roxygen2, a few RCMD issues were fixed.

However, some problems prevailed and I was not sure how to best solve them.

- NOTE: No visible global function binding of 'token' and 'text' from `add_ws_to_parse_data`. Simple (but potentially uggly) solution: add the variables to `utils::globalVariables()` in styler.R, which was done for "."  already set. 
- ERROR: The two tests for our `in.R` example fail, since styler is not yet where we want it to be. Hence, I was wondering whether we should just turn the two tests into comments (and hence deactivate them for now).


krlmlr:
You can use `mutate_()` instead of `mutate()` to get rid of the "no visible ..." warning, this is what we do elsewhere. The next version of dplyr will provide better ways to program, but the `_()` versions should continue to work for some time.

krlmlr:
You can update the `out*.R` files, but perhaps keep a copy of the desired output formatting (which styler cannot do yet) somewhere in the repo.

lorenzwalthert:
Ok, I adapted your suggestions. I now got a perfect R CMD Check locally.

krlmlr:
Awesome, thanks!

lorenzwalthert:
Regarding the second part of the test [question](https://github.com/rstats-gsoc/gsoc2017/wiki/Noninvasive-source-code-formatting) : Good code quality is ensured in various ways:

* From a formal point, passing R CMD Checks with a high code coverage of tests (i.e. with codecov) are important. Also, one might want to use CI tools such as travis and AppVeyor to ensure the code works on other machines and operating systems as well. 
* Thorough code reviewing is key as well, because there can always be things the author might have missed and people with a fresh view on the code can spot. 

* Also, after having completed a task, one might realise that there is actually an easier or better way to  achieve the same result. Then, refactoring code makes sense. 

* Consistent programming style and documentation also adds to the code quality.

PR #8, opened 2017-03-27
vignette draft / status badge

lorenzwalthert:


vignettes/data_structures.Rmd
```diff
@@ -0,0 +1,122 @@
...
+serialize <- function(x) {
+  out <- Map(
+    function(terminal, token, leaf) {
+      if (terminal)
+        token
+      else
+        serialize(leaf)
+    },
+    x$terminal, x$token,
+    if (is.null(x$leaves)) rep(list(NULL), length(x$terminal)) else x$leaves
```
krlmlr: I think `x$leaves` cannot be NULL anymore.

vignettes/data_structures.Rmd
```diff
@@ -0,0 +1,122 @@
...
+the sub-expression is split up further - until we are left with tibbles that 
+only contain terminals.
+
+
+Recall the above example. `1+1` 
+is a sub-expression of the expression `code`, and we would like to set the 
+indention for `1+1`, even before we do anything with `1+1`.
+
+The code for a function that kind of sketches the idea and illustrates such a 
+recursion is given below:
```
krlmlr: What does that code do? What data do we need to implement serialization with this pattern?

vignettes/data_structures.Rmd
```diff
@@ -0,0 +1,122 @@
...
+```
+
+How to exactly implement a similar recursion to not just return the tokens, but 
+the styled code is subject to future work, so would be the functions to be
+applied to a sub-expression parse table that create correct indention.
+
+## Final Remarks
+
+Although a flat structure would possibly also allow to solve the problem, it is
+a less elegant and flexible solution to the problem, so the authors of this
```
krlmlr: Why?

vignettes/data_structures.Rmd
```diff
@@ -0,0 +1,122 @@
...
+Roughly speaking, a string containing code to be formatted is parsed with `parse`
+and the output is passed to `getParseData` in order to obtain a parse
+table with detailed information about every token. For a simple example string
+"`a <- function(x) { 1+1 }`" to be formatted, the parse table on which `styler`
+performs the manipulations looks similar to the one presented below.
+```{r, message = FALSE}
+library("styler")
+library("dplyr")
+
+code <- "a <- function(x) { 1+1 }"
```
krlmlr: Perhaps you could use a function that has another brace expression (perhaps with an `if` clause), and comment on the structure of both brace subexpressions in the nested parse trees?

lorenzwalthert: yes, I changed the example

vignettes/data_structures.Rmd
lorenzwalthert: Why not? I mean it will be named `x$child`, but what are the problems with it apart from that?

vignettes/data_structures.Rmd
lorenzwalthert: Not entirely sure. Partly because of duplication? And the inability of capturing expressions and sub-expressions?

vignettes/data_structures.Rmd
lorenzwalthert: I extended the description for it.

krlmlr:
Thanks!

vignettes/data_structures.Rmd
krlmlr: No problem, but the code can be made shorter if you just use `x$child` instead of this `if` expression.

PR #12, opened 2017-05-13
some documentation

lorenzwalthert:


lorenzwalthert:
@krlmlr sorry it seems like a dump question but can you clarify the meaning of ws in the project (and specifically add_ws_to_parse_data and the file called ws.R)? Is it just white space?

krlmlr:
ws are white space, yes. We can be more verbose here if you think it helps.

NAMESPACE
```diff
@@ -1,7 +1,9 @@
 # Generated by roxygen2: do not edit by hand
 
+export(add_ws_to_parse_data)
```
krlmlr: Why do we need to export these functions?

man/add_ws_to_parse_data.Rd
```diff
@@ -0,0 +1,28 @@
...
+add_ws_to_parse_data(parse_data)
+}
+\arguments{
+\item{parse_data}{a parse table.}
+}
+\value{
+a pre-processed parse table.
+}
+\description{
+Pre-processing parse data
```
krlmlr: The title is used as description if you omit the description in the documentation. The description can be more than one paragraph if you use the `@description` tag (I think...).

lorenzwalthert:
ok, yes we will see. It seems to me that it made sense in the beginning but since the code base grew, maybe the functions (and the .R files) now cover more than just white space, so the naming is maybe a bit confusing. Will look into that.

R/parsed.R
```diff
@@ -9,6 +9,18 @@ compute_parse_data_flat_with_ws <- function(text) {
...
 }
 
+#' Pre-processing parse data
+#'
+#' @details Preprocessing includes
+#' * removing non-terminal entries.
+#' * remove columns id, parent and terminal.
+#' * adding a start token.
+#' * add linebreak and space information.
+#' * replace spaces in comments at the end of the line.
```
krlmlr: in comments only?

lorenzwalthert: well it should probably read "remove". Should I change that? I thought it is in comments only since relevant line is #70
`mutate_(text = ~if_else(token == "COMMENT", gsub(" +$", "", text), text))`

krlmlr: Okay, that makes sense. We'll need to check if this is still necessary, because I remember code that removes all space at EOL (which isn't safe if a string constant contains an embedded newline, btw.).  For this PR the wording looks good.

NAMESPACE
lorenzwalthert: I usually use the combination 
`#' @export`
`#' @keywords internal `
and if I want to call the function during the development phase, I can save some keystrokes since I write `hidden_function()` instead of `package:::hidden_function()`. But I assume that's not best practice. Do you prefer to export only top-level functions right from the beginning then I guess?

krlmlr: Yes, we usually don't want to export everything because we don't want to take responsibility that these items remain unchanged for a "long time". During development you can use `devtools::load_all()` (Cmd+Shift+L in RStudio), it loads all internal functions too.

lorenzwalthert: ok, did not know that, thanks. I will update the PR then.

man/add_ws_to_parse_data.Rd
lorenzwalthert: So would you prefer to have the bullet points in the @description instead of @details?

krlmlr: I'm not sure, but I don't like an empty description section.

lorenzwalthert: I think the description section gets too long with the bullet points in it. I will add a sentence to @description and leave the bullet list in @description if that's ok.

krlmlr: Sure, happy to review. GitHub will notify me when you push.

lorenzwalthert: in general should I use forced pushes to update PRs if i "improve" the PRs but don't "extend" them?

krlmlr: I'd suggest we simply append commits and use regular push as a rule, and force-push only in special cases (e.g., cleanup of a long-living branch). This gives some free extra information about the evolution of the code.

lorenzwalthert: ok, sure

NAMESPACE
lorenzwalthert: what about the other internal functions that are exported now? Should I remove the export tag there as well?

lorenzwalthert: export(compute_parse_data_flat_with_ws)
export(compute_parse_data_nested)
export(get_transformers)
export(serialize_parse_data)

krlmlr: Were these used by the vignette? Does `R CMD check` complain about `:::` in the vignette?

lorenzwalthert: No, it seems to be fine. Should I remove them in this case and adapt the vignette?

krlmlr: Yes, we only export what we absolutely must export.

krlmlr:
One more thing about merging: (tbc)

krlmlr:
Everything that is in the commit description in a commit in the master branch is copied to NEWS during release of an intermediate/final version. Therefore, if you hit "Merge pull request", please enter what you'd like to end up in the NEWS file. Same if you commit directly to master: trivial commits only get a headline, larger ones can have a NEWS bullet in the DESCRIPTION. I'm always reviewing/editing the NEWS before I finish the release.

This doesn't affect commits you do in a branch (even in your fork's master), just the merge commits or other top-level commits to this repo's master.

lorenzwalthert:
Ok. So just to get it right: I create a merge commit (not rebase) and either use the short form (i.e. git commit x -m "message") with a commit message or a message and a description (with git commit x -m "message" -m "description" and the message and / or description gets dumped into NEWS which you edit prior to a release. Is that correct?

krlmlr:
Yes, merge commits please. Only the description ends up in the NEWS, not the message. If you merge in GitHub, the line you edit before merging becomes the description of the commit.

PR #15, opened 2017-06-01
more documentation

lorenzwalthert:
Documenting for:
* transform_files
* compute_parse_data_flat_with_ws
* verify_roundtrip
* seriealize_parse_data and rename it

Minor clarifications in documentations.


R/parsed.R
```diff
@@ -1,3 +1,15 @@
+#' Parse and pre-process character vector
+#'
+#' @description The function obtains detailed parse information for `text` via
+#'   [utils::getParseData] and does some minimal pre-processing by calling
```
krlmlr: I prefer `[utils::getParseData()]` for references to functions.

R/parsed.R
```diff
@@ -1,3 +1,15 @@
...
+#'   [add_ws_to_parse_data].
```
krlmlr: Same, also in a few other places.

R/parsed.R
```diff
@@ -1,3 +1,15 @@
...
+#' @inheritSection add_ws_to_parse_data Details
```
krlmlr: Would merging the documentation for the two functions work?

lorenzwalthert: would even merging the two functions make sense?

R/parsed.R
lorenzwalthert: ok, will update everywhere.

R/parsed.R
lorenzwalthert: ok, will update everywhere.

R/parsed.R
krlmlr: I'm not sure, I prefer smaller functions with dedicated responsibilities. Are the two functions opposites, or is one calling the other, or...?

lorenzwalthert: Well `compute_parse_data_flat_with_ws` is the complement of `serialize_parse_data_flat`. I thought about merging `add_ws_to_parse_data` with `compute_parse_data_flat_with_ws` since the latter calls the former (see line 13 to 18 in `parsed.R`. I would not merge the complements.

krlmlr: When in doubt, leave two separate functions, provided the name (which should uniquely describe the function's purpose) fits one line horizontally ;-) When splitting or refactoring functions, I'd suggest to think about what a good interface for unit-testing them would look like.

lorenzwalthert: I agree that small functions with well defined scope are preferable, but it can also add more overhead than benefit.

R/parsed.R
lorenzwalthert: Maybe it even make sense to extend the style guide with regard to `roxygen2` documentation. I opened an issue in the [tidyverse style](https://github.com/tidyverse/style/issues/21) repo.

R/parsed.R
lorenzwalthert: Let's leave it for now as it is.

PR #18, opened 2017-06-08
Documentation and renaming 

lorenzwalthert:
Documenting `compute_parse_data_nested` and `nest_parse_data`. Increase consistency in function and argument naming.


lorenzwalthert:
@krlmlr maybe you can have a look at the documentation added?

krlmlr:
Looks good. Generally, I'm fine with the naming as long as the names are clear and consistent and the tests pass. It helps if they are sufficiently distinct so that a full-text search doesn't have too many false positives.

lorenzwalthert:
Ok, good point about full-text search, was not aware of that. Is it now not distinct enough anymore? I mean we could for example change the name in R/rules.R back to just `pd` (or something that is not a substring of pd_flat / pd_nested, .e.g. `pd_to_style`), since these are the functions that actually *modify* the tables, the functions don't really have this property.

krlmlr:
I'm not sure we need to search for `pd_nested` or `pd_flat`, but perhaps for function names. The current naming looks fine, we can always improve later on.

lorenzwalthert:
Ok.

PR #21, opened 2017-06-12
Serialisation with indention

lorenzwalthert:
This PR contains an implementation of functions `indent_round` and `serialize_parse_data_nested`, so that 

* indention information for round brackets can be added to the nested parse table (`indent_round`)
* and this information can be used when serialising the parse table (`serialize_parse_data_nested`).

Further, a new utility function `newlines_and_spaces` is introduced to facilitate a common operation.
These functions are accompanied by tests for function calls, a vignette and further documentation.


R/nested.R
```diff
@@ -88,3 +88,49 @@ create_filler_nested <- function(pd_nested) {
...
+#'   [serialize_parse_data_nested_helper] spaces before the tokens have to be
+#'   removed.
+#' @param pd_nested A nested parse table with line break, spaces and indention
+#'   information.
+#' @return A character string.
+serialize_parse_data_nested <- function(pd_nested) {
+  raw <- serialize_parse_data_nested_helper(pd_nested, pass_indent = 0) %>%
+    unlist()
+  newline <- which(raw == "\n")
+  token <- setdiff(1:length(raw), union(which(raw == ""),
```
krlmlr: `seq_along()` is slightly safer. Why do we need this post-processing at all?

R/nested.R
```diff
@@ -88,3 +88,49 @@ create_filler_nested <- function(pd_nested) {
...
+#' @param pass_indent Level of indention of a token.
+#' @return A character vector with all tokens in `pd_nested` plus whitespace,
+#'   line break and indention information added between the tokens.
+#' @importFrom purrr pmap
+serialize_parse_data_nested_helper <- function(pd_nested, pass_indent) {
+  out <- pmap(list(pd_nested$terminal, pd_nested$text, pd_nested$child,
+             pd_nested$spaces, pd_nested$newlines, pd_nested$indent),
+           function(terminal, text, child, spaces, newlines, indent) {
+             if (terminal) {
+               c(rep_char(" ", pass_indent), text,
```
krlmlr: Can `text` contain newlines?

R/modify_pd.R
```diff
@@ -0,0 +1,43 @@
...
+#'   [create_filler_nested].
+#' @param indent_by How many spaces should be added after the token of interest.
+#' @name update_indention
+NULL
+
+#' @rdname update_indention
+indent_round <- function(pd, indent_by) {
+  start <- which(pd$token == "'('") + 1
+  stop <- which(pd$token == "')'") - 1
+  if (length(start) == 0 && length(stop) == 0) {
```
krlmlr: Please check logic.

R/modify_pd.R
```diff
@@ -0,0 +1,43 @@
+#' Update indention information of parse data
+#'
+#' @param pd A nested or flat parse table that is already enhanced with
+#'   line break and space information via [create_filler] or
```
krlmlr: `[create_filler()]` gives nicer markup.

R/modify_pd.R
```diff
@@ -0,0 +1,43 @@
...
+#'   [create_filler_nested].
+#' @param indent_by How many spaces should be added after the token of interest.
+#' @name update_indention
+NULL
+
+#' @rdname update_indention
+indent_round <- function(pd, indent_by) {
+  start <- which(pd$token == "'('") + 1
```
krlmlr: Are there cases where `start != 2` or `stop != nrow(pd)` ?

R/modify_pd.R
```diff
@@ -0,0 +1,43 @@
...
+  stop <- which(pd$token == "')'") - 1
+  if (length(start) == 0 && length(stop) == 0) {
+    pd$indent <- 0
+  } else {
+    pd$indent <- ifelse(1:nrow(pd) %in% start[1]:stop[1], indent_by, 0) *
+      lag(pd$newlines, default = 0)
+  }
+  # general, should maybe not go here.
```
krlmlr: Agreed, we could wrap this in a function that updates the `spaces` attribute.

tests/testthat/test-indetion_round_brackets.R
```diff
@@ -0,0 +1,65 @@
+context("Function calls with round brackets")
+
+test_that("one-line function call yields correct indention", {
+  code <- "a <- xyz(x, 22, if(x > 1) 33 else 4)"
+
+  back_and_forth <- code %>%
+    styler:::compute_parse_data_nested() %>%
```
krlmlr: You can omit `styler:::` .

R/modify_pd.R
lorenzwalthert: Well, do you mean `stop != nrow(pd) - 1`? 
Probably not in the case of round brackets. However, I wanted to keep the function rather general, since later on, we probably want to use a closure that returns functions like `indent_round`, but also one that indents after operators such as `%>%`.
`data_frame(x = 1, y = 2) %>%` 
__`  do_one_thing()` 

Here, we want `stop = nrow(pd)`, which is different from the round brackets, where probably always `stop != nrow(pd) -1 `. 
But anyways, I think you are right, in the end we might just use two different closures then, i.e. one for backets and one for other operators such as `%>%` `+` etc.

R/modify_pd.R
lorenzwalthert: I am not sure I understand what you mean. Probably it's enough to check for just one of them to be zero.

tests/testthat/test-indetion_round_brackets.R
lorenzwalthert: ok, will do so.

R/modify_pd.R
lorenzwalthert: ok, will do so.

R/modify_pd.R
lorenzwalthert: sorry for that one, should know it by now...

R/nested.R
lorenzwalthert: Ok. Well I could not find a way how to add indention to the first token on a line only. Also, it seems not a big performance deal to just remove the spaces immediately before every token that are not proceeded by a new line in the serialisation. I think it's safe but not very nice.

R/nested.R
lorenzwalthert: Yes, in general I think it can, but if this if clause applies, we are in a terminal token, for which text cannot be newlines ("\n") I think. It could be a string like "'\n'" though, but then it would remain a string and not get evaluated.

krlmlr: I was thinking about newlines embedded in strings, like this:

```r
"a
b"
```

But we just cannot apply indention on these without changing their semantics. So the point is moot.

R/nested.R
lorenzwalthert: I think generally every multi-line input for which at least one level of indention is required and there is more than one token per line. Since the indention level is currently added before every token all but the first token end up with too many spaces before them.
Here, I drop the lines 127 to 131 in nested.R, which are

```r
  newline <- which(raw == "\n")
  token <- setdiff(1:length(raw), union(which(raw == ""),
                                        union(grep("^ +$", raw), newline)))
  to_zero <- setdiff(token - 1, newline + 1)
  raw[to_zero] <- ""
```
Using [reprex](https://github.com/tidyverse/reprex) , we get the following example:
``` r
library(magrittr)
indented_multi_line_random <- c(
"            call(       ", 
"  1,", 
"  call2(               ", 
"    2, 3,", "call3(1, 2, 22),", 
"    5", 
"),",
"           144", 
")")
raw <- styler:::compute_parse_data_nested(indented_multi_line_random) %>%
  styler:::create_filler_nested() %>% 
  styler:::indent_round_nested() %>% 
  styler:::serialize_parse_data_nested_helper(pass_indent = 0) %>% 
  unlist()

out <- raw %>% 
  paste0(collapse = "") %>% 
  strsplit("\n", fixed = TRUE) %>% 
  .[[1L]]

out
#> [1] "call("                                             
#> [2] "  1,"                                              
#> [3] "  call2  ("                                        
#> [4] "    2  ,   3  ,"                                   
#> [5] "    call3    (    1    ,     2    ,     22    )  ,"
#> [6] "    5"                                             
#> [7] "  ),"                                              
#> [8] "  144"                                             
#> [9] ")"
cat(out, sep = "\n")
#> call(
#>   1,
#>   call2  (
#>     2  ,   3  ,
#>     call3    (    1    ,     2    ,     22    )  ,
#>     5
#>   ),
#>   144
#> )
```

krlmlr: I see a problem with the current data structure: The line break information (`newline` variable) is given as newline *after* a token, but perhaps the information that a newline appears *before* a token is more useful. Can you replace `newline` with a new variable `lag_newline = lag(newline)` on the flat parse data (before nesting), and work with that? I hope this will eliminate the need for postprocessing.

R/modify_pd.R
krlmlr: I'd like to use a different handler for other expression types, but we can think about it when we're done with the indention of function calls.

R/nested.R
lorenzwalthert: Ok, I tried that and it works. The question is how exactly it should be implemented. I see two possible approaches. 

* Use *lag_newlines* only (and drop *newlines*) and always inserting newlines and indention before a token, spaces after it.

* Use *lag_newlines* to determine whether there is indention before a token. Continue using newlines and spaces for inserting non-indention spacing after each token.

I think for now we can stick with the former. 
However, the latter might be preferable because it is more flexible in the sense that spacing information other than our computed indention can technically be stored. At the moment, we just set spacing after a token to zero when a line break comes after the token (see R/modify_pd.R, line 20). Maybe the flexibility is not needed later and it just complicates things, so I suggest to go with the former.

krlmlr: I agree that we should try resorting only on `lag_newlines`. Do we still need the postprocessing now?

R/modify_pd.R
krlmlr: Are you going to extract a function here?

R/nested.R
lorenzwalthert: No, it's not necessary anymore. In the next step, I can maybe even see if we need newlines at all, maybe we can get rid of it completely.

R/modify_pd.R
lorenzwalthert: yes, I will do that separately.

R/modify_pd.R
lorenzwalthert: Ok, resolved that differently anyways

lorenzwalthert:
No I have not yet pushed because things in this PR depended quite a bit on how to use `lag_newlines` and I was not sure which approach you prefer.  Now it's clear.

lorenzwalthert:
Regarding EOL stripping, I am not sure whether that is the best way to do it. I mean I looked at how we can integrate these internal functions with `make_transformer()` and I think we can use all (flat) functions from R/rules.R without the need of writing nested versions of them (see [this](https://github.com/lorenzwalthert/styler/commit/b4c59498bd9029b6509d79d09171c785fdb5fb36) commit in another branch). Hence, maybe the EOL stripping should just be one of them - or maybe not, if you don't want to think of EOL stripping as it as a rule, but it can be thought of being one.

R/nested.R
```diff
@@ -85,6 +85,50 @@ create_filler_nested <- function(pd_nested) {
...
+#' Collapses a nested parse table into its character vector representation.
+#' @param pd_nested A nested parse table with line break, spaces and indention
+#'   information.
+#' @return A character string.
+serialize_parse_data_nested <- function(pd_nested) {
+  raw <- serialize_parse_data_nested_helper(pd_nested, pass_indent = 0) %>%
+    unlist() %>%
+    paste0(collapse = "") %>%
+    strsplit("\n", fixed = TRUE) %>%
+    .[[1L]]
```
krlmlr: dplyr 0.7.0 has `pull()`, but we can address this in a separate PR

tests/testthat/test-indetion_round_brackets.R
```diff
@@ -0,0 +1,68 @@
...
+    indent_round_nested() %>%
+    strip_eol_spaces_nested() %>%
+    serialize_parse_data_nested()
+
+  expect_identical(code, back_and_forth)
+})
+
+##  ............................................................................
+
+indented_multi_line_correct <- c(
```
krlmlr: I wonder if this is easier in example files (like `in.R` and `out.R`). Perhaps you can extend to support an arbitrary number of `in.R` files? Again, separate PR.

R/modify_pd.R
```diff
@@ -0,0 +1,53 @@
...
+#' @param pd A nested or flat parse table that is already enhanced with
+#'   line break and space information via [create_filler()] or
+#'   [create_filler_nested()].
+#' @param indent_by How many spaces should be added after the token of interest.
+#' @name update_indention
+NULL
+
+#' @rdname update_indention
+indent_round <- function(pd, indent_by) {
+  if (any(pd$token == "')'")) {
```
krlmlr: Can we look only at `pd$token[[2]]` or `pd$token[[nrow(pd)]]`?

R/modify_pd.R
```diff
@@ -0,0 +1,53 @@
...
+    start <- 2
+    stop <- nrow(pd) - 1
+  } else {
+    start <- stop <- 0
+  }
+  pd$indent <- ifelse(seq_len(nrow(pd)) %in% start:stop, indent_by, 0)
```
krlmlr: This might look clearer if we move this assignment to the `if` branch, and remove the `else` branch.

R/modify_pd.R
```diff
@@ -0,0 +1,53 @@
...
+#'   line break and space information via [create_filler()] or
+#'   [create_filler_nested()].
+#' @param indent_by How many spaces should be added after the token of interest.
+#' @name update_indention
+NULL
+
+#' @rdname update_indention
+indent_round <- function(pd, indent_by) {
+  if (any(pd$token == "')'")) {
+    start <- 2
```
krlmlr: Please check boundaries: Do we really need to indent the second token, or perhaps only starting from the third?

R/modify_pd.R
```diff
@@ -0,0 +1,53 @@
...
+#' @param pd A nested parse table that is already enhanced with
+#'   line break and space information via [create_filler_nested].
+#' @name update_indention_nested
+NULL
+
+#' @rdname update_indention_nested
+indent_round_nested <- function(pd) {
+  if (is.null(pd)) return(pd)
+  pd <- indent_round(pd, indent_by = 2)
+  pd$child <- map(pd$child, indent_round_nested)
```
krlmlr: I'd prefer `mutate()` over sub-assignment here and in the functions below and above. This should work well with dplyr 0.7.0. Separate PR?

R/modify_pd.R
lorenzwalthert: Not sure what you @krlmlr mean. We need to assign a value to `pd$indent` anyways, since this column is used in the recursion of `serialize_parse_data_nested_helper`. Otherwise we have to change this function so it can handle the case for which the column does not exist, which I think is not a good idea.

R/modify_pd.R
lorenzwalthert: see above, I think we need to explicitly find the start token because of the two scenarios.

R/modify_pd.R
lorenzwalthert: I don't think so because we have some tibbles with only one row. Looking at the last token is probably also not a good idea because we have two cases
`mycall(1, 2)` and ` (1 + x)`
Which have their opening brace at the second and the first position respectively. I think that's really the reason why I implemented the approach with `which` in the first place: To get the start right.
I think this function should really handle both cases and it's easier to implement it in one function instead of two. Hence, I propose to go back to the `which` approach, at least for the start. I think the end is unambiguous.

R/modify_pd.R
krlmlr: Oh, I see. Let's leave it for now.

R/modify_pd.R
krlmlr: I wasn't aware that this also covers parens when they are used to group arithmetic expressions. Would you mind adding a test?

R/modify_pd.R
lorenzwalthert: Ok. In a pipe like this?
```r
pd <- pd %>%
  indent_round(pd, indent_by = 2) %>%
  mutate(child = map(child, indent_round_nested)
```

krlmlr: Yes. What exactly do you want to add to the style guide?

lorenzwalthert: Avoiding sub assignment if possible. See [here](https://github.com/tidyverse/style/issues/25)

Issue #23, opened 2017-06-14
refine testing


lorenzwalthert:
Consider using 
* files like `in.R` and `out.R`
* a function that runs through all files (to avoid code duplication)

This issue originates in #21

lorenzwalthert:
Closed by #28

PR #28, opened 2017-06-14
Add tools for scalable testing

lorenzwalthert:
the testing is based on the following concept:
* There are *-in.R files that should be transformed and
  then compared against *-out.R files
* There can be multiple *-in.R files that should be matched
  against one *-out.R file
* The main function test_collection takes (1) a directory
  and (2) optionally a regular expression to find *-in.R and
  *-out.R files, (3) a transformer function that will be
  applied to *-in.R files before comparing with the reference
  specified in *-out.R


R/serialized_tests.R
```diff
@@ -0,0 +1,116 @@
...
+#' @inheritParams transform_and_check
+#' @importFrom purrr flatten_chr pwalk
+test_collection <- function(test, sub_test = NULL,
+                            write_back = FALSE,
+                            transformer) {
+
+  path <- if_else(interactive(),
+                  file.path("tests", "testthat", test),
+                  file.path(test))
+
```
lorenzwalthert: @krlmlr This path creation is suboptimal, I know, but unfortunately, `here` did not work here, maybe because during R CMD check, the test folder is placed somewhere else? `devtools::check()` succeeded, but `R CMD check` not. Maybe you know a better way. Rather than `interactive()`, it should be soemthing like `r_cmd_check_is_running()` but I could not find such a function. Maybe we can make it dependent on some global variables `R CMD check` sets, but I have tool little knowledge about that.

R/serialized_tests.R
```diff
@@ -0,0 +1,116 @@
...
+    strip_eol_spaces_nested() %>%
+    serialize_parse_data_nested()
+}
+
+#' Pass input right through
+#'
+#' Passes an input as is without changing its class (unlike [base::I()].
+#'   Helpful for testing in conjunction with [utf8::transform_lines_enc()].
+#' @param x The an R object
+as_is <- function(x) x
```
krlmlr: Can you use `identity()`?

krlmlr: Can you use `identity()`?

R/serialized_tests.R
krlmlr: Can you use  `rprojroot::is_testthat()`?

R/serialized_tests.R
lorenzwalthert: sure.

R/serialized_tests.R
lorenzwalthert: good point. Did so.

lorenzwalthert:
Closes #28.

lorenzwalthert:
Closes #23

krlmlr:
Good to know that a post-merge comment like this leads to closing the corresponding issue.

lorenzwalthert:
no it does not actually.

krlmlr:
I tend to add a comment "Closed by #..." or "Closed by deadbeef" to the corresponding issue when closing.

PR #34, opened 2017-06-19
Transform nested

lorenzwalthert:
## Content of PR
* `get_transformers()` calls either `get_transformers_flat()` or `get_transformers_nested()`. We use `purrr::partial()` so the user can pre-fill arguments such as `indent_by` of `indent_round()` when  `get_transformers_nested()` is called.
* introducing the visitor concept: Not every transformation such as `create_filler()` or `indent_round()` is carried out going through the whole hierarchy. Rather, the visitor applies all transformer functions on one level before going one level deeper and applying them there.
* Adding top-level argument `flat` to `style_text()` et al.
* adapting initial tests with in.R/ out.R to use `flat = TRUE` so they all pass.

## Questions

Currently, the user can use `flat` at the top-level functions to use either the flat or the nested styling approach and `get_transformers()` will have `flat = flat` by default. If the user wants to use a non-default value in `get_transformers()`, he or she needs to specify `flat` twice, i.e. as in tests/testthat/test-styler#17, 
```r
style_text("file_in.R")),
    flat = TRUE,
    transformers = get_transformers(flat = TRUE, strict = FALSE))
```
This is the case since we must know 

* which functions are the transformer functions to be used. We need `create_filler` et al. in addition to the rules in R/rules.R for `flat = FALSE`.
* which procedure should be applied when doing the transformations (nested or flat).

I don't see a better way to do that at the moment. I think we can leave it for now since eventually, the flat argument will be obsolete anyways or at least people will just use the default `flat = FALSE` mostly. Should we leave it like that for now @krlmlr?

## To do's (regarding transformations):
* use visitor concept for tests also.


lorenzwalthert:
The transformers in R/rules.R work yes. For the nested case, we need all transformers that we need for the flat case plus `create_filler` and all the indention function. You can see that best [here](https://github.com/krlmlr/styler/pull/34/files#diff-198bef27b7f4090009710dd779baaaf4R49)

krlmlr:
I missed that. Why would we need user-visible `flat` arguments then? I'd be happy moving over to nested handling entirely.

lorenzwalthert:
Just in case you  want to  use flat styling, mainly as a bench mark. Flat styling does not use the visitor approach, so both the transformers as well as the function make_transformer() are different (as described in the description of the PR). We have not yet I implemented a verify_roundtrip() and START token, styling multiple expressions ect. So I just felt we can leave a possibility to call the "old" flat styler conveniently from a function like style_text().

Issue #29, opened 2017-06-15
Support indenting more patterns


krlmlr:
- [x] braces
- [x] function definitions
- [x] brackets
- [x] line break after operator, incl. pipe
- [x] combined parentheses and braces
- [x] ...

lorenzwalthert:
This is closed by #71

PR #35, opened 2017-06-19
support more indentions

lorenzwalthert:
This PR is based on #29 to support more indention patterns.
The indention based on operators such as `+`, `-` etc. is different from the indention caused by brackets like `(`. In the latter case, the whole sub-expression has to be indented (that means everything until the closing brace occurs), whereas for the former, only the next token should be indented (if it is on a new line). Otherwise, the outcome looks like this:
```r
1 + 
  4 /
    5
```
But we want this:
```r
1 + 
  4 /
  5
```
Hence, I added a new column `spaces_before` that can take that into account. Not sure whether it can be solved differently (and more elegantly). The thing is that we need to strip end-of-line spaces since we build indention from scratch without taking into consideration the space between two tokens if they are separated by a line break. Therefore I think it is a reasonable solution.
I can rename `spaces` to `spaces_after` if you @krlmlr approve of this approach.


tests/testthat/indention_operators/operators_and_round_and_sign-out.R
```diff
@@ -0,0 +1,5 @@
+1 + (
+  + 2 /
```
lorenzwalthert: I think so, see #38.

tests/testthat/test-indention_operators.R
```diff
@@ -0,0 +1,28 @@
+# library("testthat")
```
lorenzwalthert: Ok. You mean instead of executing the file?

krlmlr: Yes. One of those many paths to Rome...

lorenzwalthert:
This PR is out-dated and a more elegant solution (namely to take operators of the same level of precedence to a higher level) is implemented (and merged into master) with #69.

PR #37, opened 2017-06-19
Parse multiple expressions

lorenzwalthert:


R/nested.R
```diff
@@ -46,13 +49,17 @@ compute_parse_data_nested <- function(text) {
...
+#' @param top_level_expr The ids of the expressions that cannot be attributed
+#'   to another expression since they are not nested in any other expression.
+#'   This is important for terminating the recursion, i.e. further nesting is
+#'   impossible if our nested parse table has `length(top_level_expr)` rows.
 #' @seealso [compute_parse_data_nested()]
 #' @return A nested parse table.
-nest_parse_data <- function(pd_flat) {
-  if (nrow(pd_flat) <= 1) return(pd_flat)
+nest_parse_data <- function(pd_flat, top_level_expr) {
+  if (nrow(pd_flat) <= length(top_level_expr)) return(pd_flat)
```
krlmlr: Could we avoid the new argument by using `all(parent == 0)` as condition for termination?

lorenzwalthert:
Travis CI on macOs did not work because `dyplr::pull()` was [not found](https://travis-ci.org/krlmlr/styler/jobs/244542483). Probably the latest version of dplyr was not installed. Has it to do with the fact that we use image xcode7.3 only for the macos builds?

krlmlr:
Never mind, I think this will vanish when dplyr 0.7.1 is out.

R/nested.R
lorenzwalthert: I think we cant' directly because the stopping condition is only half of the story. The other reason we need the ids of the top_level expressions is because we nest `pd_flat` into `internal` and `child`. 
```r
  split <-
    pd_flat %>%
    mutate_(internal = ~id %in% c(parent, top_level_expr)) %>%
    nest_("data", names(pd_flat))

```
However, we could get rid of it by also changing `mutate_()`
```r
nest_parse_data <- function(pd_flat) {
  if (all(pd_flat$parent == 0)) return(pd_flat)
  split <-
    pd_flat %>%
    mutate_(internal = ~id %in% parent | parent == 0) %>%
    nest_("data", names(pd_flat))
```
Should I do that?

krlmlr: Yes please, this looks simpler to me than introducing a new argument to the recursive function. Be careful with operator precedence (`%in%` vs. `|`).

lorenzwalthert: ok, should I use 
```r
mutate_(internal = ~ (id %in% parent) | (parent == 0)) %>%
```
?

krlmlr: Yes, this looks safer to me.

krlmlr:
dplyr 0.7.0 can't be built for recent OS X versions; cleared the Travis cache to be sure.

R/nested.R
```diff
@@ -27,6 +27,9 @@
 compute_parse_data_nested <- function(text) {
   parsed <- parse(text = text, keep.source = TRUE)
   parse_data <- tbl_df(utils::getParseData(parsed, includeText = TRUE))
+  top_level_expr <- parse_data %>%
```
krlmlr: Do we still need this?

lorenzwalthert: no, forgot to remove that.

lorenzwalthert:
Make nested approach wok on multiple expressions too.

PR #39, opened 2017-06-20
Refine testing

lorenzwalthert:
In general, we should probably use the internal functions for testing parts of styler instead of always applying all styling via the top-level functions. Hence, we this PR, we adapt the style_indent function that only does indention styling to use the visitor approach.


R/serialized_tests.R
```diff
@@ -99,8 +99,9 @@ transform_and_check <- function(in_item, out_item,
 style_indent_round <- function(text) {
   text %>%
     compute_parse_data_nested() %>%
-    create_filler_nested() %>%
-    indent_round_nested() %>%
-    strip_eol_spaces_nested() %>%
```
krlmlr: Do we need to remove these functions?

lorenzwalthert: oh yes sure, we don't need them anymore.

PR #40, opened 2017-06-20
Fix broken build due to *_nested dependencies

lorenzwalthert:
Actually I did not properly take into account the dependencies linked to *_nested() functions in #39 so the build broke. Hence, here comes the update. Since we are now using the visitor approach, I also updated the vignette.

@krlmlr Should we use `devtools::load_all()` at the beginning of the vignette? It seems like with the visitor concept, using `styler:::` everywhere adds too much noise everywhere.


lorenzwalthert:
`pkgload::load_all()` seems to work yes.

lorenzwalthert:
travis builds seem to work, but there is something [missing](https://ci.appveyor.com/project/lorenzwalthert/styler-bp2vm/build/1.0.75) related to pkgload to get a successful built with appveyor. I could not quite figure out what is missing in the appveyor file. Maybe you @krlmlr can help?

DESCRIPTION
```diff
@@ -12,15 +12,17 @@ Imports:
     tibble,
     tidyr,
     purrr,
-    magrittr
+    magrittr,
+    pkgload
```
krlmlr: I prefer alphabetic ordering of dependencies. Don't need to change anything now, just keep in mind for the next time we edit here.

krlmlr:
I registered styler with my AppVeyor account, we should change the README badge at some point. Rtools seem to be installed, running a build to see if the error is reproducible.

PR #42, opened 2017-06-20
Style comments correctly

lorenzwalthert:
See description of  b8f67a9 for a short explanation


R/modify_pd.R
```diff
@@ -16,7 +16,7 @@ indent_round <- function(pd, indent_by) {
     start <- stop <- 0
   }
   pd <- pd %>%
-    mutate(indent = ifelse(seq_len(nrow(pd)) %in% start:stop, indent_by, 0)) %>%
+    mutate(indent = indent + ifelse(seq_len(nrow(pd)) %in% start:stop, indent_by, 0)) %>%
```
krlmlr: Why was it necessary to switch from relative to absolute indent?

R/nested.R
```diff
@@ -24,12 +24,13 @@
...
 #'     - Function definitions
 #' - Remove `includeText = TRUE`
+#' @importFrom purrr rerun
 compute_parse_data_nested <- function(text) {
   parsed <- parse(text = text, keep.source = TRUE)
   parse_data <- tbl_df(utils::getParseData(parsed, includeText = TRUE))
   pd_nested <-
     parse_data %>%
-    mutate_(child = ~rep(list(NULL), length(text))) %>%
+    mutate_(child = ~rerun(length(text), tibble())) %>%
```
krlmlr: An empty tibble will cost us up to 700 bytes per terminal, whereas `NULL` costs nothing extra. I'm not sure how much this is relative to the object sizes we're already handling, but we should be aware of this.

Could we perhaps use an adapted version of `bind_rows()` that returns `NULL` iff the result has zero rows?

tests/testthat/parse_comments/mixed-in.R
```diff
@@ -0,0 +1,7 @@
+# A file with text
```
krlmlr: Can you include something we can indent already, e.g. a function call or an arithmetic expression with inline comments?

R/modify_pd.R
lorenzwalthert: I think it's more coherent to initialise in `create_filler()` (as we also do with spacing and line break information. otherwise, you need a function like `indent_round()` to initialise. Say you don't use `indent_round()` for some reason, you will get an error in `serialize_parse_data_nested_helper()` since this column is required.

tests/testthat/parse_comments/mixed-in.R
lorenzwalthert: Sure.

R/modify_pd.R
krlmlr: At some point I thought it was important to work with relative indents (local to the current sub-tibble), and compose the final indent only during deserialzation. We can initializa `indent` with zero, but do we really need to work with absolute indent values?

lorenzwalthert: Oh I think it's still relative in that sense, i.e. the values in indention are not propagated trough the nested structure.
You can see in the following example.


``` r
test <- c("# a comment", "call(", "  1,", "  call2(", "    2, 3,", "    call3(1, 2, 22),", 
  "    5", "  ), #' A roxygen comment", "  144", "  # anotehr indented comment", 
  ")")
cat(test, sep = "\n")
#> # a comment
#> call(
#>   1,
#>   call2(
#>     2, 3,
#>     call3(1, 2, 22),
#>     5
#>   ), #' A roxygen comment
#>   144
#>   # anotehr indented comment
#> )
library("magrittr")
nested <- styler:::compute_parse_data_nested(test) %>% styler:::visit(c(styler:::create_filler, 
  purrr::partial(styler:::indent_round, indent_by = 2)))
```
`call2()` (and everything within that expression) has a relative indention of two. 

```r
nested$child[[2]]
#> # A tibble: 10 x 15
#>    indent newlines short line1  col1 line2  col2    id parent   token
#>     <dbl>    <int> <chr> <int> <int> <int> <int> <int>  <int>   <chr>
#>  1      0        0  call     2     1     2     4     6     72    expr
#>  2      0        1     (     2     5     2     5     5     72     '('
#>  3      2        0     1     3     3     3     3     9     72    expr
#>  4      2        1     ,     3     4     3     4    10     72     ','
#>  5      2        0 call2     4     3     8     3    58     72    expr
#>  6      2        0     ,     8     4     8     4    59     72     ','
#>  7      2        1 #' A      8     6     8    25    62     72 COMMENT
#>  8      2        1   144     9     3     9     5    65     72    expr
#>  9      2        1 # ano    10     3    10    28    67     72 COMMENT
#> 10      0        0     )    11     1    11     1    69     72     ')'
#> # ... with 5 more variables: terminal <lgl>, child <list>, text <chr>,
#> #   lag_newlines <int>, spaces <int>
```
If we "zoom in" one more level, you can see that `call2()` now has no indention anymore, but `call3()` does have it again.
```r
nested$child[[2]]$child[[5]]
#> # A tibble: 10 x 15
#>    indent newlines short line1  col1 line2  col2    id parent token
#>     <dbl>    <int> <chr> <int> <int> <int> <int> <int>  <int> <chr>
#>  1      0        0 call2     4     3     4     7    16     58  expr
#>  2      0        1     (     4     8     4     8    15     58   '('
#>  3      2        0     2     5     5     5     5    19     58  expr
#>  4      2        0     ,     5     6     5     6    20     58   ','
#>  5      2        0     3     5     8     5     8    24     58  expr
#>  6      2        1     ,     5     9     5     9    25     58   ','
#>  7      2        0 call3     6     5     6    19    47     58  expr
#>  8      2        1     ,     6    20     6    20    48     58   ','
#>  9      2        1     5     7     5     7     5    53     58  expr
#> 10      0        0     )     8     3     8     3    55     58   ')'
#> # ... with 5 more variables: terminal <lgl>, child <list>, text <chr>,
#> #   lag_newlines <int>, spaces <int>
```

I just wanted to initialise earlier, so the column is there. Also, other functions are gonna affect indention (such as `indent_curly_brackets`) so we need to add up indention anyways later.

R/nested.R
lorenzwalthert: Ok. Yeah I was thinking about modifying `bind_rows()` too. I just thought it might add extra complication for little gain. 
Anyways, maybe we can even prevent the tibble from being created.
```r
bind_not_null <- function(...) {
  if (is.null(unlist(...))) return(NULL)
  bind_rows(...)
}
```
I can branch out and adapt the solution presented in this PR to match the one outlined above and benchmark the two in terms of speed (which is your ultimate concern I guess, not memory itself) so we can make an informed decision if that's ok.

krlmlr: The `unlist(...)` call looks expensive. Why not just check `nrow()` for the tibble returned from `bind_rows()`?

A benchmark seems low priority to me. I think we'll be much faster when we start using `getParseData(includeText = NA)` again. But we could compare the code for both solutions.

lorenzwalthert: Ok, I will check via the rows then.

R/utils.R
```diff
@@ -27,3 +27,18 @@ add_newlines <- function(n) {
 add_spaces <- function(n) {
   rep_char(" ", n)
 }
+
+#' Combine child and internal child
```
krlmlr: This looks like a function with a very narrow use case. As such I'd prefer if it was located right below the function that calls it.

lorenzwalthert: Ok.

PR #32, opened 2017-06-17
update readme

lorenzwalthert:
Adding a few words on what styler is plus links and defer large example further down in the README. Is that what you had im mind @krlmlr? Not sure how to update pkgdown now though.

PR #43, opened 2017-06-21
make readme work again with top-level api

lorenzwalthert:
just updated the function calls. It's a bit annoying to have the flat argument twice, as stated in #34, but we will change that later. Can you update the badge? I could not do it since I can't access settings of your appveyor and I was not able to reconstruct the url


lorenzwalthert:
I that would make sense and we could instead do a few small examples with narrow scope (i.e. whereas one example features just one rule). Probably this would make the README more catchy - and make it easier to get the reader's full attention. When reviewing the readme in #32 I was not really sure whether you'd prefer to keep the example as is.

krlmlr:
Doesn't work = doesn't indent properly (or even keep indention). Agree to use smaller examples.

I tend to look at changes mostly when reviewing, at that time it seemed ok to leave "as is".

lorenzwalthert:
Ok, so should we do that as part of this PR? Then I would adapt the title

krlmlr:
I'd like to leave the README in a nice consistent state. We can remove the examples in this PR and think about a nicer README later, or add a few smaller (working) examples right away.

lorenzwalthert:
Any news @krlmlr regarding the failing builds of appveyor? I added #46 to describe the problem.

krlmlr:
Could you please update the AppVeyor link to https://ci.appveyor.com/project/krlmlr/styler, and the badge too?

README.Rmd
```diff
@@ -18,62 +18,40 @@ knitr::opts_chunk$set(
...
 
 The goal of styler is to provide non-invasive pretty-printing of R source code 
 while adhering to the [tidyverse](https://github.com/tidyverse/style) formatting 
 rules. Support for custom style guides is planned.
 
-You can style a simple character vector of code with `style_text()`, a directory 
-with `style_dir()`or a whole R package with `style_pkg()`.
+You can style a simple character vector of code with `style_text()`:
 ```{r echo=FALSE, message=FALSE}
 devtools::load_all()
```
krlmlr: Did we mean to use pkgload here (and add it to Suggests)?

README.Rmd
```diff
@@ -1,7 +1,7 @@
 ---
 output:
   github_document:
-    html_preview: false
+    html_preview: true
```
krlmlr: `html_preview` doesn't work on my machine :-(

lorenzwalthert: Ok, will remove that.

lorenzwalthert:
Did you see that I added a GIF that demonstrates styler with the RStudio Add-in? Was not sure whether you wanted a 2 MB file in the repo, that is why I pushed it to some other repo where I store larger files and made a link to it. What's the best way of dealing with that?

lorenzwalthert:
@krlmlr I can't update the badge since the url involves some random has. I think only you as the owner of the repo can access via the settings in appveyor.com. Can you please change it yourself? As of now, the build image is "passing" although when clicking on the link, it is clearly not.

PR #48, opened 2017-06-23
Outsource tokenise and spacing

lorenzwalthert:


R/nested.R
```diff
@@ -37,6 +36,18 @@ compute_parse_data_nested <- function(text) {
   pd_nested
 }
 
+#' Obtain token from text
```
krlmlr: token table?

R/nested.R
```diff
@@ -37,6 +36,18 @@ compute_parse_data_nested <- function(text) {
...
+#'
+#' [utils::getParseData()] is used to obtain a flat parse table from `text`.
+#' @param text A character vector.
+#' @return A flat parse table
+tokenize <- function(text) {
+  parsed <- parse(text = text, keep.source = TRUE)
+  parse_data <- tbl_df(utils::getParseData(parsed, includeText = TRUE))
```
krlmlr: I think `as_tibble()` is preferred these days. Do we still need `includeText = TRUE`?

R/nested.R
```diff
@@ -37,6 +36,18 @@ compute_parse_data_nested <- function(text) {
   pd_nested
 }
 
+#' Obtain token from text
+#'
+#' [utils::getParseData()] is used to obtain a flat parse table from `text`.
+#' @param text A character vector.
+#' @return A flat parse table
+tokenize <- function(text) {
```
krlmlr: Do we have another instance of `getParseData()` that can be replaced with a `tokenize()` call?

lorenzwalthert: oh yeah for the flat calculations. Replaced that.

lorenzwalthert: Should I move it to R/utils.R then or should it stay where it is?

krlmlr: It's fine where it is now.

R/nested.R
lorenzwalthert: ok

R/nested.R
lorenzwalthert: ok.

R/nested.R
lorenzwalthert: technically no, but `includeText = TRUE` is sometimes convenient so you can see everything inside a token without going through the whole nested structure. But I think we could also use `create_node_from_nested_root` for that so I can set the argument to `NA`.

PR #49, opened 2017-06-24
Indent curly

lorenzwalthert:
Similar to #21, where indention based on round brackets was added, this adds indention for curly brackets. However, there is one case that is not yet styled correctly (and the corresponding test fails on purpose). Spacing before curly brackets. Ideally, it would look like that:
```r
a <- function(x) {
  1
}
# or
if (x > 10) {
}
# and so on
```
The rule `add_space_before_brace()` remains ineffective since spacing is based on the concept of "space after tokens", i.e. spacing is added on the parse table to the column `spaces` the row before the token `{`. In the nested structure, there is no token before `{`, so no space is added. One way around that would be to add an empty token to every level of nesting so we can add spacing that way. This is pragmatic since we could still use all rules in R/rules.R, which are based on the assumption that the column `spaces` indicates the number of spaces after the token.


tests/testthat/indention_curly_brackets/multi_line_curly_only-out.R
```diff
@@ -0,0 +1,4 @@
+{
+  {1 + 3}
```
krlmlr: Do we have a transformer that adds spaces after/before the braces here?

lorenzwalthert: The spaces are added due to the first `{`. `transform_indent_curly()` adds braces  to everything between two curly braces.

lorenzwalthert:
Ok, I added the `"FIXME"` note so the tests should pass. So what do you think regarding the idea with the empty token, that would allow us to add spaces before `{`. Otherwise, I think we had to change the whole project and adapt "spaces before operators" and change all rules, serialisation etc. accordingly.

tests/testthat/indention_curly_brackets/multi_line_curly_round_spacing-out.R
```diff
@@ -0,0 +1,9 @@
+b <- function(x){ # FIXME indent before {
```
lorenzwalthert: Yes. I doubt that that is a very good idea because it is not a general rule to have a space after a round closing parenthesis, e.g. `call1(call2(1, 2), 3)`. We could make it conditional on having `function(` before `)` though. That might work.

lorenzwalthert:
@krlmlr I updated the PR (with rebase on master, new commits from `55babe2`). Can you have a look at `55babe2` in particular?

krlmlr:
I can't find the commit you mentioned.

krlmlr:
Sorry, can find them now. If you avoid backticks around Git refs (I edited your comment), the refs are converted to links.

tests/testthat/indention_curly_brackets/multi_line_curly_round_only-out.R
```diff
@@ -0,0 +1,9 @@
+a <- function(x) {
+  x <- c(1,
+    2 + 3,
+    sin(pi))
```
krlmlr: Does the second closing paren need to be on a new line?

tests/testthat/indention_curly_brackets/multi_line_curly_round_only-in_tree
```diff
@@ -0,0 +1,39 @@
...
+             ¦       ¦--',': , [1/0]                  
+             ¦       °--')': ) [0/0]                  
+             ¦--expr:  [1/8]                          
+             ¦   ¦--IF: if [0/0]                      
+             ¦   ¦--'(': ( [0/0]                      
+             ¦   ¦--expr:  [0/0]                      
+             ¦   ¦   ¦--'{': { [1/4]                  
+             ¦   ¦   ¦--expr:  [1/16]                 
+             ¦   ¦   ¦   ¦--expr:  [0/0]              
+"             ¦   ¦   ¦   ¦   °--STR_CONST: ""done [0/0]"
```
krlmlr: This looks a bit weird.

tests/testthat/indention_curly_brackets/one_line_curly-out.R
```diff
@@ -0,0 +1 @@
+a <- {1+1}
```
krlmlr: Spaces before/after these braces would be good.

tests/testthat/indention_curly_brackets/multi_line_curly_round_only-out.R
lorenzwalthert: well we have not yet attempted changing the line on which a token occurs, only the position within the line. This is to be addressed in the future.

tests/testthat/indention_curly_brackets/one_line_curly-out.R
lorenzwalthert: this test is testing indention based on curly brackets *only* (with the transformer function `style_indent_curly()`, see tests/testthat/test-indention_curly.R#3+). Such tests are aimed to only test a subset of the styling. Over-all tests are done with transformer function `style_text()` and friends, and they would add the proper spaces for this example (tested elsewhere).

tests/testthat/indention_curly_brackets/multi_line_curly_round_only-in_tree
lorenzwalthert: It's a double quote escaping issue. I fixed it using `write.table()` which allows for more options than `write_delim()`, so it should be fine now.

PR #50, opened 2017-06-26
write tree structure to file

lorenzwalthert:
Writes tree structures in the respective test folder as text files (without extension) so they can be opened in a text editor or Rstudio conveniently.
@krlmlr Should we commit these files as well? They are many but small, so I guess it makes sense since you want to have a look at those without having to run all test first...


DESCRIPTION
```diff
@@ -13,7 +13,8 @@ Imports:
     tidyr,
     purrr,
     magrittr,
-    pkgload
+    pkgload,
+    readr
```
lorenzwalthert: alphabetically you mean, right?

krlmlr: Yes, please

lorenzwalthert:
Ok. Should I commit the tree output files also?

krlmlr:
Yes, please commit the tree files too, so we can see what's happening at a glance.

tests/testthat/indention_round_brackets/arithmetic_no_start-in_tree
```diff
@@ -0,0 +1,19 @@
+levelName
+xxx
```
krlmlr: Can we make this a bit nicer?

lorenzwalthert: Yes, I made it look a bit better.
```
ROOT (token: short_text [newlines/spaces])
  °--expr:  [0/0]                          
      ¦--'(': ( [0/0]   
```

Issue #53, opened 2017-06-29
comments


lorenzwalthert:
There is still one case in which comments are not styled correctly, namely if they appear not on the first position of the line.

``` r
library("magrittr")
library("styler")
style_text(c("a <- {", 
  "q        #ijaf", 
  "}       #jkdaf", 
  "          #2")) %>% 
  cat(sep = "\n")
#> a <- {
#>   q        #ijaf
#> }       #jkdaf
#> #2
```
Also, there should be a space after `#`, `#'`, or `##` and so on.

PR #54, opened 2017-06-29
More on styling comments

lorenzwalthert:
This PR aims to solve the problems outlined in #53. 

* To that effect, two new rules `start_comments_with_space()` and `set_space_between_levels()` are introduced that set the first space within comments and before comments respectively.
* To test more flexibly, `test_collection()` and friends now support `...`.


R/nested.R
```diff
@@ -41,12 +41,31 @@ compute_parse_data_nested <- function(text) {
 #' [utils::getParseData()] is used to obtain a flat parse table from `text`.
 #' @param text A character vector.
 #' @return A flat parse table
+#' @importFrom tidyr separate
```
krlmlr: Are you still using `separate()`?

lorenzwalthert: Actually no, I use extract now. But since the whole package is imported anyways in R/ws.R, we would not need tidyr imports anyways.

Issue #51, opened 2017-06-27
indention with multiple brackets


lorenzwalthert:
Currently, styler adds indention for every opening bracket, which is the first problem.
``` r
library(magrittr)

test <- c("(((", "1", ")))")

styler::style_text(test) %>% cat(sep = "\n")
#> (((
#>       1
#>     )))
```
Instead of just 
```r
#> (((
#>   1
#>     )))
```
The tree looks as follows:
```r
styler:::compute_parse_data_nested(test) %>% styler:::visit(c(styler:::create_filler)) %>% 
  styler:::create_node_from_nested_root()
#>                                 levelName
#> 1  xxx                                   
#> 2   °--expr:  [0/0]                      
#> 3       ¦--'(': ( [0/0]                  
#> 4       ¦--expr:  [0/0]                  
#> 5       ¦   ¦--'(': ( [0/0]              
#> 6       ¦   ¦--expr:  [0/0]              
#> 7       ¦   ¦   ¦--'(': ( [1/0]          
#> 8       ¦   ¦   ¦--expr:  [1/0]          
#> 9       ¦   ¦   ¦   °--NUM_CONST: 1 [0/0]
#> 10      ¦   ¦   °--')': ) [0/0]          
#> 11      ¦   °--')': ) [0/0]              
#> 12      °--')': ) [0/0]
```
Second, it cannot "undo" indention when we put multiple closing brackets on one line, i.e. the last three brackets should not be indented, but this depends on what comes after the first bracket, and styler does not "look into the future" when serialising. That's why it gets it right when the brackets are on different lines.
```r
(((
      1
    )
  )
)
```

Rstudio only adds one level of indention per line, no matter how many opening brackets are there.
```r
((((((((
  1
))))))))
```
If the closing parenthesis are not on the same line, it un-indents all of them anyways:
```r
(((((
  1
)
)
)
)
)

```

I think the way to go is to remove indention whenever there is a child that also has indention and starts on the same line. I had it the other way round just now (see [here](https://github.com/lorenzwalthert/styler/commit/43981c877e4d97273a75b15ea695bb1e84308867)) which means only the first parenthesis has indention, all the subsequent "inner" parenthesis' indention was removed, but then the problem remains that the closing brackets are still indented by one level. 
The solution I suggest would involve setting the indention after the visitor, having to go through all levels of nesting  again. Alternatively, we could try to integrate it with the visitor we already have but I think that's a bad idea.

lorenzwalthert:
[...]

PR #57, opened 2017-07-01
Fixing indention with multiple parenthesis

lorenzwalthert:
This is supposed to be WIP and it's not yet optimised for performance. 

# Overview
* `indent_round()` and `indent_curly()` now first check (via `needs_indention()`) whether indention 
  is needed before they add it. The conditions are - loosely speaking - that there are not further
  indentions deeper in the nested structure that start on the same line. For the exact definition, I refer 
  to the help file of `needs_indention()`. This solves the first point brought up in #51, which is that
  multiple opening brackets on the same line should cause only one level of indention.

* `indent_round()` and `indent_curly()` now use `set_unindention_child()`, which sets the relative
  indention of closing token within children such as `{` and `(` to a negative number if necessary. This
  is necessary in cases where multiple closing brackets follow on the same line. In this case, an 
  additional closing bracket on a line causes the indention of the first bracket to become smaller (or 
  even zero). This addresses the second point brought up in #51. 

Consider the simple example:
```r
(((
  1
)))
```
The changes introduced by the first bullet above make sure that the indention of the innermost expression (which is `1`) is set to two spaces, the outer expressions don't feature any indention anymore. This means also that the closing brackets are unindented. The changes from the second bulled would not be necessary to make this simple example work.

Consider the more complicated example:
```r
call(
  call(
    16,
    5
))
```
The changes introduced by the second bullet above make sure the two closing brackets appear
right at the beginning of the line, with no spaces before them. Without the changes introduced by the second bullet above (but with the changes from the first bullet point), the expression `call(16, 5)` would be indented by two spaces (so would be the closing bracket of this expression). However, the closing bracket of this expression should not be indented since the last `)` of the whole example "undoes" the indention. To "undo" indention, we can assign a negative value (of two in this example) to the second last `)` and hence un-indent this very token (and consequentially, everything that follows on this line).

# To Do
* One change that was necessary due to the second bullet point was that `create_filler()` (which is
  called with the visitor on every level of nesting before anything else is done) cannot be used 
  anymore to set the indention level  (i.e. create the column indention) since we may have 
  manipulated it already when we were one level up since we set the indention of the children with
  `set_unindention_child()` (that is, `set_unindention_child()` affects not only the current level of
  nesting, but potentially also the children). That seems slightly suboptimal because now we need 
  to check in `create_filler()` and `set_unindention_child()` whether the column already exists.
* Regarding performance, maybe we can reduce the `which()` statements. Also, the `mutate()`
  statement in R/modify_pd.R#20 and the counterpart in `indent_curly()` may be moved into the 
  if-statment.


R/modify_pd.R
```diff
@@ -8,33 +8,93 @@ NULL
...
 indent_round <- function(pd, indent_by) {
-  opening <- which(pd$token == "'('")
-  if (length(opening) > 0) {
+  indention_needed <- needs_indention(pd, token = "'('")
+  if (indention_needed) {
+    opening <- which(pd$token == "'('")
     start <- opening + 1
     stop <- nrow(pd) - 1
   } else {
     start <- stop <- 0
```
lorenzwalthert: Yeah I think that's true. I was referring to that under To Do in the PR description.

R/modify_pd.R
```diff
@@ -8,33 +8,93 @@ NULL
...
+
+#' Check whether a parse table contains a token
+#'
+#' @param pd A parse table.
+#' @param token The token for which it should be checked whether it is in the
+#'   parse table.
+#' @return `TRUE` if the token is in the parse table, `FALSE` otherwise.
+pd_has_token <- function(pd, token) {
+  has_indention_token <- token %in% pd$token
+  if (any(has_indention_token)) {
```
lorenzwalthert: No, you are right, way more concise and even easier to read I think.

lorenzwalthert:
Regarding `"(((\n1 + 2) * (3 + 4\n)))"`: It gives this:
```
(((
    1 + 2) * (3 + 4
  )))
```
However, I think since arithmetic operators are involved, we better look at that once we solved the problem with the arithmetic operators.

In addition, we now face the following problem:
``` r
library("styler")
library("magrittr")
test <- c("call(", "call(", "call(", "call(", "1", "))))")
style_text(test) %>% cat(sep = "\n")
#> call(
#>   call(
#>     call(
#>       call(
#>         1
#>     ))))
```
This is not correct. The problem is that we do not currently add up unindention. In the case above, the innermost token should not just be unindented by 2 spaces, but by 3 levels times 2 spaces. However, we just indent once. Therefore, I added a commit that sums up the unindention over the nested levels of the parse data so the unindention of the innermost closing bracket is the sum over all higher-level unindentions, not just of this very level. Then, it gives the desired result.

krlmlr:
Would you mind adding `FIXME` tests for the last two examples you have shown?

PR #59, opened 2017-07-01
add function to style a single .R file. Closes #27.

### LINKED ISSUE #27: support for single-file-styling
Currently, `style_src()` only supports formatting all .R files in a directory. However, I think it is reasonable to support styling single files, i.e. with a function that takes a character vectors of file paths as input. 
What do you @krlmlr think? If we want to implement that, should it be a separate function? I think then it would be appropriate to change the name of `style_src()` to `style_dir()` and create a function `style_file()` or something like that.

lorenzwalthert:


lorenzwalthert:
@krlmlr maybe we should rename `style_src()` to `style_dir()` so the two functions (styling one file vs. styling all files in a directory) are easier to distinguish?

krlmlr:
Okay to rename as you suggested.

PR #62, opened 2017-07-02
remove spaces before comma

lorenzwalthert:


lorenzwalthert:
Yeah that's right. I think these were not yet merged into master then, so I could not fix them. In the meantime, these examples are in master, so I can fix them now - after rebasing on master.

PR #69, opened 2017-07-07
Indent with renesting

lorenzwalthert:
This pull request addresses two linked topics:

* The more complex topic is indention based on operators `+`, `-`, `*`, `/`, `%>%` etc. To solve this
  problem, I first had to add a function that changes the nested structure of the parse table. 
  This function rearranges the parse table so that elements between these infix operators are 
  on the same level of nesting. Also see the commit description in 613fb5e or the help files
  `re_nest()`, `move_children_up()`. I did not see a way of doing that differently while getting
  all edge cases right.

* correctly handle unary operators. The function that worked on the flat parse table does not work
  with the nested structure, so I created a new one. However, so far, the transformers for the 
  nested structure was just the transformers for the flat structure and some special transformers 
  for the nested structure. That is no longer true since this flat transformer is not needed on the 
  nested structure. How should we handle that? I mean eventually we can delete them anyways.

A remaining topic is indention in situations where we have both braces and parenthesis.
```r
# desired
1 + (
  2
)
# actual
1 + (
    2
)
```

Currently, the indention of parenthesis and operators such as `+` are summed up. Note that we
fixed the case of multiple parenthesis already (#68), i.e. their indention does not cumulate if they 
start on the same line.
```
((((
  1 # indented by one level
))))
```


lorenzwalthert:
I was not sure whether we want the raw trees in our *_tree_in or the re-nested trees. Since the name is *tree_in, I just used the raw trees. Alternatively, we may want to (also) write out the re-nested trees. What do you think?

PR #70, opened 2017-07-10
Post visitor

lorenzwalthert:
* in visit_post: we probably want to call `visit_post()`, not `visit()`, right? The trees are different
  but the outcome identical, which I don't understand.
* Also, maybe we should rename `visit()` and `visit_post()` to `pre_visit()` and `post_visit()`?
* Are you @krlmlr aware that since we use `flatten_partially()` within `compute_parse_data_nested()`,
  all tree outputs show the tree structure after having applied `flatten_partially()`. This can be 
  resolved by taking `flatten_partially()` out of `computed_parse_data_nested()` (as I had it 
  originally with `re_nest()` if we want to show the original tree structure in the `*tree_in` files.


R/relevel.R
```diff
@@ -1,72 +1,30 @@
...
-#' # tree-structure with re_nest post-processing
-#' styler:::create_tree(code, re_nest = TRUE)
-#' @inheritParams move_children_up
-re_nest <- function(pd_nested) {
-  if (is.null(pd_nested)) return()
-  pd_nested <- pd_nested %>%
-    move_children_up(token = c("'+'", "'-'", "SPECIAL", "'/'", "'*'")) %>%
-    mutate(child = map(child, re_nest))
-  pd_nested
+flatten_partially <- function(pd_nested) {
```
krlmlr: Currently we flatten only operators, right? Perhaps the function name could reflect this.

R/transform.R
```diff
@@ -96,6 +95,15 @@ visit <- function(pd_nested, funs) {
   pd_transformed
 }
 
+#' @rdname visit
+visit_post <- function(pd_nested, funs) {
```
krlmlr: I have no idea why I put it here, but I meant to put it in close proximity to `visit()`.

Issue #36, opened 2017-06-19
work with text instead of token


lorenzwalthert:
So far, all rules in R/rules.R operate on tokens, i.e. on this vector that contains tokens.
```r
op_token <- c(
  "'+'", "'-'", "'*'", "'/'", "'^'", "AND", "AND2", "EQ", "EQ_ASSIGN",
  "GE", "GT", "LE", "LEFT_ASSIGN", "LT", "NE", "OR", "OR2", "RIGHT_ASSIGN",
  "SPECIAL", "EQ_SUB", "ELSE")
```

For my PR #35, I was using the argument `op_token` in the function `indent_op()` in order to be consistent with previous work. It allows specifying all tokens that trigger indention. However, I think we should rather use text than token for two reasons:

* If people want to customise styler, they rather want to specify a text of a token (like `>=`) than spending time on figuring out what "LE", "GT" etc. stand for.
* More importantly, the token "SPECIAL" does not uniquely map to a text element, i.e. I suppose a case that is not covered by the tokens in `op_token` is declared as "SPECIAL". Hence, we can only define indention for all text elements corresponding to "SPECIAL".

How should we proceed?

krlmlr:
Thanks. I think `"SPECIAL"` are user-defined operators. If there's a unique correspondence between text and token, we should be able to choose between either, but we need to handle `"SPECIAL"`, too. We could keep using the token and define a user-friendly mapping.

lorenzwalthert:
#71 closes this issue since we work now with a modified text-token-mapping

PR #71, opened 2017-07-13
Tidy rules

lorenzwalthert:
This PR is WIP, it's not yet intended to be merged and some points need further
discussion.

## Content of PR
So far, styler did essentially one thing. Adding and removing spaces between
tokens within one line. With this PR, styler is extended to also add and remove
line breaks between the tokens. Concretely, we add

* line breaks according to the tidyverse style guide, section [2.4](http://style.tidyverse.org/syntax.html#indenting). I could not really produce valid R code where there was something else than a closing bracket after a closing curly bracket, so there is no rule that forces a line break after a closing curly bracket.
* assignment operator `<-` replaces equal assignment `=` as in section [2.6](http://style.tidyverse.org/syntax.html#assignment).
* Semicolons are removed and statements are put on a new line as section [2.7](http://style.tidyverse.org/syntax.html#semicolons) requires.
* Pipes `%>%` cause a line break as section [4.2](http://style.tidyverse.org/pipes.html#spacing-and-indenting) states.
* Brackets are added to function calls without arguments and no parentheses if they are chained by a pipe, as section [4.3](http://style.tidyverse.org/pipes.html#no-arguments) requires.

For those changes, it is essential to place the line break information on the
right level of nesting. Otherwise, another round of styling will add add
another line break since styler thinks that there is none yet.

Relating to #36, the text-to-token-mapping was slightly modified, that is,
all text that had token SPECIAL (like `%>%` or `%in%`) now map to a unique
identifier (like "SPECIAL-PIPE", or "SPECIAL-IN") or to the fall-back class
("SPECIAL-OTHER"). This is required as for example `%>%` should be followed
by a line break whereas we don't want to force a line break after `%in%`. It's
also necessary to add brackets to function calls in pipes as stated by the last
bullet point above.

### `line1`/ `line2`
A new problem coming up with manipulating line break information relates to
operations that use the (not updated) line information `line1` and `line2` -
notably the functions that create indention. Before, `indention_needed()`
checks whether there is a sub-expression within the parse table that starts
on the same line as the expression in the parse table but does not stop on
the very same line. If that was the case, no indention was needed. This check
was implemented by comparing `line1` and `line2`.

One way to go about that would be to base the check upon `lag_newlines`, but I
think this is quite expensive as we needed to check *recursively* whether *any*
of the sub-expressions has `lag_newline` > 0.
Another would be to change `line1` and `line2` while updating `lag_newlines`
and still use the *old* approach. Can't tell how well that would work accros
levels of nesting, since for example with `if (x < 3) { 2 } else { 4 }`, it's
not just moving one token from one line to the other, it is essentially moving
all children after the first opening parenthesis one line below, so for the
closing parenthesis etc. Maybe we can just update them on one level of nesting
and hope that will work? This seems more of a hack though.
Another not apprach
that is not very elegant either is to first do flat styling for line-breaks
only, write back and then read in again to do the nested styling.

@krlmlr do you see other options?

### Backward compatibility of flat approach
92129249 in rules-spacing.R changes `set_space_after_comma()` so spaces after
a comma are set to one. The space was previously only set to one if there was
no line break after the comma. This condition was dropped since now we may
rearrange the line breaks and we may remove the line-break. However, this
means that initial indention is not preserved for these cases anymore (for the
flat approach). Changing the order of the transformer functions to first
modify the line breaks would not help much since all spacing rules (written by
@krlmlr for the flat approch initialy) are based on the column `newlines`, not
`lag_newliens`. Maybe we should eventually consider rewriting these rules anyways to be
based on `lag_newlines` since I think there is no other dependency on
`newlines` apart from these rules, so we could simplify things and get rid of
this column.


R/rules-spacing.R
```diff
@@ -1,8 +1,16 @@
 math_token <- c("'+'", "'-'", "'*'", "'/'", "'^'")
+
+special_token <- lookup_new_special()
```
krlmlr: This looks brittle, we'd need an `#' @include` roxygen tag to ensure that the file where `lookup_new_special()` is defined is processed before this code is seen.

R/rules-replacement.R
```diff
@@ -0,0 +1,14 @@
+force_assignment_op <- function(pd) {
+  to_replace <- pd$token == "EQ_ASSIGN"
```
krlmlr: Is this replacement safe in all circumstances? I'd like to differentiate between styling options that only add/remove whitespace (because there we can easily prove that the code is unchanged) and those that do change the code. For a new package, the user can then gradually turn on styling, and review potentially breaking changes separately.

lorenzwalthert:
Ok, sorry @krlmlr I will file a separate PR first with these changes then. I rearranged commits before I pushed and forgot about a dependency. Should't happen...

R/rules-replacement.R
lorenzwalthert: I think it is safe, at least one test indicates that, not all `=` are replaced with `<-`.
```r
a <- 3
data_frame(a = 3)
```
However, I think you are right, maybe we should add more options to `get_transformers_nested()` so the user has more control over which transformer functions are included. What do you think?

krlmlr: Yes, at some point we need to think about categorizing transformers.

R/rules-other.R
```diff
@@ -0,0 +1,19 @@
+add_brackets_in_pipe <- function(pd) {
+  has_no_brackets <- (pd$token_before == "SPECIAL-PIPE") & (pd$token == "SYMBOL")
```
krlmlr: How does this deal with `. %>% base::names` ?

tests/testthat/line_breaks_and_other/other-out.R
```diff
@@ -0,0 +1,30 @@
...
+  b() %>%
+  c()
+
+# adding brackets in pipes
+a %>%
+  b() %>%
+  c()
+
+# adding brackets in pipes combined with line breaks
+a %>%
```
krlmlr: I think we should allow single-line pipes if all operations are on the same line.

R/rules-spacing.R
lorenzwalthert: ok, forgot that in the PR, but now it's committed to master.

Issue #77, opened 2017-07-14
Order of transformers


krlmlr:
1. Add/remove newlines
2. Indention
2. Add/remove spaces around tokens

Needs #76.

lorenzwalthert:
Reference: #263.

Issue #76, opened 2017-07-14
Simplify needs_indention()


krlmlr:
Is it enough to check if there's a newline after `opening` in `needs_indention()`?

```r
needs_indention <- function(pd, token = "'('") {
  opening <- which(pd$token %in% token)
  length(opening) > 0 && pd$newlines[opening + 1L]
}
```

Goal should be to avoid using `line1` and `line2` in this computation, see #75.

Think about removing `child_indents()` and perhaps `unindent_child()`.

Issue #75, opened 2017-07-14
Create new "multiline" attribute


krlmlr:
with a "post" visitor: Indicates if the current row (terminal or non-terminal) contains a newline. Compute this flag after adding/removing newlines, use it instead of `line1` and `line2`.

Note that this is only needed if we still need to use `line1` and `line2`.

Issue #73, opened 2017-07-14
Think about grouping transformers


krlmlr:
- Only add/remove space
- Only add/remove space or newlines
- Also edit text

PR #79, opened 2017-07-20
Use multi line

lorenzwalthert:
In the main, this PR refines `needs_indention()` and `set_unindention_child()` (as proposed in #76) such that we do no longer rely on `line1`/ `line2`, which is a problem because as introduced by #71, we do now modify line breaks but don't update `line1`/ `line2`. The problem is described in detail in #71. The solution is the creation of a multi-line attribute as proposed in #75 through a post-visitor after the line break rules were applied.
Linked to this, the transformers are now grouped into the following groups as suggested in #73:

* space modifiers
* Line break modifiers
* token modifiers

Plus they were rearranged (#77) so we first modify line breaks, then compute the multi-line attribute, then modify spaces and tokens.


R/parsed.R
```diff
@@ -89,7 +89,8 @@ create_filler <- function(pd_flat) {
       newlines = ~line3 - line2,
       lag_newlines = ~lag(newlines, default = 0),
       col2_nl = ~if_else(newlines > 0L, 0L, col2),
-      spaces = ~col3 - col2_nl - 1L
+      spaces = ~col3 - col2_nl - 1L,
+      multi_line = FALSE
```
krlmlr: We should use `NA` to initialize here.

R/serialized_tests.R
```diff
@@ -158,27 +164,41 @@ style_empty <- function(text) {
...
-                   partial(indent_curly, indent_by = 2),
-                   partial(indent_round, indent_by = 2),
-                   strip_eol_spaces)) %>%
-
+    apply_transformers(transformers) %>%
     serialize_parse_data_nested()
+
 }
 
 #' @describeIn test_transformer Transformations for indention based on operators
```
krlmlr: Do you want to update `style_op()`, too? Do we have a separate function for compute + apply transformers + serialize?

R/transform.R
```diff
@@ -71,11 +71,45 @@ make_transformer_nested <- function(transformers) {
   function(text) {
     text <- gsub(" +$", "", text)
     text <- gsub("\t", "        ", text)
+    if (is.null(transformers$space)) return(text)
 
     pd_nested <- compute_parse_data_nested(text)
```
krlmlr: compute + apply + serialize again

R/transform.R
```diff
@@ -71,11 +71,45 @@ make_transformer_nested <- function(transformers) {
...
+#'   information is not to be modified), all transformers are applied in one
+#'   step. The former requires two pre visits and one post visit, the latter
+#'   only one pre visit.
+#' @param pd_nested A nested parse table.
+#' @param transformers A list of *named* transformer functions
+#' @importFrom purrr flatten
+apply_transformers <- function(pd_nested, transformers) {
+  if (is.null(transformers$line_break)) {
+    transformed_all <- pre_visit(pd_nested, unlist(transformers))
+  } else {
```
krlmlr: If we keep only the `else` branch, code complexity is reduced. Does the code still work if we use `pre_visit()` instead of `post_visit()`?

tests/testthat/transformer_grouping/scope_none-out.R
```diff
@@ -0,0 +1,30 @@
...
+    if (x > 10) {
+      for (x in 22) { # FIXME in operator only to be surrounded by one space. What about %in%?
+        prin(x)
+      }
+    }
+  }     )
+  #we like comments too
+  c(list(x + 2),
+    c(    c(
+      26 ^ 2, # FIXME ^ operator has to be surrounded by one space (or none?!), never multiple
```
krlmlr: Is this still a FIXME?

tests/testthat/transformer_grouping/scope_none-out.R
```diff
@@ -0,0 +1,30 @@
...
+#'
+#' @param x a parameter.
+#'    indented comments
+a<- function(x){
+  test_that("I want to test",{
+    out <- c(1,c(
+      22 +1
+    ))
+    if (x > 10) {
+      for (x in 22) { # FIXME in operator only to be surrounded by one space. What about %in%?
```
krlmlr: Is this still a FIXME?

tests/testthat/transformer_grouping/scope_spaces-out.R
```diff
@@ -0,0 +1,20 @@
...
+  {
+    my_test(call)
+})
+
+# do not replace assignment
+a = 3
+data_frame(a = 3)
+
+# do not resolve semicolon
+a <- function(x) x + 1;b;c
```
krlmlr: I wouldn't mind a rule that converts semicolons (is there a rule in the style guide?), but this seems sufficiently low priority.

lorenzwalthert: Ok, It's already implemented though.

tests/testthat/transformer_grouping/scope_none-out.R
lorenzwalthert: Well I guess you don't want `x ^ y` but rather `x^y`, right?

tests/testthat/transformer_grouping/scope_none-out.R
lorenzwalthert: yes, too many spaces around `in` are not currently fixed.

R/transform.R
lorenzwalthert: you mean pre_visit for `set_multi_line`?

tests/testthat/transformer_grouping/scope_none-out.R
krlmlr: I think a space is fine, but we could offer alternatives (e.g., space only around + and -, never around *, /, ^).

tests/testthat/transformer_grouping/scope_none-out.R
krlmlr: Can you please file an issue, and perhaps change this example?

R/serialized_tests.R
lorenzwalthert: What do you mean regarding `stle_op()`? No, we don't have a function, but I can create one. But I am not sure whether it would make stuff more readable or just pack too much into one function.

```r
make_transformer_nested <- function(transformers) {
  function(text) {
    text <- gsub(" +$", "", text)
    text <- gsub("\t", "        ", text)
    if (is.null(transformers$space)) return(text)

    pd_nested <- compute_parse_data_nested(text)
    transformed_pd <- apply_transformers(pd_nested, transformers)
    # TODO verify_roundtrip
    new_text <- serialize_parse_data_nested(transformed_pd)
    new_text
  }
}
```
would become
```
make_transformer_nested <- function(transformers) {
  function(text) {
    text <- gsub(" +$", "", text)
    text <- gsub("\t", "        ", text)
    if (is.null(transformers$space)) return(text)
    new_text <- parse_transform_serialize(text, transformers)
    new_text
  }
}
```

Anyways, I think the two `gsub()` expressions should be packed in a function since they are rather low-level. What do you think?

R/transform.R
lorenzwalthert: see above

tests/testthat/transformer_grouping/scope_none-out.R
lorenzwalthert: ok.

R/get_transformers.R
```diff
@@ -47,32 +47,88 @@ get_transformers_flat <- function(strict = TRUE,
 #'
 #' Similar to [get_transformers_flat()], but additionally, returns some
 #'   functions needed due the fact that styling is done in a nested way.
+#' @param scope The extend of manipulation. Can range from "none" (least
```
lorenzwalthert: ups yes.

R/get_transformers.R
```diff
@@ -47,32 +47,88 @@ get_transformers_flat <- function(strict = TRUE,
...
+#'   invasive) to "token (most invasive). See 'Details'. This argument is a
```
lorenzwalthert: ups yes.

R/transform.R
krlmlr: Typo. Can we use `post_visit()` everywhere?

R/serialized_tests.R
krlmlr: `style_op()`: https://github.com/krlmlr/styler/pull/79/files/15d2e8b5cf094034a1bc607a6f4fec9fab793863#diff-4a86437341bb263c7208800d4495eb97R205

Please double-check that this is the right place to strip spaces at EOL and convert tabs to spaces. What are the semantics of `transformers$space`?

The fact that we can use `parse_transform_serialize()` in at least four places justifies extracting a function IMO.

R/get_transformers.R
```diff
@@ -47,33 +47,90 @@ get_transformers_flat <- function(strict = TRUE,
...
-    remove_space_after_unary_pm_nested,
-    set_space_before_comments,
-    set_space_between_levels,
-    add_brackets_in_pipe,
-    strip_eol_spaces,
+get_transformers_nested <- function(
+  scope = "tokens",
+  strict = TRUE,
+  indent_by = 2,
+  start_comments_with_one_space = scope) {
```
lorenzwalthert: yes, if you go down all the way. Should be boolean though and independent of scope, so I am going to change the default. In Roxygen comments, indention should not be removed I think

R/parsed.R
lorenzwalthert: Not sure. Terminals are not multi-line and we use a post-visitor. If we set it to `NA`, we have to set it to `FALSE` for terminals in order for it to work. Do you think it is worth it in terms of performance?
Also, create filler could be called on the flat structure, which may improve performance. However, then it would not be a transformer anymore and initialising variables would happen before. I think it's worth doing that. What do you think?

R/serialized_tests.R
lorenzwalthert: Regarding `style_op()`: Ok, I will do that in a separate PR then. Regarding EOL space stripping, I also thought about that. I think in terms of performance, it does not matter. But it might be more coherent to strip these in the serialisation step. Ok?

`transformers$space` is a unnamed list of space transformers, so are `transformers$line` and `transformers$eol` (which has only one function though). The distinction is made since between line manipulation and space manipulation, we need to set the multi-line attribute.

tests/testthat/transformer_grouping/scope_none-out.R
lorenzwalthert: ok, will open an issue.

R/serialized_tests.R
krlmlr: Agree to extract `parse_transform_serialize()` later: #84.

The stripping of space is done only to verify the lossless roundtrip (which has been there for the flat transform but not yet for the nested): #85.

tests/testthat/transformer_grouping/scope_none-out.R
krlmlr: For reference: #82.

krlmlr:
NB: If you push intermediate changes, the discussions that affect outdated changes are hidden.

R/parsed.R
lorenzwalthert: Ok, I opened an issue.

R/transform.R
lorenzwalthert: No, because for the un-indention we need to propagate from outside in. But it seems as we could use `post_visit()` for the line-breaks and setting the multi-line attribute with the same visitor. Will look into that in #88

Issue #78, opened 2017-07-19
make styler faster


lorenzwalthert:
The following could make styler run faster: 

* dropping unnecessary `select()` and `arrange()` statements.
* early termination (e.g. rules?)
* consider replacing some tidyverse calls with base R if substitution
   does not make the code "uggly"? What do you think @krlmlr ?
* which statements related to `needs_indention()`

``` r
microbenchmark::microbenchmark(dplyr = dplyr::mutate(mtcars, driver = 3), 
                               baseR = mtcars$driver <- 3)
#> Unit: microseconds
#>   expr      min       lq      mean   median        uq        max neval cld
#>  dplyr 1969.341 2009.056 3740.7995 2037.379 2091.4205 152117.420   100   b
#>  baseR   11.663   13.735   24.1628   20.456   22.9275    433.822   100  a

microbenchmark::microbenchmark(dplyr = dplyr::slice(mtcars, 1:3), 
                               baseR = mtcars[1:3, ])
#> Unit: microseconds
#>   expr      min       lq      mean    median       uq      max neval cld
#>  dplyr 2094.290 2148.480 2446.0150 2191.8545 2271.761 8700.220   100   b
#>  baseR  138.072  155.203  176.9532  165.3705  189.216  431.295   100  a

# pipe does probably not slow down things much.
library(magrittr)
microbenchmark::microbenchmark(dplyr = mtcars %>% summary(), 
                               baseR = summary(mtcars))
#> Unit: milliseconds
#>   expr      min       lq     mean   median       uq      max neval cld
#>  dplyr 4.550347 4.619500 5.089398 4.695070 4.833018 9.004815   100   a
#>  baseR 4.390407 4.497096 4.924098 4.548395 4.833101 8.284952   100   a
```

These are just some thoughts, profiling of *our* code is probably a better place to start anyways...

krlmlr:
Let's identify our bottlenecks first, you could try profvis.

lorenzwalthert:
WIth #90, I think we can close this, at least for now.

PR #90, opened 2017-07-25
Speedup

lorenzwalthert:
On an example file, the changes introduced with this PR speed up styler up to 4x. For styling a whole package, the speed gains turn out to be much less, only about 2.5x. Main bottle neck now is  `tidyr::nest()`.

This PR is based on #78


R/modify_pd.R
```diff
@@ -9,33 +9,27 @@ NULL
...
   }
 
   pd %>%
     set_unindention_child(token = "')'", unindent_by = indent_by)
 }
 #' @rdname update_indention
 indent_curly <- function(pd, indent_by) {
-  indention_needed <- needs_indention(pd, token = "'{'")
+  opening <- which(pd$token == "'{'")
+  indention_needed <- needs_indention(pd, token = "'{'", opening[1])
```
krlmlr: Can we define a function `get_indent_indices()` instead that returns a (possibly empty) integer vector of positions that need to have indention added?

```r
indent_indices <- get_indent_indices(pd, ...)
if (length(indent_indices) > 0L) pd$indent[indent_indices] <- pd$indent[indent_indices] + 2L
```

Also, the benchmarks suggest that this doesn't buy us anything.

R/modify_pd.R
```diff
@@ -127,6 +121,6 @@ token_is_multi_line <- function(pd) {
 #' @param pd_flat A flat parse table.
 #' @return A nested parse table.
 strip_eol_spaces <- function(pd_flat) {
-  pd_flat %>%
-    mutate(spaces = spaces * (lead(lag_newlines, default = 0) == 0))
+  pd_flat$spaces <- pd_flat$spaces * (lead(pd_flat$lag_newlines, default = 0) == 0)
```
krlmlr: The following pattern seems slightly easier to read:

```r
idx <- which(...)
pd_flat$spaces[idx] <- 0L
pd_flat
```

R/nested.R
```diff
@@ -29,10 +29,10 @@ compute_parse_data_nested <- function(text) {
     add_terminal_token_before() %>%
     add_terminal_token_after()
 
+  parse_data$child <- rep(list(NULL), length(parse_data$text))
+  parse_data$short <- substr(parse_data$text, 1, 5)
```
krlmlr: The `short` is optional and could be added to the "flat" parse data, too.

R/nested.R
```diff
@@ -98,19 +98,23 @@ NULL
...
+
+  data_frame(id = terminals$id,
+             token_after = lead(terminals$token, default = "")) %>%
     left_join(pd_flat, ., by = "id")
 }
 
 #' @rdname add_token_terminal
 add_terminal_token_before <- function(pd_flat) {
-  pd_flat %>%
+  terminals <- pd_flat %>%
```
krlmlr: Maybe faster:

```r
terminals <- which(pd_flat$terminals)
order <- order(pd_flat$line1, pd_flat$col1)[terminals]
data_frame(id = pd_flat$id[order], token_before = ...) %>% ...
```

Or:

```r
terminals <- which(pd_flat$terminals)
order <- order(pd_flat$line1[terminals], pd_flat$col1[terminals])
data_frame(id = pd_flat$id[terminals][order], token_before = ...) %>% ...
```

R/nested.R
```diff
@@ -146,24 +150,25 @@ set_spaces <- function(spaces_after_prefix, force_one) {
 #' @importFrom purrr map2
 nest_parse_data <- function(pd_flat) {
   if (all(pd_flat$parent <= 0)) return(pd_flat)
+
+
+  pd_flat$internal <- with(pd_flat, (id %in% parent) | (parent <= 0))
   split <- pd_flat %>%
-    mutate_(internal = ~ (id %in% parent) | (parent <= 0)) %>%
-    nest_("data", names(pd_flat))
+    nest_("data", setdiff(names(pd_flat), "internal"))
```
krlmlr: You could try `split_data <- split(pd_flat[...], pd_flat$internal)` and use `split_data` instead of `split$data` below.

R/rules-spacing.R
```diff
@@ -144,29 +151,37 @@ set_space_between_levels <- function(pd_flat) {
...
+  if (all(!comment_pos)) return(pd)
 
-  if (nrow(comments) == 0) return(pd)
+  comments <- pd[comment_pos, ]
 
-  non_comments <- pd %>%
-    filter(token != "COMMENT")
+  non_comments <-pd[pd$token != "COMMENT", ]
 
   comments <- comments %>%
```
krlmlr: A pipe with just one step?

R/rules-spacing.R
```diff
@@ -144,29 +151,37 @@ set_space_between_levels <- function(pd_flat) {
...
+
+  comments[, setdiff(names(comments), c("space_after_prefix", "prefix"))] %>%
+    bind_rows(non_comments) %>%
     arrange(line1, col1)
 }
 
 
 set_space_before_comments <- function(pd_flat) {
   comment_after <- pd_flat$token == "COMMENT"
+  if (all(!comment_after)) return(pd_flat)
```
krlmlr: `!any()` might be slightly faster (also elsewhere).

R/unindent.R
```diff
@@ -5,7 +5,7 @@
 #' @inheritParams unindent_child
 #' @importFrom purrr map
 set_unindention_child <- function(pd, token = "')'", unindent_by) {
-  if(all(pd$terminal) | all(pd$indent == 0)) return(pd)
+  if(all(pd$indent == 0) | all(pd$terminal) ) return(pd)
```
krlmlr: `||` is slightly clearer, no performance difference here.

R/utils.R
```diff
@@ -5,9 +5,10 @@ parse_text <- function(x) parse(text = x)[[1L]]
 #' @param char A character vector.
 #' @param times an integer giving the number of repetitions.
 #' @return A character vector.
+#' @importFrom purrr map map_chr
 rep_char <- function(char, times) {
-  lapply(times, rep.int, x = char) %>%
-    vapply(paste, collapse = "", character(1L))
+  map(times, rep.int, x = char) %>%
```
krlmlr: Can `times` be a vector here?

R/visit.R
```diff
@@ -17,18 +17,19 @@ NULL
 pre_visit <- function(pd_nested, funs) {
   if (is.null(pd_nested)) return()
   pd_transformed <- pd_nested %>%
-    visit_one(funs) %>%
-    mutate(child = map(child, pre_visit, funs = funs))
+    visit_one(funs)
```
krlmlr: One-step pipe?

R/modify_pd.R
lorenzwalthert: Ok, sounds good. I think since `dplyr::between()` returns boolean values, we better call it `compute_indent_flags()` and do boolean subsetting instead of integer subsetting. Does that sound reasonable?
```r
indent_round <- function(pd, indent_by) {
  indent_flags <- compute_indent_flags(pd, token = "'('")
  pd$indent[indent_flags] <- pd$indent[indent_flags] + indent_by
  pd %>%
    set_unindention_child(token = "')'", unindent_by = indent_by)
}

compute_indent_flags <- function(pd, token = "'('") {
  npd <- nrow(pd)
  opening <- which(pd$token == token)
  if (!needs_indention(pd, token, opening[1])) return()
  start <- opening + 1
  stop <- npd - 1
  between(seq_len(npd), start, stop)
}
```
It has only slightly worse performance (<1%) so I think we should do it. It also reduces code duplication.

R/rules-spacing.R
lorenzwalthert: Yes. Should we rather do?

```r
extract(comments, text, ...)
```

R/rules-spacing.R
lorenzwalthert: Ok, Can try that. Had it before, but then I felt it's less legible, but did not think in terms of speed.

R/unindent.R
lorenzwalthert: 👍 Will try.

R/utils.R
lorenzwalthert: For the flat serialization, `newlines_and_spaces()` uses `rep_char()` with a vectorised `times` input and for setting spaces at the beginning of comments we do that. I think we can use `map()` in these two places. It speeds things up quite a bit (~15%). Thanks for the hint.

R/visit.R
lorenzwalthert: I can change this to a one-liner without pipe.

R/modify_pd.R
lorenzwalthert: Or flags instead of the (expensive) which?
```r

idx <- lead(pd_flat$lag_newlines, default = 0) != 0
  pd_flat$spaces[idx] <- 0
```

R/rules-spacing.R
lorenzwalthert: Changed it, but it's not really faster.

lorenzwalthert:
I added a profiling of commit cbf94f2 of a package (with 10 styled files) [here](http://rpubs.com/lorenzwalthert/styler-profiling-v3). With this example, styling is about 4 times faster than before optimising for performance (commit 59de93f) .

R/nested.R
lorenzwalthert: yes, that works well.

R/modify_pd.R
krlmlr: Indices have the advantage that they can be checked for length zero, and that they only contain the positions that we care about. Essentially, `compute_indent_indices <- function(...) which(compute_indent_flags(...))`.

R/nested.R
krlmlr: Do we still need `short`?

R/modify_pd.R
lorenzwalthert: Ok. I can try that. When I checked the [profiling](http://rpubs.com/lorenzwalthert/styler-profiling-v3) I just got the impression that `which()` is expensive.

R/nested.R
lorenzwalthert: We don't use it further no. I It just helps when working interactively. Should we drop it?

R/nested.R
lorenzwalthert: I tried that but I felt since this function is only called once and it seems pretty [inexpensive](http://rpubs.com/lorenzwalthert/styler-profiling-v3) (10 ms out of 15'460 for the whole run, file R/nested.R), I left it as is, for better legibility. Or do you prefer the rearrangement anyways?

krlmlr: I didn't know that, I just noticed you changed it and assumed that performance matters here. Never mind.

R/nested.R
krlmlr: If it helps we should keep it for now.

R/modify_pd.R
krlmlr: I'm fine with flags if it works well enough.

lorenzwalthert: Ok no, I just figured out that `which()` is not expensive. It was the comparison inside `which()`, which we can't avoid anyways. Also, always having a numerical vector is better than having NULL sometimes, so I use indices anyways.

R/nested.R
lorenzwalthert: moved it into `tokenize()`

R/rules-spacing.R
lorenzwalthert: done

Issue #94, opened 2017-07-27
Indention of function declarations


krlmlr:
Example:

```r
a <- function(x, y,
              z)
{
}
```

To achieve this, we could:

1. When processing `z`, memoize the ID of `x` as a reference for the indent depth (because we can't know how deep `x` will be indented)
2. After applying all transformers, convert relative indents to absolute indents with a "pre" visitor
3. Flatten the nested structure with a "post" visitor
4. In the flat structure, update the absolute indention depth of `z` with that of `x` with a lookup in the flat structure
5. Serialize

We could omit step 3 if the reference token is always at the same level in the nested structure.

lorenzwalthert:
Closed in favour of the more general #66

Issue #85, opened 2017-07-25
Verify lossless parse + serialize (without transforms)


krlmlr:
A run without any transformers enabled should yield exactly the same results as the input, minus spaces at EOL and conversion of tabs to spaces. This should be tested separately on all input files, for easier isolation of parse/serialize problems from transformer problems during later refactorings.

Reference: https://github.com/krlmlr/styler/pull/79#discussion_r129269504

lorenzwalthert:
This is actually not easily possible at the moment. The reason is that all space information is discarded for the last token on a line, since the column `indention` will take care of the space between the last token on a line and the next token. If we want to change that, we need  to add `lag_spaces` and change the serialisation. This may not be easy.

krlmlr:
Agreed. The roundtrip test doesn't need to care about spaces at EOF and tabs. We just verify that for each text t, s(p(e(t))) = e(t) with s() = serialize, p() = parse, e = strip spaces at EOF + convert tabs to spaces.

lorenzwalthert:
I am not sure if I understand you correctly, but I think that is not easily possible. It is easily possible if we remove all spaces, which you probably don't mean, right? Since we add spaces *after* the token and not before it (that is actually why we need eol stripping as a transformer) but add new lines *before* it , applying no transformers and serialising back will not yield the input. If we wanted this to be possible, we had to use `lag_spaces` instead of spaces which we currently don't do anywhere and adapt the serialization. Also, I think issues like in #89 cannot be detected with the approach you are suggesting, so I am not sure how much we can gain and the complexity of implementing this feature, should not be underestimated.

krlmlr:
I'm fine if we can find a set of transformers that will allow us to test the roundtrip: for each text t, s(x(p(e(t)))) = e(t), with (in addition) x() = minimally sufficient set of transformers.

I think the main reason why we strip spaces at all is that spaces between tokens that also have a newline aren't properly recognized: The parse data isn't different for

```r
a +
  b
```

and

```r
a +    ~
  b
```

(substitute `~` for another space in the second example). OTOH, spaces at the end of comments and spaces before newlines embedded into string literals are kept. So, erasing space in the text is potentially destructive.

Let's think a bit more about testing the roundtrip, just remove the two `gsub()` calls at https://github.com/krlmlr/styler/pull/79/files/15d2e8b5cf094034a1bc607a6f4fec9fab793863#diff-ed8a73a1134b4bd8c65cef9e3365dff5R72.

[...]

Issue #66, opened 2017-07-04
Indention dependent on token-length


lorenzwalthert:
styer indents two spaces for each level:
```r
call(
  call(
    call(3
    )
  )
)
```
However, we generally don't want two spaces. If the first argument of the indenting call follows on the same line, we want to indent depending on the length of the indenting token (at least RStudio does that):
```r
very_long_call_here(1  
                    second_argument,
                    third_argument
)
```

lorenzwalthert:
The illustrative example above is actual a special case for which it is easy to get the correct indention, since we simply have to check the token length of `very_long_call_here` (and add two spaces) and we are there. However, the real challenge is to get something like the following right:

```r
a <- call(2,
          2
)
```
Here, we essentially have to know where the token `call` starts, not just how long it is. We only need to know the position of the token relative to the indention though, which is different from #65, where we probably need to know the absolute position of the token.

krlmlr:
This is relevant for `function()` declarations in the tidyverse style guide.

lorenzwalthert:
Similar to #94

lorenzwalthert:
Agree that for the tidyverse style guide, only function declarations are relevant. However, I think it is preferable to also implement it for function calls and then add a rule that adds a line break after the opening round brace so we can leave it to the style guide in use whether or not to also indent token-dependent for function calls.

PR #106, opened 2017-07-31
Token dependent indention

### LINKED ISSUE #104: Decide on the  usage of the `newlines` attribute
A few rules still use `newlines`, which leads to problems since all line-break rules only modify `lag_newlines`, so `newlines` may contain outdated information. 
Also, if we want to allow non-invasive styling without indention correction, we need to set spaces only for tokens that are not at the end of a line. Otherwise, we "destroy" the indention information provided in the raw data. 
One may think we could use `dplyr::lead(pd$lag_newlines)`, which may work since to create `lag_newlines`, we only introduced one unknown value (at the top of the parse table) as this was done in the flat structure. Using `dplyr::lead(pd$lag_newlines)` in each nested structure will create one unknown value for the last token (instead of `pd$newlines(nrow(pd))`. We may loose information

We may need another visitor to update `newlines` (or update it right in the place where we update `lag_newlines`) so we can use this attribute when manipulating spacing information (i.e. making sure we do not update spacing information for the last token on a line).

### LINKED ISSUE #105: Allow styling without modifying indention
This seems possible by adapting #94.

lorenzwalthert:
This PR consists of two parts:

**Step 1: Allow for token-dependent indention**
This implements #94 (and hence #66) rudimentary. By flattening out the nested parse table completely and computing the absolute position a token will take when serialised, we can use a look-up approach to set the indention of any token dependent on the indention of another token. There is room for improvement since `set_token_dependent_indention()` is really just work in progress and can be improved.

**Step 2: Allow for styling without touching indention**
Since step 1 implies to merge the spacing and indention information in `enrich_terminals()`into the column `spaces`, we are also able to discard the indention computed and use what we called raw indention, that is, take the indention from the input and don't modify it. The only thing for it to work is to not overwrite the spacing information if there is a line break after a token, which is checked via `newlines`. Since we modify `lag_newlines` before (in R/rules-line_breaks.R) and not `newlines`, this attribute needs to be updated after line breaks are set (closes #104). Hence, step 2 closes #105 and lays the foundations for #85.


lorenzwalthert:
I think this is quite a bit to review. Maybe we should even split it up. I believe that the code is not too bad, apart from everything in R/indent-token-dependent. I think I still have not found a good way of indenting token-dependent. It's all based on indices and maybe there are simpler ways to solve it. 
Since for the tidyverse style guide, we only have token-dependent indention, we would not to have to bother about everything that is not a function declaration, but I found it better to solve the function calls also, i.e. 
```r
call(1, 
     3)
```
And then add a rule that turns that into
```r
call(
  1, 
  3)
```
Or even
```r
call(
  1, 
  3
)
```

krlmlr:
If you can split it into smaller pieces, please do by all means -- these will be much easier to review.

lorenzwalthert:
Ok, will do so. Will require some time though.

lorenzwalthert:
This PR was split up in the above referenced pull requests and can hence be closed.

PR #115, opened 2017-08-07
definitively fixing eol issue with comments

lorenzwalthert:
I think the problem can be solved by simply trimming white spaces *after* the space is inserted between `#` and text, otherwise we miss spaces at the end of comments.

### DIFF
--- a/tests/testthat/indention_multiple/overall-in.R
+++ b/tests/testthat/indention_multiple/overall-in.R
@@ -1,30 +1,31 @@
-#'this function does
+#' this function does
 #'
 #' @param x a parameter.
 #'    indented comments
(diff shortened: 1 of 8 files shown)


tests/testthat/indention_multiple/overall-in.R
```diff
@@ -1,30 +1,31 @@
-#'this function does
+#' this function does
```
krlmlr: Did you mean to change the "in" file?

lorenzwalthert: oops, actually not but we have another test case to cover this so I leave it for now if that's ok

PR #116, opened 2017-08-07
Pr106 4 raw indention

### LINKED ISSUE #104: Decide on the  usage of the `newlines` attribute

### LINKED ISSUE #105: Allow styling without modifying indention

lorenzwalthert:
This is the third part of #106. Styler can now leave indention as is 😊. A new level of the argument `scope` is introduced ("indention"). "spaces" will not modify indention anymore (closes #104, closes #105). Note that since we use named transformers in R/transform.R (because depending on what is modified, the transformer needs to be called at a certain point in the process), we can also use values other than functions in the transformers. I think that's a natural extension plus since the parse table is already flattened out, I think this cannot be done well with a transformer function.


R/modify_pd.R
```diff
@@ -110,3 +110,21 @@ set_multi_line <- function(pd) {
...
+#' attribute. We cannot simply use `dplyr::lead(pd$lag_newlines)` since we would
+#' loose information for the last token. `spaces` is left as is in
+#' R/rules-spacing.R for tokens at the end of a line since this allows styling
+#' without touching indention.
+#' @param pd A parse table.
+#' @return A parse table with syncronized `lag_newlines` and `newlines` columns.
+#' @seealso choose_indention
+update_newlines <- function(pd) {
+  npd <- nrow(pd) - 1
+  pd$newlines[seq_len(npd)] <- pd$lag_newlines[seq_len(npd) + 1]
```
krlmlr: Any reason not to use `dplyr::lead()`?

lorenzwalthert: Yes, because then we'd lose information since we would end up with `lead(lag(newlines))` and the last entry would be `NA`. Therefore, we do the shift manually and do not replace the last value in `newlines` (since this one was not modified anyways by our rules). I think this is safe because we anyways only work with `lag_newlines` in all of the rules, so if someone wants to add blank lines after a token, he'd have to insert it not after the token, but before the next token (this is possible via `pd$token_before`).

PR #99, opened 2017-07-28
Op assignment indention

lorenzwalthert:
fixing:
* EOL spaces in empty comments (and in general) (closes #98).
* mal-indention in conditional statement due to wrong specification of indent_without_paren) (fixes #95).
* complicated indentions based on arithmetic and special operators (fixes #96).
* indention interaction on with assignment operator and other operators (closes #97).

#97 and #96 was a bit complicated since solving the problem required to further flatten the nested parse tables.

Issue #118, opened 2017-08-08
implementation of breaking long function calls 


lorenzwalthert:
The [tidyverse style guide](http://style.tidyverse.org/syntax.html#long-lines) states that long function calls are not compliant with the style guide:

```r
# Good
do_something_very_complicated(
  "that",
  requires = many,
  arguments = "some of which may be long"
)

# Bad
do_something_very_complicated("that", requires, many, arguments,
                              "some of which may be long"
                              )
```

It also states that multiple arguments can remain on the same line if they are related.
```r
# Good
paste0(
  "Requirement: ", requires, "\n",
  "Result: ", result, "\n"
)

# Bad
paste0(
  "Requirement: ", requires,
  "\n", "Result: ",
  result, "\n")
```

I am not sure how to implement that and how strictly we should use it. A simple rule would be to add a line break after `(`, and leaving everything until `)` untouched in terms of line breaks and putting the closing `)` on a new line. Also, I think this should be a transformer that is only used if `strict = TRUE` since I think its one of the rules in the style guide that is not consistently obeyed by many people. @krlmlr maybe you can advise on that.

krlmlr:
Your "simple rule" sounds good. It appears that we see the entire function call at the same level in the nested parse data -- we could look at `col2` for the closing paren and decide, but this seems brittle if we insert line breaks or even space beforehand. We seem to have a similar problem for function declarations.

I suggest we leave this open for now.

lorenzwalthert:
On a related note, I am also not sure about how closing brackets and pipes should interact, i.e.
```r
call(b,
     c, d) %>%
  x()

# vs.
call(
  b, c, d
) %>%
  x
```

lorenzwalthert:
I am not sure whether I understand you correctly, but I think we don't need to use `col2`. I think we can simply use `lag_newlines` to determine if there is a line break on the currently visited level. Since we use `pre_visit()`, we will deal with all multi-level expressions later during the visitting.
A simple implementation would be the following:
```r
break_line_after_opening_if_call_is_multiline <- function(pd) {
  npd <- nrow(pd)
  if (npd < 2) return(pd)
  if (all(is.na(pd$token_before[2]))) return(pd)
  is_call <- pd$token_before[2] == "SYMBOL_FUNCTION_CALL"
  if (!is_call) return(pd)
  if (npd == 3) {
    pd$lag_newlines[3] <- 0L
    return(pd)
  }
  comments <- which(pd$token == "COMMENT")
[...]

Issue #112, opened 2017-08-07
Create key for sorting


lorenzwalthert:
`line1`/ `col1` is no longer a unique sorting (composite) key because we insert tokens like with 
`add_brackets_in_pipe()`. One way to go about that would be to setting `col1`/ `col2` of these tokens 
"in between" the existing tokens. New tokens could then get intermediate values (e.g., when `()` is added: key + 1/3 and key + 2/3 respectively).
See discussion in #111 R/visit.R L136+.

krlmlr:
Reference: https://github.com/krlmlr/styler/pull/111#discussion_r131527625

lorenzwalthert:
Reference ([#119](https://github.com/krlmlr/styler/pull/119#discussion_r132647691)).

lorenzwalthert:
Maybe can simply use column `id`. Check with help file of `getParseData()`.

lorenzwalthert:
Reference: #147.

PR #119, opened 2017-08-08
Re-indent token-dependent

### LINKED ISSUE #66: Indention dependent on token-length

lorenzwalthert:
This is the fourth part of #106 and it is revised significantly. This PR closes #99 and closes #66 by allowing token-dependent re-indention. This is achieved by using the visitor to add references to tokens if necessary. Tokens that don't need re-indention have `NA`s in the corresponding reference column. The reference is propagated to the terminals. Once the parse table is flattened out and the position of every token is defined in terms of `line1`, `line2`, `col1`, `col2`, the indention of reference token is applied to the token that should inherit the indention from the reference token and the corresponding attributes in the parse table are updated.

A transformer function that breaks lines for long function calls as suggested in the [tidyverse style guide](http://style.tidyverse.org/syntax.html#long-lines) is not yet pushed to this branch. It seems rather invasive compared to the other rules. Nevertheless, I think we should include it. I was just not sure about the concrete implementation and opened #118.


tests/testthat/indention_curly_brackets/multi_line_curly_round_only-out.R
```diff
@@ -1,7 +1,7 @@
 a <- function(x) {
   x <- c(1,
     2 + 3,
-    sin(pi))
+    sin(pi)) # FIXME add tidyverse-comliant rule to break after '('
```
krlmlr: Can we always move the closing paren to a separate line?

tests/testthat/indention_token_dependent/token_dependent_comments-out.R
```diff
@@ -0,0 +1,7 @@
+call(call( # comment
+  3, 4
+))
+
+call(call(1, # comment
+          3
```
krlmlr: I'd rather have the `3` indented by two space by default, and only optionally support indention after the opening paren. I remember a discussion about this but can't currently find it.

R/reindent.R
```diff
@@ -0,0 +1,134 @@
...
+#' Applies the reference indention created with functions
+#' [update_indention_ref()] to the flattened parse table. The indention
+#' is applied to all token that inherit from a reference token sequentially,
+#' i.e. by looping over the target tokens.
+#' @inheritParams apply_ref_indention_one
+apply_ref_indention <- function(flattened_pd) {
+  target_tokens <- which(flattened_pd$id %in% flattened_pd$indention_ref_id)
+
+  # needs sequential update
+  for (target_token in target_tokens) {
```
krlmlr: `Reduce()`? I'm not in love with this code; a vectorized solution that updates the parse table only once would be great, but we can refactor later if necessary.

R/reindent.R
```diff
@@ -0,0 +1,134 @@
...
+    flattened_pd <- apply_ref_indention_one(flattened_pd, target_token)
+  }
+  flattened_pd
+}
+
+#' Applying reference indention of a target token
+#'
+#' Applies the indention level of `target_token` to all tokens that have
+#' `target_token` as a reference. This includes adding spaces to the first
+#' tokens on a line and updating the column `col1` and `col2` for all token
```
krlmlr: token -> tokens?

R/reindent.R
```diff
@@ -0,0 +1,134 @@
...
+  token_to_update <- which(token_points_to_ref & first_token_on_line)
+
+  # udate spaces
+  copied_spaces <- flattened_pd$col2[target_token] + 1
+  old_spaces <- flattened_pd$lag_spaces[token_to_update[1]]
+  shift <- copied_spaces - old_spaces
+  flattened_pd$lag_spaces[token_to_update] <-
+    flattened_pd$lag_spaces[token_to_update] + shift
+
+  # update col1 / col2
```
krlmlr: I wonder if we really need to drag along `col1` and `col2` (#112), but we can scrap that later.

R/reindent.R
```diff
@@ -0,0 +1,134 @@
...
+#'   ends in terms of col2.
+#' @examples
+#' \dontrun{
+#' a <- function(x,
+#'               y) {
+#' x + y
+#' }
+#' }
+update_indention_ref_fun_dec <- function(pd_nested) {
+  if (pd_nested$token[1] == c("FUNCTION") &&
```
krlmlr: The `c()` looks redundant.

R/reindent.R
```diff
@@ -0,0 +1,134 @@
...
+      nrow(pd_nested) > 3) {
+    seq <- 3:(nrow(pd_nested) - 1)
```
krlmlr: Maybe add a safe `seq_from_to()` helper that will return an increasing sequence starting at `from` and ending at `to - 1`, empty if `from >= to`? I really think `:` should be avoided in package code; you do check for safety but this adds complexity to the code (here and elsewhere).

R/reindent.R
lorenzwalthert: I think we have to since indention of a token can be based on a re-indented token, hence the col information needs to be kept up to date.

tests/testthat/indention_curly_brackets/multi_line_curly_round_only-out.R
lorenzwalthert: I think we could. The question is whether that is non-invasive. I guess you only refer th the case where the function call spans over multiple lines, right? I suggest to move that discussion to #118.

tests/testthat/indention_token_dependent/token_dependent_comments-out.R
lorenzwalthert: Ok. Then, I guess the question is whether we should break after `(` by default. Also, what means "by default"? I would suggest to make it dependent on the `strict` argument. Alternatively, we can define another level of invasiveness. 
See #118.

R/reindent.R
lorenzwalthert: Ok, that sounds like a good idea. Why should it go to `to -1 `and not to `to`?

R/reindent.R
lorenzwalthert: yep

R/reindent.R
lorenzwalthert: I think it needs to be sequential, because some re-indention is based on token that were itself re-indented (at least in the case of function call), so I guess `Reduce()` (or `purrr::reduce()` is the way to go.

R/reindent.R
krlmlr: Both work, it's a matter of taste.

tests/testthat/indention_token_dependent/token_dependent_comments-out.R
krlmlr: We could indent with two space if there's a break, or optionally indent below the opening paren. Adding line breaks looks like an independent problem to me.

R/reindent.R
krlmlr: A bit late, but rlang already has `seq2()`.

lorenzwalthert: Ok, I opened issue #128.

PR #120, opened 2017-08-09
Supporting more indention patterns

### LINKED ISSUE #108: Add indention support for multi-line method chaining in R6
Here's a simple example from [Advanced R](http://adv-r.hadley.nz/r6.html):

```
x$
  add(10)$
  add(10)$
  sum
```

lorenzwalthert:
See tests for examples. This closes #108.


R/token.R
```diff
@@ -10,6 +10,7 @@ token <- tribble(
   ">=" , "logical"     , "GE",
   "!=" , "logical"     , "NE",
   "==" , "logical"     , "EQ",
+  "="  , "assign_left" , "EQ_SUB",
```
krlmlr: What's the difference between `"EQ_SUB"` and `"EQ_ASSIGN"`? (Just curious.)

tests/testthat/example/out.R
```diff
@@ -115,11 +115,11 @@ test <- function() {
 
   nested(function_call(
   "in",
- multiple, lines))
+      multiple, lines))
 
   nested(
   function_call(with),
- many
+      many
```
krlmlr: This looks odd.

tests/testthat/example/out.R
```diff
@@ -130,7 +130,7 @@ test <- function() {
   difficult(nested(
     "function", call
   ),
- with, more, args
+    with, more, args
```
krlmlr: And this looks like we should force the `nested()` call on a new line (#118).

R/modify_pd.R
```diff
@@ -110,3 +126,21 @@ set_multi_line <- function(pd) {
...
   any(pd$multi_line, pd$lag_newlines > 0)
 }
+
+#' Update the newlines attribute
+#'
+#' As we work only with the `lag_newlines` attribute for setting the linebreaks,
+#' (R/rules-line_break.R) but we need `newlines` to determine
+#' whether or not to set `spaces` (R/rules-spacing.R), we have to update the
+#' attribute. We cannot simply use `dplyr::lead(pd$lag_newlines)` since we would
+#' loose information for the last token. `spaces` is left as is in
```
krlmlr: loose -> lose

R/token.R
lorenzwalthert:
```r
data_frame(a = 3))
# vs.
a = 3
```

tests/testthat/example/out.R
lorenzwalthert: I think it's fine, because the rule `add_space_after_comma()` was not checking for line break (which we need if we want to support raw indention) and this is the flat styling anyways. The token is now where it is in the in.R file.

tests/testthat/example/out.R
lorenzwalthert: Same here, this is flat styling. Maybe we should remove tests for it anyways when implementing #121.

R/modify_pd.R
lorenzwalthert: uups

PR #123, opened 2017-08-14
Redesign API

### LINKED ISSUE #121: remove relicts of flat styling completely
This includes changing the argument structures of some exported functions.

krlmlr:
This is a major API change. It makes closes #121 and makes working with top-level stylers much easier, in particuar through enabling the use of `...`. 
For example we change 
```r
style_text(text, flat = FALSE, transformers = get_transformers(flat = flat))
```
To
```r
style_text(text, ..., style = tidyverse_style, transformers = style(...))
```

All other top-level stylers such as `style_dir()`, `style_pkg()` etc. are also adapted to be consistent with the example above. Second argument is always `...`.

In addition, the function `create_style_guide()` is introduced, which will help useres to create the input corresponding to the `style` argument. Since we use `tibble::lst()`, all transformer functions (and not just the lists containing them) are now named too.


lorenzwalthert:
Ok, do you have any specific use case in mind already or is it a matter of principle? I wonder which user-control should then be passed via the transformers argument and what other options should be passed through via `...`. Some needs to be passed via the transformers since it influences them (like `strict`).

krlmlr:
I just wanted to write `style_pkg(scope = "spaces")` . There might be a better design, but this one looked simple enough.

lorenzwalthert:
Ok, cool, I did not know that `...` can also be used in that context.

lorenzwalthert:
@krlmlr can you review quickly? Can't assign the PR for review to you since you created it.

API
```diff
@@ -2,10 +2,9 @@
...
 
-get_transformers(flat = FALSE, ...)
-get_transformers_flat(strict = TRUE, start_comments_with_one_space = FALSE)
-get_transformers_nested(scope = "tokens", strict = TRUE, indent_by = 2, start_comments_with_one_space = FALSE)
-style_dir(path = ".", flat = FALSE, recursive = TRUE, transformers = get_transformers(flat = flat))
-style_file(path, flat = FALSE, transformers = get_transformers(flat = flat))
-style_pkg(pkg = ".", flat = FALSE, transformers = get_transformers(flat = flat))
-style_text(text, flat = FALSE, transformers = get_transformers(flat = flat))
+create_style_guide(filler, line_break, space, token, indention, use_raw_indention = FALSE)
+style_dir(path = ".", recursive = TRUE, ..., style = tidyverse_style, transformers = style(...))
```
krlmlr: Maybe also move `recursive = FALSE` past the dots?

lorenzwalthert: sounds good.

lorenzwalthert:
Also, I set defaults for all arguments of `create_style_guide()`, so `create_style_guide()` gives the *empty* style guide.

Issue #125, opened 2017-08-14
{ / } and line breaks in function calls


lorenzwalthert:
The tidyverse style guide states in section 2.4 that 

A `}` should always go on its own line, unless it’s followed by `else` or `)`. 

So that's why we implemented `remove_line_break_before_round_closing()` which is responsible for
the line breaks removed in [tikzDevice](https://github.com/yihui/tikzDevice/commit/1e4c965d4b3873966109bfc8b877ae7b8fbdfbeb). 

I personally think that's fine, but we may only include this rule if `strict = TRUE`.

Also it says that  a `{` should never go on its own line and should always be followed by a new line, so 
```r

tryCatch(
{
  ...
} 
```
from the above-referenced commit is in my eyes not compliant with the tidyverse style guide.

krlmlr:
The rules might need a refinement. In [this example](https://github.com/yihui/tikzDevice/commit/1e4c965d4b3873966109bfc8b877ae7b8fbdfbeb#diff-9162005f07b9d7b776be882dbf341544R298), the closing brace `}` belongs to an argument that is indented already, which is why indenting the closing parenthesis looks wrong:

```r
# good
fun(
  arg,
  {
    block
  }
)

fun(
  {
    block
  },
  {
    another_block
  }
)

# bad

fun({
  block
},
{
  another block
})

fun(
  arg,
  {
    another block
  })
```

lorenzwalthert:
I am not sure whether I understand you correctly. Why is 

```r
fun(
  arg,
  {
    block
  }
)
```
a good example with regard to the second rule cited? Shouldn't it be

```r
fun(
  arg, {
    block
  }
)
```
or even
```r
fun(
  arg, {
    block
})
```


Also

```r
fun(
  arg,
  {
    another block
  })
```
is bad, I agree, but I thought
```r

fun(
  arg, {
    another block
})
```
would be correct, so `}` is unindented since ) follows.

krlmlr:
I haven't found a specification for long function calls in the tidyverse guide. Maybe we could contribute one, that also covers complex function arguments that consists of whole blocks?

No need to hurry though, I'm fine with this limitation for now.

krlmlr:
Attempt to formulate a line-break rule:

> A multi-line argument to a function must start on a new line, unless it's the last argument (like `test_that(..., {})` *and* all other arguments fit on one line.

krlmlr:
Another potential exception, perhaps if *all* arguments are multi-line?

```r
[...]

PR #127, opened 2017-08-14
solve eq_sub indention in general

### LINKED ISSUE #126: Indent indices for token "EQ_SUB" are different.
The problem with the token "EQ_SUB" mentioned [here](https://github.com/yihui/tikzDevice/commit/d30b17e7011688c14984a6c7575c12da7075b2b8) arises because of the fact that expression "a = 3" and "b = 44" are not nested one level deeper, i.e. they are all direct sub-expressions of the root token:
``` r
styler:::create_tree("data_frame(a = 3, b = 44)")
#>                                              levelName
#> 1  ROOT (token: short_text [lag_newlines/spaces] {id})
#> 2   °--expr:  [0/0] {18}                              
#> 3       ¦--expr:  [0/0] {3}                           
#> 4       ¦   °--SYMBOL_FUNCTION_CALL: data_ [0/0] {1}  
#> 5       ¦--'(': ( [0/0] {2}                           
#> 6       ¦--SYMBOL_SUB: a [0/1] {4}                    
#> 7       ¦--EQ_SUB: = [0/1] {5}                        
#> 8       ¦--expr:  [0/0] {7}                           
#> 9       ¦   °--NUM_CONST: 3 [0/0] {6}                 
#> 10      ¦--',': , [0/1] {8}                           
#> 11      ¦--SYMBOL_SUB: b [0/1] {11}                   
#> 12      ¦--EQ_SUB: = [0/1] {12}                       
#> 13      ¦--expr:  [0/0] {14}                          
#> 14      ¦   °--NUM_CONST: 44 [0/0] {13}               
#> 15      °--')': ) [0/0] {15}
```

We used `compute_indent_indices()` so far to compute the indices that need indention. However, with all other operators such as braces or `+` `%>%`, the indention is always from the token on until the last or the penultimate token in that parse table. Here, it is only the next token, so we could go about it as follows:

```r
indent_eq_sub <- function(pd,
                          indent_by,
                          token = "EQ_SUB") {
  eq_sub <- which(pd$token == "EQ_SUB")
  if (length(eq_sub) == 0) return(pd)
  has_line_break <- which(pd$lag_newlines > 0)
  indent_indices <- intersect(eq_sub + 1, has_line_break)
  pd$indent[indent_indices] <- pd$indent[indent_indices] + indent_by
  pd
}
```

Alternatively, we had to modify `compute_indent_indices()`, which has already an argument `indent_last`. We could make that argument non-boolean and change it to `indent_up_to` and allow values "next", "penulatimate", "last".

lorenzwalthert:
closes #126


tests/testthat/indention_operators/eq_assign-out.R
```diff
@@ -0,0 +1,14 @@
+switch(engine,
+  pdftex = {
+    if (any) {
+      x
+    }
+  },
+  new = (
```
krlmlr: This pattern is rare enough, but what happens if we use `{}` instead of `()`?

lorenzwalthert: Like at line 2 you mean? It also indents by two spaces.

krlmlr: Yes, but what happens to the closing brace?

lorenzwalthert: It is unindented
```r
switch(engine,
  pdftex = {
    if (any) {
      x
    }
  },
  new = {
    2
})

{
  a <-
    3
}
```

krlmlr: Could you please add a `FIXME` that refers to #125?

Issue #134, opened 2017-08-18
Drop update_indention_ref_fun_call() from style_tidyverse() 


lorenzwalthert:
Plus adapt tests.

PR #135, opened 2017-08-18
Fix unindention

### LINKED ISSUE #131: tikzDevice glitch
https://github.com/yihui/tikzDevice/commit/6e8cef0e6ecbadc698c4625f763805965b343198##commitcomment-23693950

lorenzwalthert:
This fixes #131. The problem is that it is not really really fixing it because of the interaction with the rule `update_indention_ref_fun_call()` (the first argument of `switch` in the example is right after `(`). `update_indention_ref_fun_call `  is implemented currently but should not really be part of `style_tidyverse()`. Hence, we have to drop the rule and adapt the tests first. I opened #134 for that.


tests/testthat/unindention/unindent_multi_line-in.R
```diff
@@ -0,0 +1,10 @@
+call(
+  engine,
+  xetex = {
+         x
+       },
+       luatex= {
+       x
+       },{
+   xx
+  } )
```
krlmlr: Do you want to add a `FIXME` with a reference here?

lorenzwalthert: `FIXME` to leave a line blank after `}`you mean?

krlmlr: Yes.

PR #136, opened 2017-08-18
Non invasive line breaker

### LINKED ISSUE #103: provide non-invasive alternatives for manipulating spaces and line-breaks
* rename invasive functions such as `add_line_break_afer_curly_opening()` to what they really are
    namely `set_line_break_afer_curly_opening()` etc.
* add less invasive functions `add_*()`, following the existing convention in R/rules-spacing.R. 
* update `get_transformers*()` and friends correspondingly (that is, use the invasive functions only if 
  `strict = TRUE`.

lorenzwalthert:
Closes #103 
Must be rebased on master after #133 is merged since it contains one commit of it.


lorenzwalthert:
`set*` functions are used if `strict = TRUE`, `add*` functions are used otherwise.

PR #138, opened 2017-08-19
Public API tests

### LINKED ISSUE #58: add tests for top-level styler (style_src() etc.) with various options

lorenzwalthert:
This closes #58.


lorenzwalthert:
No, it does not. I think we have enough tests to test the styling, so I just wanted to make sure no error is thrown when these top-level styler functions are called.

PR #139, opened 2017-08-20
Adding exception handling for invalid parse data.

### LINKED ISSUE #137: Exception Handling
When working on #58, I tried this
``` r
styler::style_text("call(x), { } ")
#> Error in parse(text = text, keep.source = TRUE): <text>:1:8: unexpected ','
#> 1: call(x),
#>            ^
```
This is not particuarly useful but one can guess the root of the problem. If you want to style a file that contains such an invalid expression, you get
```r
style_dir("tests/testthat/public-api/xyzdir/subdir/")
Error in if (!any(ret)) { : missing value where TRUE/FALSE needed
In addition: Warning message:
 Show Traceback
 ...
```

How should we handle these cases? I don't have experience in error handling in packages so @krlmlr can you advise? Should we implement before submitting to CRAN? Also related to assertive tests etc.

lorenzwalthert:
This closes #137.

For `style_text()`
``` r
library("styler")
style_text("a 3")
#> Error in parsing the supplied text. Here is the parsing error message: 
#> Error in parse(text = text, keep.source = TRUE): <text>:1:3: unexpected numeric constant
#> 1: a 3
#>       ^
```And for `style_file()` and friends.
```r
path <- tempfile(fileext = ".R")
writeLines("a 3", path)
style_file(path)
#> Error in parsing file /private/var/folders/8l/fhmv3yj12kncddcjqwc19tkr0000gr/T/RtmpSOgZBl/file728e3c9f60a9.R Here is the parsing error message: 
#> Error in parse(text = text, keep.source = TRUE): <text>:1:3: unexpected numeric constant
#> 1: a 3
#>       ^
#> --------------------------------------------------------------------------------
```
As you can see from the source code, it will still style all parsable files if there are any.

That's how it looks now. Much better I think.


lorenzwalthert:
The only thing that bothers me a bit is that the return value is just a message, not an error, but I think there is no way around it if we want to continue with other files, right?

R/transform.R
```diff
@@ -4,11 +4,17 @@
...
 #' @param files A character vector with paths to the file that should be
 #'   transformed.
+#' @param assert_code Boolean to indicate whether checking for parsing errors
+#'   should be done prior to styling.
 #' @inheritParams make_transformer
 #' @return A logical value that indicates whether or not any file was changed is
 #'   returned invisibly. If files were changed, the user is informed to
 #'   carefully inspect the changes via a message sent to the console.
-transform_files <- function(files, transformers) {
+transform_files <- function(files, transformers, assert_code = TRUE) {
```
krlmlr: I'm not sure we need a new argument here. Rather the behavior should always be reasonable if parsing fails. If necessary, we could change `transform_lines_enc()` to account for this.

lorenzwalthert: Ok, so we can remove the argument then. I am not sure whether it makes sense to implement the error handling in `transform_lines_enc()` since it has already a error handling but I don't know why it does only give a warning instead of an error.
``` r
library("styler")
style_text("a 3")
#> Error in parse(text = text, keep.source = TRUE): <text>:1:3: unexpected numeric constant
#> 1: a 3
#>       ^
path <- tempfile(fileext = ".R")
writeLines("a 3", path)
style_file(path)
#> Warning: When processing file6b1f0f56be.R: <text>:1:3: unexpected numeric constant
#> 1: a 3
#>       ^
#> Error in if (any(changed)) {: missing value where TRUE/FALSE needed
style_dir(dirname(path))
#> Warning: When processing ./file6b1f0f56be.R: <text>:1:3: unexpected numeric constant
#> 1: a 3
#>       ^
#> Error in if (any(changed)) {: missing value where TRUE/FALSE needed
```

krlmlr: I forgot how `transform_lines_enc()` works, the error handling seems sufficient already. Do you need `any(changed, na.rm = TRUE)` ?

lorenzwalthert: Yes, with `any(changed, na.rm = TRUE)`, we only get the first warning message. So I will undo my error handling implementation and adapt the tests.

lorenzwalthert: In my eyes, there is one more tweak necessary in the `utf8` package. Unfortunately, it is not visible in the reprex above because `reprex` strangely does not show all warnings I get in the console. I opened [#15](https://github.com/krlmlr/utf8/issues/15) and submitted a [PR](https://github.com/krlmlr/utf8/pull/16).

lorenzwalthert: Also, do you need to submit `utf8` to CRAN in order to submit `styler` to CRAN, right?

krlmlr: Yes, I'm about to submit utf8.

lorenzwalthert:
Ok, I now reverted everything but kept the tests, so error handling is now done by `utf8`.

PR #142, opened 2017-08-21
Fix indention

### LINKED ISSUE #141: Indention based on operators only checks first potential trigger
Indention based on operators only checks first potential trigger.
``` r
test <- c(
  "1 + (", 
  "3", 
  ") %>%", 
  "x()"
)
cat(styler::style_text(test), sep = "\n")
#> 1 + (
#>   3
#> ) %>%
#> x()
```

`x()`  is not indented since it is only checked whether `+` causes indention (which is not the case since a multi-line expression follows and `(` will cause indention within the child). However, we need to check for all potential token that are triggers  (`+` and `%>%` here) and then indent on the first one for which we need indention.

lorenzwalthert:
Closes #141


R/modify_pd.R
```diff
@@ -80,30 +80,41 @@ indent_without_paren <- function(pd, indent_by = 2) {
...
+#'   for which it should be checked whether it should trigger indention.
+#' @return Returns `TRUE` if indention is needed, `FALSE` otherwise.
 #' @return `TRUE` if indention is needed, `FALSE` otherwise.
-needs_indention <- function(pd, opening) {
-  if (is.na(opening)) return(FALSE)
+needs_indention_one <- function(pd, potential_trigger) {
   before_first_break <- which(pd$lag_newlines > 0)[1] - 1
   if (is.na(before_first_break)) return(FALSE)
-  !any(pd$multi_line[opening:before_first_break])
+  !any(pd$multi_line[potential_trigger:before_first_break])
```
krlmlr: `seq2()`?

R/modify_pd.R
```diff
@@ -80,30 +80,41 @@ indent_without_paren <- function(pd, indent_by = 2) {
...
-  opening <- which(pd$token %in% token)[1]
-  if (!needs_indention(pd, opening)) return(numeric(0))
-  start <- opening + 1
+  potential_triggers <- which(pd$token %in% token)
+  needs_indention <- needs_indention(pd, potential_triggers)
+  trigger <- potential_triggers[needs_indention][1]
+  if (is.na(trigger)) return(numeric(0))
+  start <- trigger + 1
   stop <- npd - ifelse(indent_last, 0, 1)
   which(between(seq_len(npd), start, stop))
```
krlmlr: Could `seq2()` work here, too?

R/modify_pd.R
lorenzwalthert: Right.

R/modify_pd.R
lorenzwalthert: True.

PR #143, opened 2017-08-21
Addin to style highlighted region

### LINKED ISSUE #107: RStudio Addin to style highlighted region of file
Like `style_active_file()` but for the highlighted region only.

lorenzwalthert:
This closes #107. It seems a bit complicated because of the fact that you can highlight also just parts of lines. Did not really know how we could add formal testing to that. One way would be to make `context` an argument of `style_active_region()` with default `find_active_context()` so we can supply some file region for testing purposes.


lorenzwalthert:
Yes, I thought about that too. I think it gives feedback, e.g.  "No files changed" appears on the console but I think the problem is also that the error handling is not very good. Same with the other Addins, although I think the "style active file"-Addin is the most important one. The reason I have not explored interactive Addins is that I can't do that as part of Milestone CRAN due to time constraints. Should we leave it as is or should we keep all Addins for a future CRAN release?

krlmlr:
"Interactive" just means that it runs in the active R session, and blocks it -- a simple change to `inst/rstudio/addins.dcf`, not a GUI or anything.

lorenzwalthert:
Ok, I see. Let's do that then.

lorenzwalthert:
It actually does not work the way I thought because we use the RStudio API to get the active document. If we set `interactive = TRUE`, the Addin is executed in the console and can't get the active document context. So this mode would only work as a short cut for `style_pkg()` but not for styling the active file or the highlighted region on the active file.

krlmlr:
`rstudioapi::getActiveDocumentContext()` works for me when I run it from the console. What's the difference to an interactive add-in?

lorenzwalthert:
Oh now it works, had to restart RStudio also, not just rebuild the package and restart R. With regard to tests, do you think we should make `context` an argument as stated above.

krlmlr:
You could implement:

```r
style_active_file <- function() {
  file <- get_active_context()$path
  style_file(file, style = tidyverse_style)
}

get_rstudio_context <- function() {
  rstudioapi::getActiveDocumentContext()
}
```

and then in the test ask

```r
test_that("...", {
  result <- mockr::with_mock(
    get_rstudio_context = function() structure(list(...), class = "document_context"),
    style_active_file()
  )
  expect...(result, ...)
})
```

lorenzwalthert:
I tried to use `mockr` but when running [interactively](https://github.com/krlmlr/styler/pull/143/files#diff-42e43bd7ad8f646ad9898959d1e1e60cR20) it gave me an error with `mockr::with_mock()`, but not with `testthat::with_mock()`. When runnign via `devtools::test(f = "public"), both did not give an error. I used testhat for now, maybe @krlmrl can have a look why that happends? Also, I did not mock `get_rstudio_context()` but `find_active_context()` since this function is the one that getscalled by the function we want to test and I was not sure whether it was possible to mock functions that are not directly called by the function we want to test.

R/addins.R
```diff
@@ -0,0 +1,111 @@
+#' Style the active file
+#'
+#' Helper function for RStudio Addin.
+style_active_file <- function() {
+  file <- rstudioapi::getActiveDocumentContext()$path
```
krlmlr: Do you need to call `find_active_context()` here to allow mocking?

R/addins.R
```diff
@@ -0,0 +1,111 @@
...
+
+#' Helper to extract and preprocess relevant attributes from
+#' [rstudioapi::getActiveDocumentContext()].
+#'
+find_active_context <- function() {
+  context <- rstudioapi::getActiveDocumentContext()
+  path <- context$path
+  start <- context$selection[[1]]$range$start
+  end <- context$selection[[1]]$range$end
+  if (all(start == end)) return()
```
krlmlr: ... and then return a list with only one component here?

tests/testthat/test-public_api.R
```diff
@@ -17,3 +17,37 @@ test_that("styler can style file", {
 })
 
 # style_active_file() must be tested manually.
+test_that("...", {
```
krlmlr: ...?

tests/testthat/test-public_api.R
```diff
@@ -17,3 +17,37 @@ test_that("styler can style file", {
...
+  before_styling <- utf8::read_lines_enc(paste0(base, "/xyzaddin/addin_region-in.R"))
+  reference <- utf8::read_lines_enc(paste0(base, "/xyzaddin/addin_region-out.R"))
+  result <- expect_error(
+    testthat::with_mock(
```
krlmlr: `mockr::with_mock()` should really work here, because `testthat::with_mock()` might not at some point. Can you please simplify the code (e.g., extract the hard-coded `"document_context"` structure to a variable) and try again? For troubleshooting the error message is helpful.

R/addins.R
```diff
@@ -0,0 +1,111 @@
...
+
+  merged_text <- append(neighbourhood$body, styled_lines, context$start[1] - 1)
+  utf8::write_lines_enc(merged_text, context$path)
+}
+
+#' Helper to extract and preprocess relevant attributes from
+#' [rstudioapi::getActiveDocumentContext()].
+#'
+find_active_context <- function() {
+  context <- rstudioapi::getActiveDocumentContext()
```
krlmlr: I thought we could just create a simple forwarder to `rstudioapi::getActiveDocumentContext()` in our package, and mock this forwarder:

```r
get_rstudio_context <- function() {
  rstudioapi::getActiveDocumentContext()
}
```

mockr still works if functions are called indirectly from our package (it patches all functions in our package).

R/addins.R
lorenzwalthert: If we want I guess so. So far, only style_actige_region is tested but we can add a test here too to increase coverage.

R/addins.R
lorenzwalthert: Ok yes I see. That's what I a say unsure about (see comment above).

tests/testthat/test-public_api.R
lorenzwalthert: `mockr::with_mock()` works using `devtools::test()` but not interactively while `testthat::with_mock()` works in both sistuations, which is why I have not used `mockr` yet. Can you have a look at why that is the case?

What do you mean with extracting hard-coded `"document_context"` structure to a variable?  Like this?

```r
doc_cont <- structure(...)
testthat("...", {
  result <- mockr::with_mock(
    get_rstudio_context = function() doc_cont, 
    ...
```
Unfortunately, there is no way to create the structure with a function call directly.

R/addins.R
lorenzwalthert: I think this should rather be `stop("No region selected")` instead of `return()`.

tests/testthat/test-public_api.R
lorenzwalthert: oops

tests/testthat/test-public_api.R
krlmlr: Yes, something like this code.

You could try `mockr::with_mock(..., envir = asNamespace("styler"))` .

R/addins.R
krlmlr: Even if we want to call this for getting the active file?

krlmlr:
Do you need more input here?

R/addins.R
lorenzwalthert: No then not. But currently, we don't use find_active_context() for styling files. Should we? If yes, then the error message should not be implemented within the function find_active_context, I agree.

lorenzwalthert: It's currently only used for styling regions within files.

lorenzwalthert: So you mean we add an argument whole_file to to find_active_context() so it can be used for styling files as well as for highlighted regions?

tests/testthat/test-public_api.R
lorenzwalthert: Ok, so assign the variable elsewhere in the name space? Or should leave the declaration where it is?

R/addins.R
lorenzwalthert: Ok, I adapted it so it is used in both cases.

lorenzwalthert:
Can you review the updated PR?

R/styler.R
```diff
@@ -1,8 +1,19 @@
 #' Non-invasive pretty printing of R code
 #'
 #' styler allows you to format .R files, packages or entire R source trees
-#'  according to a style guide. See the INDEX for more information.
-#'
+#'  according to a style guide.
+#'  The fo,lowing functions can be used for styling:
```
krlmlr: Typo?

tests/testthat/test-public_api.R
```diff
@@ -19,5 +19,36 @@ test_that("styler can style file", {
...
+                    class = "document_position")), .Names = c("start", "end"),
+    class = "document_range"), text = "1+1 "), .Names = c("range", "text"))),
+    .Names = "", class = "document_selection")),
+  .Names = c("id", "path", "contents", "selection"), class = "document_context"
+)
 
+# style_active_file() must be tested manually.
+test_that("styling active region works", {
 
+  before_styling <- utf8::read_lines_enc(paste0(base, "/xyzaddin/addin_region-in.R"))
```
krlmlr: Maybe just copy to a temporary directory?

```r
dir <- tempfile("styler")
dir.create(dir)
file.copy(".../addin_region-out.R", dir)
```

The current approach seems slightly brittle if the tests aborts before the original contents have been written.

R/addins.R
```diff
@@ -0,0 +1,115 @@
...
+  )
+  styled_lines <- style_text(lines_to_style)
+  styled_lines
+}
+
+#' Complete styled expression with unstyled fraction on start / end line
+#'
+#' Essentially adding the parts before the highlighted area on the line where
+#' the higlighted area starts to the styled expression. Also, after the styled
+#' expression, the remainer of that line that was not styled is added.
```
krlmlr: remainder?

tests/testthat/test-public_api.R
lorenzwalthert: Ok, right. So if I understand you right, we copy the `*-in.R*` to the temporary directory, style it there and compare it with the `-*out`. in the permanent directory?

R/addins.R
lorenzwalthert: yes

R/styler.R
lorenzwalthert: yes

tests/testthat/test-public_api.R
krlmlr: Yes.

README.md
```diff
@@ -23,7 +23,7 @@ There are a few variants of `style_text()`:
 -   `style_file()` styles a single .R file.
 -   `style_dir()` styles all .R files in a directory.
 -   `style_pkg()` styles the source files of an R package.
--   An RStudio Addin that styles the active file .R file
+-   RStudio Addins for styling the active file, styling current package and styling the highlighted code region.
```
krlmlr: the current package?

tests/testthat/test-public_api.R
```diff
@@ -21,3 +21,39 @@ test_that("styler does not return error when there is no file to style", {
...
+    end = structure(c(1, 21), .Names = c("row", "column"),
+                    class = "document_position")), .Names = c("start", "end"),
+    class = "document_range"), text = "1+1 "), .Names = c("range", "text"))),
+    .Names = "", class = "document_selection")),
+  .Names = c("id", "path", "contents", "selection"), class = "document_context"
+)
+
+# style_active_file() must be tested manually.
+test_that("styling active region works", {
+  reference <- utf8::read_lines_enc( paste0(base, "/xyzaddin/addin_region-out.R"))
```
krlmlr: `file.path()` better explains your intentions. Even better, a small function

```r
testthat_file <- function(...) {
  file.path(rprojroot::find_testthat_root_file("public-api"), ...)
}
```

instead of the `base` variable. Here and elsewhere.

tests/testthat/test-public_api.R
```diff
@@ -21,3 +21,39 @@ test_that("styler does not return error when there is no file to style", {
 })
 
 
+##  ............................................................................
+
+dir <- tempfile("styler")
```
krlmlr: All these variables seem to fit better in the `test_that()` block below.

tests/testthat/test-public_api.R
```diff
@@ -21,3 +21,39 @@ test_that("styler does not return error when there is no file to style", {
 })
```
krlmlr: Would the preceding tests change the original files?

R/addins.R
```diff
@@ -0,0 +1,115 @@
...
+  ind_to_style <- seq2(context$start[1], context$end[1])
+  lines_to_style <- text[ind_to_style]
+  lines_not_to_style <- text[setdiff(seq_along(text), ind_to_style)]
+  fract_of_line_start <- substring(lines_to_style[1], 1, context$start[2] - 1)
+  last <- length(lines_to_style)
+  fract_of_line_end <- substring(lines_to_style[last], context$end[2])
+  list(
+    start = fract_of_line_start,
+    end = fract_of_line_end,
+    body = lines_not_to_style
```
krlmlr: `body` seems a misnomer if it's

> the lines that are not overlapping with the highlighted text.

README.md
lorenzwalthert: Well, the package which has its root in the current working directory (if any). Could not find a better formulation for it. Any ideas?

tests/testthat/test-public_api.R
lorenzwalthert: No, they should not, because the input is already styled. The tests are only to check whether the call returns an error or not. It's a rather superficial test.

README.md
krlmlr: styling current package -> styling the current package

tests/testthat/test-public_api.R
krlmlr: These files will then need an update whenever the style guide changes, but that's fine with me.

README.md
lorenzwalthert: Sure.

krlmlr: ?

R/serialized_tests.R
```diff
@@ -209,3 +209,10 @@ style_op <- function(text) {
   transformed_text
 
 }
+
+#' Create the path to a test that file
+#' @param ... Arguments passed to [file.path()] to construct the path after
+#'   ".../tests/testthat/"
+testthat_file <- function(...) {
```
krlmlr: This might be better suited in a `tests/testthat/helper-*.R` file.

lorenzwalthert: 1. ups, forgot to knit `.Rmd` to `md`.
2. Ok, but then we won't get `roxygen2` documentation...

krlmlr: Yes, I forgot. Let's keep it here.

lorenzwalthert:
@krlmlr Do you want to have a final look? I think now we got them all 😄

Issue #16, opened 2017-06-01
Renaming


lorenzwalthert:
# Naming of data
At the moment, for the parse table returned by utils::getParseData and 
processed through various functions, we use many different names in different 
functions for essentially this (flat) table:

* pd (all rules in rules.R)
* parse_data (serialize_parse_data)
* parse_data_with_ws (serialize_parse_data_flat)
* data (create_filler)

I think it would simplify things to just use one variable name throughout. I
suggest parse_data. We could use `parse_data_flt` for the flat version and and 
`parse_data_nstd` for the nested version for situations where the processing
differs.

# Naming of functions
As far as I understand, the terminology around `ws` is more of a legacy thing, i.e. in the functions `compute_parse_data_flat_with_ws` or `add_ws_to_parse_data` since it is not really just white space information that is affected by these functions, but rather a bunch of preprocessing Hence, I would suggest to rename the latter to `pre_process_parse_data` and the former to `compute_parse_data_flt` for now.

krlmlr:
A consistent name is a good idea. Can we simply use `nested` and `flat`, or perhaps standardize on `pd` for parse data and use this as a prefix or suffix?

lorenzwalthert:
Ok yes, let us just use `pd` then and use `flat` and `nested` as  suffix.

lorenzwalthert:
Reference: https://github.com/krlmlr/styler/pull/146#discussion_r134675202.

lorenzwalthert:
So as stated in https://github.com/krlmlr/styler/pull/146#discussion_r134675202, should we use `nest` for "a parse table at one level of nesting, which always represents a complete expression", and `pd_nested` for the whole nested structure, i.e. the parse table at the top level of nesting?

lorenzwalthert:
This includes documentation also obviously.

krlmlr:
Yes, we can keep the terms "parse data" and "nested parse data" for simplicity. The latter consists of multiple interlinked nests, this is the only new term.

lorenzwalthert:
Reference: #132

lorenzwalthert:
Reference: #132.

lorenzwalthert:
# Renaming of files

* `modify_pd.R` -> `indent.R`
* `nested.R` -> `nest.R`
* `ws.R` -> `ui.R`
* `style_md.R` -> `transform-code.R`.
* `transform.R` -> `transform-files.R`

krlmlr:
[...]

PR #146, opened 2017-08-22
Vignette on customizing styler 

### LINKED ISSUE #10: Extensibility: Third-party style guide support
At it’s core, a style guide is a collection of transformer functions that fulfill certain formal requirements. 
It would be nice if styler could support style guides defined by third parties. 
In the following, I lay out a potential mechanism for how such a third-party support could look like. This were my first thoughts on the topic. Maybe there are better approaches to it.

Styler could provide a function to import style guides  and save them in the local package directory of styler. People who want to create style guides could provide the transformer functions for example in a GitHub repository. The style guide that should be used to format a package’s source code (and potentially further options) could then be specified in a file in the root directory of that package (e.g. in styler.yml). Styler would then use the the information from this .yml file, source the corresponding .R files from the styler package directory and apply them (and potentially further options) to the source code to be formatted. The formal requirements of style guides had to be defined (i.e. before it can be imported). Potential requirements: Style guide must not yield an error when applied to a test source code. Transformer functions must have follow certain naming conventions, must be documented ect.  Provide detailed instructions how to create / test style guides etc.

lorenzwalthert:
When merged, this should close #10 I think.

### DIFF
--- a/vignettes/customizing_styler.Rmd
+++ b/vignettes/customizing_styler.Rmd
@@ -21,14 +21,19 @@ There are three major steps that styler performs in order to style code:
    this a nested parse table. You can learn more about how that is done
    exactly in the vignettes "Data Structures" and "Manipulating the nested 
    parse table".
-2. Apply transformer functions at each level of the nested parse table. In 
-   particular, we create two columns `lag_newlines` and `spaces`, which contain
+2. Apply transformer functions at each level of the nested parse table. We 
+   use a visitor approach, i.e. a function that takes functions as arguments and
+   applies them to every level of nesting. You can find out more about it on
+   the help file for `visit`. Note that the function is not exported by styler.
+   The visitor will take care of applying the functions on every 
+   level of nesting - and we can supply transformer functions that operate on 
+   one level of nesting. In the sequel, we use the term *nest* to refer to 
+   such a parse table at one level of nesting. A *nest* always represents a 
+   complete expression. Before we apply the transformers, we have to initialize
+   two columns `lag_newlines` and `spaces`, which contain
    the number of line breaks before the token and the number of spaces 
    after the token. These will be the columns that most of our transformer 
-   functions will modify. To go through the whole nested structure and apply
-   the transformer functions repeatedly on every level of nesting, we use
-   a visitor approach, i.e. a function that takes functions as arguments and
-   applies them to every level of nesting. 
+   functions will modify. 
 3. Serialize the nested parse table, that is, extract the terminal tokens from 
    the nested parse table and add spaces and line breaks between them as 
    specified in the nested parse table. 
@@ -57,10 +62,10 @@ tidyverse_style()$space$remove_space_after_opening_paren
 ```
 
 As the name says, this function removes spaces after the opening parenthesis. But
-how? Its input is a flat parse table. Since the visitor functions `pre_visit()`
-and `post_visit()` will go through all levels of nesting, we don't need to care
-about that - we just need a function that can be applied to a flat parse table,
-i.e at one level of nesting.
+how?
+Its input is a *nest*. Since the visitor will go through all levels of nesting, 
+we just need a function that can be applied to a *nest*, that is, to a parse 
+table at one level of nesting.
 We can compute the nested parse table and look at one of the levels of nesting
 that is interesting for us (more on the data structure in the vignettes 
 "Data structures" and "Manipulating the parse table"):
--- a/vignettes/customizing_styler.Rmd
+++ b/vignettes/customizing_styler.Rmd
@@ -1,0 +1,268 @@
+---
+title: "Customizing styler"
+author: "Lorenz Walthert"
+date: "8/10/2017"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Customizing styler}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+This vignette gives a high-level overview about how styler works and how you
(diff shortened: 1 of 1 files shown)


vignettes/customizing_styler.Rmd
```diff
@@ -0,0 +1,268 @@
...
+`use_raw_indention` is not a function, it is just an option. All transformer
+functions have a similar structure. Let's pick one and look at it:
+```{r}
+tidyverse_style()$space$remove_space_after_opening_paren
+```
+
+As the name says, this function removes spaces after the opening parenthesis. But
+how? Its input is a flat parse table. Since the visitor functions `pre_visit()`
+and `post_visit()` will go through all levels of nesting, we don't need to care
+about that - we just need a function that can be applied to a flat parse table,
```
krlmlr: The parse table isn't actually flat, it's just one level of nesting. I'd prefer a different name here, partly to avoid confusion with the previously used "flat" approach. Each level of nesting corresponds to a complete expression, maybe the choice of name could reflect this.

vignettes/customizing_styler.Rmd
```diff
@@ -0,0 +1,268 @@
...
+should be aware of, which are described in the next section.
+
+# Implementation details
+
+For both spaces and line break information in the nested parse table, we use
+four attributes in total: `lag_newlines`, `newlines`, `spaces`, `lag_spaces`.
+`lag_spaces` is created from `spaces` only just before the parse table is
+serialized, so it is not relevant for manipulating the parse table as
+described above. These columns are to some degree redundant, but with just lag
+or lead, we would loose information on the first or the last element
```
krlmlr: lose

vignettes/customizing_styler.Rmd
```diff
@@ -0,0 +1,268 @@
...
+
+```{r, eval = FALSE}
+a <- function() # comments should remain EOL { 
+  3
+}
+```
+
+The easiest way of taking care of that is not applying the rule
+if there is a comment, which can be checked for within your transformer 
+function.
```
krlmlr: Maybe enhance the example code to support what you said in the "Notes"?

vignettes/customizing_styler.Rmd
```diff
@@ -0,0 +1,268 @@
...
+You can also look it up in the function that applies the transformers: 
+`apply_transformers()`:
+```{r}
+styler:::apply_transformers
+```
+
+This means that the order of the styling is clearly defined and it is for 
+example not possible to modify line breaks based on spacing, because spacing
+will be set after line breaks are set. Do not rely on the column `col1`,
+`col2`, `line1` and `line2` in the parse table in any of your function since
```
krlmlr: We don't need to mention these columns here if we get rid of them early enough in the process, maybe after computing spaces and newlines?

lorenzwalthert: We can get rid of them yes, but we need them in the end again for token-dependent indention (function `enrich_terminals()`), so I did not actively remove them in between. Whatever you prefer is fine for me.

vignettes/customizing_styler.Rmd
lorenzwalthert: Yes, I also did not really like the formulation. So maybe just "[...] Its input is a parse table at one level of nesting, which always represents a complete expression [...]". So we may want to change the naming convention of all source code also at some point? So replacing `pd_flat` simply with `pd`?

vignettes/customizing_styler.Rmd
krlmlr: I think removing obsolete data is the safer strategy, even if we recreate the same columns with up-to-date values later.

vignettes/customizing_styler.Rmd
krlmlr: Yes, we just need a short unambiguous name for "a parse table at one level of nesting, which always represents a complete expression" ;-) We can even formally introduce it in this vignette when describing the AST.

lorenzwalthert: Well, technically it is still a nested parse table (potentially with no child). So it would not be wrong to just say "[...] Its input is a nested parse table and we want to apply transformer functions to the top level of that nested parse table [..]". Alternatively, we can call it PATOLON (Parse table AT One Level Of Nesting). :-)

vignettes/customizing_styler.Rmd
lorenzwalthert: I opened #147

lorenzwalthert: I opened #147.

lorenzwalthert: So not sure what priority that has since it is not totally straight forward (see #147).

krlmlr: Low prio.

lorenzwalthert: So then I guess we just leave it for now.

vignettes/customizing_styler.Rmd
lorenzwalthert: What do you prefer?

krlmlr: How about "nest"?

krlmlr: With a proper definition of course.

lorenzwalthert: Ok, so I will change it just in this vignette then if that's ok.

krlmlr: It's ok for a start, eventually the documentation and the code should use consistent terminology.

lorenzwalthert: Yes, I reopened #16.

PR #149, opened 2017-08-22
Don't reindent function calls

### LINKED ISSUE #134: Drop update_indention_ref_fun_call() from style_tidyverse()

lorenzwalthert:
Removes function `update_indention_ref_fun_call()` from `style_tidyverse()` and adds functions to (1) add line break after opening brace of function calls if the call is multi-line and (2) adds a line break before closing brace if the call is multi-line. Closes #134 .


R/rules-line_break.R
```diff
@@ -42,8 +42,51 @@ remove_line_break_before_round_closing <- function(pd) {
...
+NULL
+
+#' @describeIn set_line_break_if_call_is_multi_line Sets line break after
+#'   opening parenthesis.
+set_line_break_after_opening_if_call_is_multi_line <-
+  function(pd,
+           except_token = NULL) {
+  if (!is_function_call(pd)) return(pd)
+  npd <- nrow(pd)
+  if (npd == 3) {
```
krlmlr: Why is this shortcut important?

tests/testthat/indention_operators/eq_assign-out.R
```diff
@@ -1,12 +1,14 @@
...
       x
     }
   },
-       new = (
-         2
-     )) # FIXME: Add line break between parens (#125)
+  new = (
+    2
+  )
+) # FIXME: Add line break between parens (#125)
```
krlmlr: Do we still need this FIXME?

tests/testthat/indention_token_dependent/token_dependent_complex-out.R
```diff
@@ -2,30 +2,39 @@ call(call(
...
-call(1,
-     call2(3, 4, call(3,
-                      4, call(5, 6, call(
-                        2
-                      )
-                      )
-     )
-     )
+  1,
+  2, c(
```
krlmlr: We seem to need a rule that requires a newline before multi-line function arguments. Do we have a self-contained writeup for the new function call rules? Maybe open a PR to the tidystyle repo so that it's not lost?

R/rules-line_break.R
lorenzwalthert: Because `seq2(3, npd - 1)` would throw an error otherwise. `npd = 3` is the case  of empty function calls like`call()`. Just making sure there this does not happen:
```r
call(
)
```
We could circumvent that using `seq2(3, npd)` instead, can't actually remember why I did not do that.

tests/testthat/indention_operators/eq_assign-out.R
lorenzwalthert: Nope.

R/rules-line_break.R
krlmlr:
``` r
rlang::seq2(3, 2)
#> integer(0)
```

lorenzwalthert: Uups, no error then, sorry. Buth by dropping the if condition, we cannot solve the problem because if `npd = 3`, and
```r
call(
)
```
we want to remove the line break before the closing parent, which we can't using `seq2(a, b)` with `a>b`. Also, my suggestion to use `seq(3, npd)` does not make sense since 
```r
call(2
)
```
would then be classified as multi-line, which I think does not make sense. We should not take the closing parenthesis into account. What do you think? Not sure whether there is another way to do that then.

R/rules-line_break.R
```diff
@@ -42,8 +42,51 @@ remove_line_break_before_round_closing <- function(pd) {
...
+    pd$lag_newlines[3] <- 0L
+    return(pd)
+  }
+  is_multi_line <- any(pd$lag_newlines[seq2(3, npd - 1)] > 0)
+  if (!is_multi_line) return(pd)
+  exeption_pos <- which(pd$token %in% except_token)
```
krlmlr: exept -> except

R/rules-line_break.R
krlmlr: It looks clearer to me to remove the shortcut and add a separate rule to remove new lines for empty calls. This new rule seems to be very low priority.

lorenzwalthert: Ok, sure.

tests/testthat/indention_token_dependent/token_dependent_complex-out.R
lorenzwalthert: Right, we don't have such a rule yet. So if I understand you correctly, you think 
```r
call(
  2, 
  3, c(
    2, 
    3
))
```

Should be styled as 
```r
call(
  3, 
  3, 
  c(
    very_long_such, 
    that_we_cant_put_it_on_one_line
  )
)
```
If we can't put the whole sub-expression `c(...)` on one line?

krlmlr: Yes.

PR #151, opened 2017-08-23
Remove flat relicts - part 2

### LINKED ISSUE #121: remove relicts of flat styling completely

lorenzwalthert:
Closes #121. As noted in https://github.com/krlmlr/styler/issues/121#issuecomment-324123183, code in vignettes that is now out-dated was replaced with rendered `md` before the respective code was removed.


lorenzwalthert:
Ok, yes right we wanted to mention it. What should the sub-directory be called then? I think `old` is a bad name... Since they were related to my initial GSOC Application, we could just call it `gsoc_proposal`?

krlmlr:
Sounds good.

lorenzwalthert:
Getting up to 93% coverage 😄. I think we should add a coverage badge to the README haha.

PR #152, opened 2017-08-23
No line break after switch and friends

lorenzwalthert:
As requested in https://github.com/krlmlr/rlang/pull/1#discussion_r134582400, it is now possible to not break the line in function calls where the text of the function is in `except_text_before`. I.e.
```r
switch(x, 
  z, 
  y
)
```
Will not be turned into 
```r
switch(
  x, 
  z, 
  y
)
```

Since there is no such rule in the tidyverse style guide yet, I suggest opening an issue or PR before merging this. 
Also, I am not sure whether the names of the functions should be an argument to `tidyverse_style()`. For now, they are not. `case_when()` might also be such a function that should be treated as an exception.


lorenzwalthert:
Sure. Should I submit an issue to the tidyverse style guide?

krlmlr:
Yes, please.

lorenzwalthert:
Do you think the text tokens be arguments in `tidyverse_style` as outlined above?

R/style_guides.R
```diff
@@ -75,7 +75,8 @@ tidyverse_style <- function(scope = "tokens",
         add_line_break_before_curly_closing,
       partial(
         set_line_break_after_opening_if_call_is_multi_line,
-        except_token = "COMMENT"
+        except_token_after = "COMMENT",
+        except_text_before = c("switch", "ifelse", "if_else")
```
krlmlr: Maybe expose an argument to `tidyverse_style()` that allows adding more exceptions?

R/style_guides.R
```diff
@@ -75,7 +75,8 @@ tidyverse_style <- function(scope = "tokens",
         add_line_break_before_curly_closing,
       partial(
         set_line_break_after_opening_if_call_is_multi_line,
-        except_token = "COMMENT"
+        except_token_after = "COMMENT",
```
krlmlr: Side note: I'm not sure about the `scope` argument to this function, maybe `scope` should be an argument of the top-level `style_()` functions? The default values of the arguments to this function here are useful, we can easily change that later.

R/style_guides.R
krlmlr: (Basically answering "yes" to your earlier question, if I understood correctly.)

R/style_guides.R
lorenzwalthert: Well you can do `style_text("a <- 3", scope = "tokens")`, so it is a top-level argument trough `...`

R/style_guides.R
lorenzwalthert: Ok, what how does that relate to https://github.com/tidyverse/style/issues/39? It seems to be about named arguments according to the answer there, and not to function names anyways.

R/style_guides.R
krlmlr: We can keep it for now and decide later.

R/style_guides.R
krlmlr: Let's merge this the way it is now and discuss rules that involve named/unnamed arguments separately.

PR #172, opened 2017-08-30
adapt line break rules.

### LINKED ISSUE #171: Don't force line break removal after opening paren in function call if next token is a comment
That should not happen.
``` r
cat(styler::style_text(c(
  "list( ", 
  "#", 
  "comment", 
  ")"
)), sep = "\n")
#> list( #
#>   comment
#> )
```

lorenzwalthert:
Closes #171.


tests/testthat/line_breaks_fun_call/line_breaks_and_comments-in.R
```diff
@@ -0,0 +1,15 @@
...
+  am
+)
+
+call(
+  # comment
+  am
+)
+
+call(am # comment
+) # FIXME: maybe move ')' up and switch position with the comment
```
krlmlr: Better push `am` to its own line then.

lorenzwalthert: Ok. So that means when checking whether an expression is multi-line, we also consider the closing `)`. So in general,
```r
call(a
)
```
Becomes
```r
call(
  a
)
```
Instead of 
```r
call(a)
```

krlmlr: I'd say only if a comment is involved, but both solutions work.

lorenzwalthert: Ok, I prefer the former then.

lorenzwalthert:
Because the `multi_line` attribute is not available at that stage, we can only check via `lag_newlines` whether an expression spans multiple lines. That means for example that 

```r
call(
  1,
  call2(3, 4, call(
    3,
    4, call(5, 6, call(
      2
    )
    )
  )
  )
)
```
Will become 

```r
call(
  1,
  call2(3, 4, call(
    3,
    4, call(5, 6, call(
      2
    ))
  ))
)
```
I think it's fine, since if you open multiple parentheses on one line, why not also closing them on the same line? It also plays nicely with indention. @krlmlr what do you think? Alternatively, we had to leave
```r
call(a
)
```
As is.

PR #182, opened 2017-09-05
Remove dplyr dependency

jimhester:
This aims to remove the dplyr dependency.

Currently this removes all dplyr dependencies but `left_join()` and `bind_rows()`. Both are slightly trick to remove, `left_join()` because we are joining nested dataframes, and `merge()` and `cbind()` both don't handle that case very well.

`bind_rows()` can be replaced in the simple case by `do.call(rbind.data.frame, ...)`, but there are some cases where styler is calling `bind_rows()` with a list of data frames, others where it is raw data.frames, and some where one of the arguments is actually `NULL`.


R/styler.R
```diff
@@ -19,7 +19,7 @@ if (getRversion() >= "2.15.1") {
   utils::globalVariables(c(
```
lorenzwalthert: Once we have removed the tidyr and dplyr dependency, we can probably also refrain from declaring the variables below as globals, right?

jimhester: No you still need them here because they are still used with Non-Standard Evaluation in arrange.

lorenzwalthert: Ok, right. And `filter` also I guess. I was just wondering whether it would be desirable to get rid of NSE completely.

krlmlr:
Awesome, thanks!

Issue #181, opened 2017-09-05
Consider removing dplyr and possibly purrr dependencies?


jimhester:
dplyr is a very heavy dependency which makes using **styler** on CI services such as travis or appveyor problematic.

Doing this would be useful so you could style the package as a CI step and ensure there are no changes as a way to verify that PRs conform to the style. But this will only work well if styler has limited (and light) dependencies.

If you are open to the possibility of removing the dependency I could work on a PR.

krlmlr:
I think removing dplyr and using compat-purrr.R is a good plan. Happy to review a PR.

lorenzwalthert:
I think since I know the project fairly well by now, I can start working on this in about two weeks if @jimhester is ok with that. If I remember correctly, the majority of dplyr calls was already removed with #90, so it should not be too much of work.

jimhester:
I have a WIP at https://github.com/krlmlr/styler/pull/182.

I also realized that tidyr also has a dplyr dependency, so we would need to remove tidyr as well.

lorenzwalthert:
Yes that might be tricky. Especially for creating the nested parse data - as the name says - we heavily rely on tidyr.

jimhester:
tidyr wasn't bad https://github.com/krlmlr/styler/pull/183

jimhester:
dplyr (#182) is complete now as well. Purrr is fairly light these days, we can remove it if we want, but probably not really needed.

krlmlr:
Should be good now.

PR #183, opened 2017-09-05
Remove tidyr dependency

jimhester:
This removes the tidyr dependency. It adds a dependency on rematch2 for convenience but that package is R only and small. We can remove it as well if we want.

Fixes part of #181


R/rules-spacing.R
```diff
@@ -144,36 +144,30 @@ set_space_between_levels <- function(pd_flat) {
 #'   after the regex "^#+'*".
 #' @importFrom purrr map_chr
 start_comments_with_space <- function(pd, force_one = FALSE) {
-  comment_pos <- pd$token == "COMMENT"
-  if (!any(comment_pos)) return(pd)
+  has_comments <- pd$token == "COMMENT"
+  if (!any(has_comments)) return(pd)
 
-  comments <- pd[comment_pos, ]
+  m <- rematch2::re_match(pd$text[has_comments],
```
krlmlr: Can you please rename `m` back to `comments`? This would simplify the review.

jimhester:
Ok updated accordingly, diff is much smaller now.

Now that dplyr and tidyr are gone we no longer have `%>%` imported implicitly, so I made it explicit in this PR.

krlmlr:
Thanks!

Issue #100, opened 2017-07-28
inconsistency in getParseData?


lorenzwalthert:
Why is the following different and what can we do about it?

``` r
arrow <- "a <- 3"
eq <- "a = 3"

styler:::create_tree(arrow)
#>                                             levelName
#> 1 ROOT (token: short_text [lag_newlines/spaces] {id})
#> 2  °--expr:  [0/0] {7}                               
#> 3      ¦--expr:  [0/1] {3}                           
#> 4      ¦   °--SYMBOL: a [0/0] {1}                    
#> 5      ¦--LEFT_ASSIGN: <- [0/1] {2}                  
#> 6      °--expr:  [0/0] {5}                           
#> 7          °--NUM_CONST: 3 [0/0] {4}
styler:::create_tree(eq)
#>                                             levelName
#> 1 ROOT (token: short_text [lag_newlines/spaces] {id})
#> 2  ¦--expr:  [0/1] {3}                               
#> 3  ¦   °--SYMBOL: a [0/0] {1}                        
#> 4  ¦--EQ_ASSIGN: = [0/1] {2}                         
#> 5  °--expr:  [0/0] {5}                               
#> 6      °--NUM_CONST: 3 [0/0] {4}

utils::getParseData(parse(text = eq))
#>  line1 col1 line2 col2 id parent     token terminal text
#> 1     1    1     1    1  1      3    SYMBOL     TRUE    a
#> 3     1    1     1    1  3      0      expr    FALSE     
#> 2     1    3     1    3  2      0 EQ_ASSIGN     TRUE    =
#> 4     1    5     1    5  4      5 NUM_CONST     TRUE    3
#> 5     1    5     1    5  5      0      expr    FALSE     

utils::getParseData(parse(text = arrow))
#>  line1 col1 line2 col2 id parent     token terminal text
#> 7     1    1     1    6  7      0        expr    FALSE     
#> 1     1    1     1    1  1      3      SYMBOL     TRUE    a
#> 3     1    1     1    1  3      7        expr    FALSE     
#> 2     1    3     1    4  2      7 LEFT_ASSIGN     TRUE   <-
#> 4     1    6     1    6  4      5   NUM_CONST     TRUE    3
#> 5     1    6     1    6  5      7        expr    FALSE     
```
Now there was just one expression, but the problem is worse if there are multiple, because with `=`, 
they will all be on the same level, i.e. more than one expression has parent 0.
```r
styler:::create_tree("a = 3 \n 3")
[...]

PR #188, opened 2017-09-07
fix parsing inconsistency

### LINKED ISSUE #187: styler executed via console and from within R does not always yield same result
As discussed at https://github.com/krlmlr/dplyr/commit/1c06efc6fa010b98d427b519e2958313895aab5c, styler does not return the same prettified code when run from within R 
```r
styler::style_file("R/bench-compare.r")
```
and from the command line
```bash
R -q -e 'styler::style_file("R/bench-compare.r")'
```

For the majority of other files in this repo (dplyr package), the styling is identical.

lorenzwalthert:
Not to be merged yet. Intended to fix #187. Could be made more functional and flexible so a list of `fix_funs` and one of `valid_funs` could be provided to `fix_p` and they would be applied/used for checking. `fix_pd` is currently really just fixing *one* issue and `is_valid_pd` checks only for this one issue.
Maybe even outsourcing these functions into a separate parsing package, along with issues detected in #100.


lorenzwalthert:
If this PR goes into the right direction, I can add a test, e.g. for the example introduced in #187.

lorenzwalthert:
Just realized that this PR does not solve the parsing error in  [R/bind.r](https://github.com/lorenzwalthert/dplyr/commits/master) because there, it seems the problem is of a different nature. Need to look into that further.

lorenzwalthert:
Redesigned PR completely according to the recent discussion in #187.

krlmlr:
Great, thanks!

PR #190, opened 2017-09-08
Substitute create filler

### LINKED ISSUE #132: rename create_filler()
I think this function name is a bit confusing and I would change it so somthing that reflects the purpose of the function better, like `initialize_attributes()` or something like that. I think this is important because people developing their own style guides will get in touch with this function, so it is not purely internal. Alternatively, we can also spit it up into multiple functions and give each a specific name like

* initialize_multi_line
* compute_space
* compute_newline
* initialize_ref_id
* initialize_indent

Which may be more clear but calling 5 functions instead of 1 on every level of nesting might add too much overhead. @krlmlr What's your take on that?

lorenzwalthert:
Renaming `create_filler()` to something more meaningful (`initialize_attributes()`) and split it up into smaller functions. Closes #132.


lorenzwalthert:
Contributes to #16.

PR #195, opened 2017-09-16
Fix space between commas

### LINKED ISSUE #192: Keep space between commas
``` r
styler::style_text("call(arg, , more_args)")
#> call(arg,, more_args)
styler::style_text("call(trailing, comma, )")
#> call(trailing, comma, )
```

The second example looks good, the first needs a space between the commas.

lorenzwalthert:
Put a space between two commas. Closes #192 .


tests/testthat/spacing/spacing_comma2-out.R
```diff
@@ -0,0 +1,2 @@
+call(arg, , more_args)
+a[, , drop = FALSE]
```
krlmlr: Do we need a space before the first comma?

lorenzwalthert: I'd say the convention is not to have one, but the example is poorly chosen. You are more likely to see something like this:
```r
a[1,, drop = FALSE]
```

krlmlr: This feels sufficiently rare, let's wait for a real use case.

PR #196, opened 2017-09-16
Sorting key

### LINKED ISSUE #112: Create key for sorting

lorenzwalthert:
Create new attributes `pos_id` that gives the number of the token (as long as no tokens are inserted). Simply using the column `id` for sorting did not work. Closes #112.

### DIFF
--- a/R/nested.R
+++ b/R/nested.R
@@ -4,9 +4,8 @@
 #'   representation into a nested parse table with
 #'   [nest_parse_data()].
 #' @param text A character vector to parse.
-#' @return A nested parse table. Apart from the columns provided by
-#'   `utils::getParseData()`, a column "short" with the first five characters of
-#'   "text" is added, the nested subtibbles are in column "child".
+#' @return A nested parse table. See [tokenize()] for details on the columns
+#'   of the parse table.
 compute_parse_data_nested <- function(text) {
   parse_data <- tokenize(text) %>%
     add_terminal_token_before() %>%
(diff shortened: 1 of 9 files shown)


R/nested.R
```diff
@@ -23,14 +22,26 @@ compute_parse_data_nested <- function(text) {
...
 #' @param text A character vector.
 #' @return A flat parse table
+#' @importFrom rlang seq2
 tokenize <- function(text) {
   # avoid https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=16041
   parse(text = text, keep.source = TRUE)
   parsed <- parse(text = text, keep.source = TRUE)
   parse_data <- as_tibble(utils::getParseData(parsed, includeText = NA)) %>%
     enhance_mapping_special()
+  parse_data$pos_id <- seq2(1, nrow(parse_data))
```
krlmlr: Using `1L` instead of `1` might just keep the data in the integer domain. I see that we later use numeric sometimes (+0.3 and +0.6), but this doesn't need to be the default.

lorenzwalthert: Thanks, sure.

lorenzwalthert: Ok, thanks. I think `seq2()` creates intervals of integers anyways, 
``` r
class(rlang::seq2(1, 3))
#> [1] "integer"
```
but we can be a bit more explicit if you like.

krlmlr:
This property of `seq2()` is good to know.

PR #200, opened 2017-09-17
Fix indention in if-else statement

### LINKED ISSUE #191: Wrong indention of else if braces are missing
``` r
styler::style_text("{if (TRUE)\n1\nelse\n2\n}")
#> {
#>   if (TRUE)
#>     1
#>     else
#>     2
#> }
```

Can we add braces around both the `if` and `else` parts here?

lorenzwalthert:
`compute_indent_indices()` assumed that if the nest expression contains `(`, the closing `)` is always the last token, which is not true for if-else clauses. For that reason, everything between `(` and the last token got indented.
``` r
styler::style_text("{if (TRUE)\n1\nelse\n2\n}")
#> {
#>   if (TRUE)
#>     1
#>     else
#>     2
#> }
```
Now, the closing token can be supplied and located. Consequentially indention can be done for all tokens between the opening and the closing token. Closes #191.


lorenzwalthert:
Well, I guess technically we could have done that, but it would not have been a clean solution I guess. Happy with the PR otherwise?

lorenzwalthert:
Ok. Yes, I am confident that these changes are required for going forward and keep code consistent.

PR #203, opened 2017-09-18
add vignette on introducing styler

lorenzwalthert:
I think it would be good to have a vignette in the package that introduces styler. The GSOC vignette is not part of the `upstream` repo.
Hence, this is mainly taken from gsco_contrib from @lorenzwalthert's fork, branch `gsoc2017`.


krlmlr:
I think an introductory vignette is nice to have, will do a more detailed review just before release (and on the rendered output).

Issue #201, opened 2017-09-17
add curly braces to if-else


lorenzwalthert:
If an if-else statement spans over mutliple lines, add curly braces, so 
``` r
#> {
#>   if (TRUE)
#>     1
#>   else
#>     2
#> }
```
Should be 
``` r
#> {
#>   if (TRUE) {
#>     1
#>   } else {
#>     2
#>   }
#> }
```

Follow-up of #191.

lorenzwalthert:
@krlmlr Should that be a `strict = TRUE` transformation?

krlmlr:
Currently we have both `strict = TRUE` and `scope = "token"`, which is a bit confusing. I think this definitely belongs in the "token" scope, not sure about strictness.

lorenzwalthert:
Ok, let it be a `strict = TRUE` transformation since I think it can be legit to write:
```r
if (also_quite_long(call(a, lazy(other(call)))
  return(very_long(call(1+1, another, long, agrument)
```

PR #204, opened 2017-09-18
Maintenance

### LINKED ISSUE #144: Properly adding tokens
Currently, `add_brackets_in_pipe()` creates an out-dated parse table since `token_before` and `token_after` of the neighbouring tokens are not updated. Has no effects on any rule to my knowledge, but we can't really allow such an inconsistency if other people want to use the column.

lorenzwalthert:
Some maintenance and infrastructure needed to solve #201 later. Closes #144.


R/dplyr.R
```diff
@@ -51,11 +51,7 @@ bind_rows <- function(x, y = NULL) {
       y[[nme]] <- NA
     }
   }
-  rbind.data.frame(x, y)
-}
-
-if_else <- function(condition, true, false, missing = NULL) {
```
krlmlr: I think we might just keep `if_else()`. In dplyr it's a stricter variant of `ifelse()`, it's just not currently implemented like this but could be.

R/dplyr.R
```diff
@@ -51,11 +51,7 @@ bind_rows <- function(x, y = NULL) {
...
-  ifelse(condition, true, false)
+  rbind.data.frame(x, y, ...)
```
krlmlr: What's the point in changing the API here? Do you support more than two arguments? In this case it might be slightly better to forward to `bind_rows(rbind.data.frame(x, y), ...)` .

lorenzwalthert: Yes that was the idea.

R/dplyr.R
lorenzwalthert: I just found it confusing how the function is defined now, because at first look, people may think it's the strict version from dplyr but it's not.

krlmlr: The dplyr implementation seems to have too many dependencies (in terms of internal functions used) to be dragged over here. But something like:

```r
stopifnot(length(condition) == length(true))
stopifnot(length(condition) == length(false))
if (!is.null(missing)) stop("missing arg not yet implemented")
```

might just do the job for now.

lorenzwalthert: Ok, the problem is that we deliberately allow for an output that has length zero, e.g. in `compute_indent_indices()`, so we have used `ifelse()` all the time.Should we allow for length zero response in "our" implementation of `if_else()`?

krlmlr: Perhaps:

```r
stopifnot(length(true) %in% c(length(condition), 1L))
stopifnot(length(false) %in% c(length(condition), 1L))
if (!is.null(missing)) stop("missing arg not yet implemented")
```

?

lorenzwalthert: Well, if length zero should be supported, I guess your `1L` should be a `0L`?

krlmlr: We support zero length for the condition only.

lorenzwalthert: We have
```r
if_else(is.null(token_closing),
  npd,
  which(pd$token %in% token_closing) - 1
)
```
So `false` is of length zero if `token` is not in `token_closing`

lorenzwalthert: And the condition is of length one.

lorenzwalthert: Or should we simply use 
```r
if (cond) {
  ret <- integer(0)
} else {
  ret <- xyz(fun)
}
```
For those cases?

krlmlr:
``` r
dplyr::if_else(TRUE, 1L, integer(0))
#> Error: `false` must be length 1 (length of `condition`), not 0
```

`if_else()` is a safety wrapper to ensure that the output length matches the input length. I guess we'll need something different, perhaps a regular `if()` for other cases.

krlmlr: (Just the way you suggested.)

lorenzwalthert: Ok, will do so.

PR #205, opened 2017-09-18
Rstudioaddin

lorenzwalthert:
Closes #178 by making use of `rstudioapi::insertText()`. Simplifies code greatly plus the Addin can now style unsaved files. No automated testing possible though since `insertText()` needs the document ID, which we cannot obtain when mocking.

PR #208, opened 2017-09-18
Simplify RStudio Addin

### LINKED ISSUE #178: Use rstudioapi::modifyRange() for Addin
I have this `data.table` code:

```r
ranges[, non_white_char_total := {range_count = nchar(str_replace_all(text, "\\s*", ""))
                                         cumsum(range_count)}]
```
I selected it, used `style active region` addin menu and there is this error

```
Error in parse(text = text, keep.source = TRUE) : 
  <text>:3:0: unexpected end of input
1: ranges[, non_white_char_total := {
2:   range_count <- nchar(str_replace_all(text, "\\s*", ""))
  ^
```

lorenzwalthert:
Replaces #205. Closes #178 by making use of rstudioapi::insertText(). Simplifies code greatly plus the Addin can now style unsaved files. No automated testing possible though since insertText() needs the document ID, which we cannot obtain when mocking.


R/addins.R
```diff
@@ -13,103 +13,15 @@ style_active_file <- function() {
...
-  end <- context$selection[[1]]$range$end
-  if (end[2] == 1 & !all(start == end)) {
-    end[1] <- end[1] - 1
-    end[2] <- 1000000L # because of range constraint in substr()
-  }
-  list(start = start, end = end, path = path)
+  text <- context$selection[[1]]$text
+  if (all(nchar(text) == 0)) stop("No code selected")
+  out <- style_text(text)
+  rstudioapi::modifyRange(
```
krlmlr: Can we extract a wrapper around `modifyRange()` and mock that wrapper?

lorenzwalthert: The problem is that `rstudioapi` assigns the document ID dynamically, and `modifyRange()` needs this ID to select the document to modify, so I think we can't mock. The approach would have been something like 3fac3a9.
I suggest we just leave it without testing for now.

lorenzwalthert: The problem is that when modifying the file with `modifyRange()` we need the document id of the file to modify, which we get with `getActiveDocumentContext()` when run interactively. However, I don't see a way to get id without running `getActiveDocumentContext()` and ids are assigned dynamically (which can be tested by simply opening a file and tet the active context of it, closing it and get the context again).

PR #212, opened 2017-09-21
Tools to insert tokens + application on if-else clauses

### LINKED ISSUE #201: add curly braces to if-else

lorenzwalthert:
This PR provides tools to insert tokens to a parse table. In particular, a mechanism to create valid `pos_id` attributes for the new token is implemented to make the method secure. The application of interest was to wrap if-else clauses in curly brackets if they span multiple lines, i.e. to turn 
```r
if (x)
  q
else
 c
```
Into
```r
if (x) {
  q
} else {
 c
}
```
styler will do the above only if `strict = TRUE` and `scope = "tokens"`.

Closes #201.


R/token-create.R
```diff
@@ -0,0 +1,135 @@
...
+    parent = parents,
+    token_before = token_before,
+    token_after = token_after,
+    id = NA,
+    terminal = rep(TRUE, len_text),
+    internal = rep(FALSE, len_text),
+    spaces = spaces,
+    multi_line = rep(FALSE, len_text),
+    indention_ref_id = indention_ref_ids,
+    line1 = rep(NA, len_text),
```
krlmlr: I'd rather omit NA columns, and let `bind_rows()` take care of it.

lorenzwalthert: Our `bind_rows()` implementation does only support matching column names.
``` r
styler:::bind_rows(data.frame(a = 1, b = 3), data.frame(b = 3, c = 4))
#> Error in match.names(clabs, names(xi)): names do not match previous names
```

lorenzwalthert: So maybe just rever that to the initial PR.

PR #217, opened 2017-09-22
Add tools for re-indenting tokens

### LINKED ISSUE #64: allow for zero-indention of comments
For breaking code into sections (either manually or with a tool such as [strcode](https://github.com/lorenzwalthert/strcode)), it is desirable that these are not indented. This is not currently supported since all tokens (including comments) will be indented as the current level requires. One way to deal with that would be to set indention of these token to `-Inf` and using `max(0, indention)` in `serialize_parse_data_nested()` although this to be not particularly elegant.

lorenzwalthert:
Add tools for re-indenting tokens that match certain regular expression patterns, in particular, to allow strcode code separators. Closes #64, but is actually much more general. The indention level for an arbitrary token text that matches any regular expression in the regex vector passed to the styling function can be controlled. One absolute indention level for all matches in one run can be set. Since we have `scope = "spaces"` (which does not modify indention), we can just run styler repeatedly and set the indention level of different tokens sequentially. But probably no one would like to do that actually.

See `tests/testthat/unindention_regex/random_non_comment_indention-out.R` to see the how the indention level for the token `bbx` was set to 5. This function call that lead to this result can be found in `tests/testthat/unindention_regex/random_non_comment_indention-out.R`


R/style_guides.R
```diff
@@ -49,7 +50,8 @@ NULL
 tidyverse_style <- function(scope = "tokens",
                             strict = TRUE,
                             indent_by = 2,
-                            start_comments_with_one_space = FALSE) {
+                            start_comments_with_one_space = FALSE,
+                            reindention = specify_reindention()) {
```
krlmlr: `default_reindention()`? What is the rationale behind the `specify_` prefix?

R/style_guides.R
```diff
@@ -166,10 +172,35 @@ create_style_guide <- function(initialize = initialize_attributes,
...
+#' defaults, so the user can specify deviations from them conveniently without
+#' the need of setting all arguments explicitly.
+#' @param regex_pattern Character vector with regular expression patterns that
+#'   are to be re-indented with spaces.
+#' @param indention The indention tokens should have if they match
+#'   `regex_pattern`.
+#' @param comments_only Whether the `regex_reindention_pattern` should only be
+#'   matched against comments or against all tokens. Mainly added for
+#'   performance.
+specify_reindention <- function(regex_pattern = regex_strcode(),
```
krlmlr: Should this default to `regex_none()`, at least for the tidyverse style?

R/reindent.R
```diff
@@ -111,3 +111,71 @@ apply_ref_indention_one <- function(flattened_pd, target_token) {
...
+#'
+#' @name regex_for_reindention
+NULL
+
+#' @describeIn regex_for_reindention Returns the regular expression pattern used
+#' in the [strcode](https://github.com/lorenzwalthert/strcode) package.
+#' Character vector with the patterns of code separator and title for the three
+#' levels of granularity.
+#' @export
+regex_strcode <- function() {
```
krlmlr: This adds a hard-to-track dependency to strcode. Maybe add to strcode instead (also requires to change the default for the tidyverse)?

Perhaps strcode could also implement a style guide, to showcase extensibility?

lorenzwalthert: Ok, agree, I can export `regex_strcode()` from the `strcode` package.

R/style_guides.R
lorenzwalthert: Ok.

R/style_guides.R
lorenzwalthert: We can use a forwarder here too to be consistent with  #221.

lorenzwalthert: Same line of argument as in https://github.com/krlmlr/styler/pull/221#discussion_r142003169. I suggest to revert to `specify_reindention()`.

lorenzwalthert: @krlmlr ok?

Issue #86, opened 2017-07-25
differentiate between math operators


lorenzwalthert:
Allow the user to set spacing separately for 
* \+ and -
* / and *
* ^

etc.

krlmlr:
Reference: https://github.com/krlmlr/styler/pull/79#discussion_r129096510

lorenzwalthert:
Should this be a separate option or go under strict? I suggest creating a new argument `math_token_spacing` that takes a list that has two lists in it: One containing the tokens where we want `1` space before/after it and one that contains the tokens that should have no spacing before/after them. We can use a helper function to create this list like that
```r
specify_math_token_spacing(zero = c("'*'", "'/'"), one = c(...)
```
Tokens that are not mentioned in `zero` or `one` will have space around it, since that the tidyverse style guide encourages.

lorenzwalthert:
Note that the proposed API is similar to the one proposed in #217 where we have
`reindention = specify_reindention(...)`

lorenzwalthert:
Closed with #221.

Issue #82, opened 2017-07-25
spacing around (more) operators


lorenzwalthert:
* `in`, like in for the fixme in tests/testthat/transformer_grouping/scope_none-out.R.

krlmlr:
Reference: https://github.com/krlmlr/styler/pull/79#discussion_r129090876

PR #221, opened 2017-09-24
Math token spacing

lorenzwalthert:
This PR gives granular control over spacing around math tokens, as envisioned in #82 and requested in https://github.com/tidyverse/lubridate/pull/577#issuecomment-331700557.  
Implements a new paradigm for transformer functions in the sense that `strict` is now an argument to
the transformer. This reduces duplication of code, since the only part that `set_*` and `add_*` functions for spacing rules are different is that `set_*` uses `pd_flat$spaces[...] <- level` wherease `add_*` uses `pd_flat$spaces[...] <- pmax(1, pd_flat$spaces[...]`.
Functions that follow this paradigm should be named `style_*` instead of `add_*` and `set_*`.
`pd_flat` is the last argument to `style_space_around_math_token()` because this allows using `purrr::partial()` without naming arguments.


lorenzwalthert:
As a follow-up and for consistency, other spacing rules could also be transformed from `set_*` and `add_*` to `style_`.

tests/testthat/math_token_spacing/strict_math_spacing_zero_all_but_power-out.R
```diff
@@ -0,0 +1 @@
+1++1-3/23*3 ^ 4
```
krlmlr: I wonder if the other way round would be more usual, because of operator precedence:

```r
+1 + +1 - 3 / 23 * 3^4
```

R/style_guides.R
```diff
@@ -186,3 +193,45 @@ character_to_ordered <- function(x, levels, name = substitute(x)) {
...
+#'   math_token_spacing = specify_math_token_spacing(zero = "'+'"),
+#'   strict = FALSE
+#' )
+#' style_text(
+#'   "1+1   -3",
+#'   math_token_spacing = specify_math_token_spacing(zero = "'+'"),
+#'   strict = TRUE
+#' )
+#' @export
+specify_math_token_spacing <-
```
krlmlr: Maybe `default_math_token_spacing`?

tests/testthat/math_token_spacing/strict_math_spacing_zero_all_but_power-out.R
lorenzwalthert: Right, I can adapt the test. The default should remain as in the tidyverse style guide, so all operators have spacing around them I think.

R/style_guides.R
lorenzwalthert: Not sure whether I understand you correctly. What does it have to do with a default? Everything will be specified unambiguously. We can also drop the `specify` from the name in order to make it sound a bit more simplistic.

krlmlr: Oh, I see. I only saw the use of the function as a default argument. Maybe implement two functions then, `math_token_spacing()` and `default_math_token_spacing()`, and have the latter forward to the former?

lorenzwalthert:
Yes, maybe we can consider adding braces to complex expressions. However, I am not sure how straightforward that is. Depends a lot on the structure of the nested parse data I think. Maybe other PR or issue?

R/style_guides.R
lorenzwalthert: Ok, we can do that.

lorenzwalthert: I am not sure whether we need the forwarder. I.e. if we only have one style guide, then that might be fine, but other style guides may have other defaults. So maybe call it `tidyverse_math_token_spacing()`?

lorenzwalthert: We cannot use a function that has the same name as an argument (`math_token_spacing()`):
```r
tidyverse_style <- function(scope = "tokens",
                            strict = TRUE,
                            indent_by = 2,
                            start_comments_with_one_space = FALSE,
                            math_token_spacing = math_token_spacing()) {
# [...]
```
Because we get 
```r
#> Error: promise already under evaluation: recursive default argument reference or earlier
#> problems?
```

I suggest to revert to `specify_math_token_spacing()`, since that is also a verb and functions are supposed to be verbs according to the tidyverse style guide  I think.

lorenzwalthert: @krlmlr ok? We can keep `tidyverse_math_token_spacing()` that forwards to `specify_math_token_spaching()`.

lorenzwalthert:
Closes #86.

Issue #13, opened 2017-05-14
style R Markdown (code chunks)


lorenzwalthert:
It would be nice if styler could not only format plain .R files, but also .Rmd files. 

## Code Chunks
This could probably by implemented by changing `transform_files`, i.e. adding a function `transform_lines_rmd_chunk`.  Main steps are:
* first `read read_lines_enc` 
* look for regex `^´´´\\{r.*\\} *$` and `^``` *$` 
* transforming lines between these expressions while leaving everything else unchanged
* write back with `write_lines_enc`

@krlmlr maybe this functionality fits even in the `utf8` package?

## Plain Text
* Line breaks if text is longer than 80 characters
* punctuation (i.e. space after `. , ? ! ` ect.)

The second point might be out of the scope of the package though...

krlmlr:
Could we use knitr to parse the .Rmd file? The `transform_lines_enc()` function in utf8 seems generic enough.

Not sure about styling plain text, realistically we'd need to parse and write back Markdown.

lorenzwalthert:
As far as I know, there is only the `knitr` code chunk option `tidy` that tidies up the output based on `formatR`, but I don't know how `knitr` could tidy the source.

krlmlr:
knitr is not supposed to test it, just to do the extraction work for us (determine what is code, what language is that code in, what is text).

jimhester:
We do this in lintr, https://github.com/jimhester/lintr/blob/aa70430282852afa15a61cc69771e71dd32e1aec/R/extract.R

lorenzwalthert:
Thanks @jimhester for that hint. I just tried out the function and it indeed seems to me that we could use it to achieve our goal. 
I know its a minor thing but I just thought about inline R code in Rmd documents. It would be nice if that could also be covered, but I think this is not the case with `extract_r_code`, right? 
Also, what was the reason against using (internal) `knitr` functionality in the `lintr` package to do the extraction, as @krlmlr suggests?

@krlmlr I found the exported function `knitr:pulr(text = character_vec)` that returns all code lines (not only R) as a character vector, but I think this does not really help us. I just had a quick look at internal functions of knitr, and probably there are tools that could help us, but I did not really have time yet to explore that further.

jimhester:
Inline R code is typically a single statement long, so likely does not have anything worth linting / styling. Supporting it would complicate the implementation for limited gain.

[...]

PR #233, opened 2017-10-09
Update to allow styling of Rmd files

jonmcalder:
This PR aims to address the first part of #13 - allowing styling of individual Rmarkdown files

* Detection of R code chunks is done by `knitr:::detect_pattern()`
* `transform_file()` is updated with a call to a new `transform_code()` function, which is a wrapper for `utf8::transform_lines_enc()` and handles the logic for transforming `R` and `Rmd` files
* New functions have been documented and a few basic test cases added


lorenzwalthert:
Cool Jon, thanks so much for woking on this. Will review tomorrow.

R/style_rmd.R
```diff
@@ -0,0 +1,68 @@
...
+#'
+#' @inheritParams utf8::transform_lines_enc
+#' @param ... Further arguments passed to `utf8::transform_lines_enc()`.
+transform_code <- function(path, fun, verbose, ...) {
+  if (grepl("\\.R$", path, ignore.case = TRUE)) {
+    utf8::transform_lines_enc(path, fun = fun, ...)
+  } else if (grepl("\\.Rmd$", path, ignore.case = TRUE)) {
+    utf8::transform_lines_enc(path, fun = partial(transform_rmd, transformer_fun = fun), ...)
+  } else {
+    stop("")
```
lorenzwalthert: Maybe we can give a more informative error message here?

R/style_rmd.R
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+#' Transform Rmd contents with a transformer function
+#'
+#' @param lines A character vector of lines from an Rmd file
```
lorenzwalthert: Can you use tidyverse style in your roxygen comments? In particular, argument descriptions should end with a full stop here and in other places.

R/ws.R
```diff
@@ -155,6 +155,7 @@ style_file <- function(path,
 #' @inheritParams style_dir
 #' @param path The path to a file that should be styled.
 prettify_one <- function(transformers, path) {
-  if (!grepl("\\.[Rr]$", path)) stop(path, " is not a .R file")
+  if (!grepl("\\.[r](md)?$", path, ignore.case = TRUE))
+    stop(path, " is not an R or Rmd file")
```
lorenzwalthert: Do we still need this check if we check further downstream?

lorenzwalthert:
I just tested this locally and it works nicely with the Add-in too.

R/ws.R
jonmcalder: You're right the downstream check does make this redundant. The downstream check has to be there since it provides the logic for R/Rmd handling. I suppose the only advantage to having this earlier check is that the process is stopped before the call to `transform_files()` which is before any `messages` are output to the console.

I'll check if it can be removed.

lorenzwalthert: Thanks. In that case, you can remove `prettify_one()`.

lorenzwalthert:
Reference: #250.

Issue #44, opened 2017-06-22
add start token in nested structure


lorenzwalthert:

lorenzwalthert:
Closed by [#4](https://github.com/lorenzwalthert/styler/pull/4)

PR #243, opened 2017-10-12
Handle styling of an unsaved active file

### LINKED ISSUE #235: style active unsaved file
Adapting the function `style_active_file()`. Implementation similar to #143 (i.e. via `rstudioapi::modifyRange()`) so unsaved files can be styled too.

jonmcalder:
Fixes #235


R/addins.R
```diff
@@ -3,7 +3,15 @@
 #' Helper function for RStudio Addin.
 style_active_file <- function() {
   context <- get_rstudio_context()
-  style_file(context$path, style = tidyverse_style)
+  if (context$path != "") {
```
lorenzwalthert: What's the advantage of case distinction here? Can't we just use `rstudioapi` to handle all cases?

lorenzwalthert:
Thanks for working on that Jon.

R/addins.R
jonmcalder: I just assumed I should avoid changing the current behaviour for the case when the active file has been saved (i.e that it is styled and saved).

Was your intention with #235 to just modify (without saving) in all cases?

lorenzwalthert: For consistency, I suggest to stick with `rstudioapi` throughout. See https://github.com/krlmlr/styler/issues/235#issuecomment-336376681.

lorenzwalthert:
Let's do away with these startup messages and implement the default to not save after styling. We can maybe add a help file "styler_addins" (with an alias "style_fule" and friends) where we describe each addin and the option quickly and make reference to .Rprofile.

krlmlr:
On that note, I've noticed that styling a chunk in an `.Rmd` file pulls the closing backticks `` ``` `` directly after the selection, is this still true after this change?

The name of the menu item "Style active region" may be confusing, can we change to "Style selection"?

jonmcalder:
Thanks for picking that up @krlmlr - was that using the add-in implementation from this branch or an earlier version? I didn't do extensive testing and haven't looked at this recently so will need to check again later when I'm home.

I agree with the suggested name change - @lorenzwalthert if you are happy with that too can I include it in this PR?

krlmlr:
This was with styler from master.

lorenzwalthert:
Yes, I am happy with the name change in this PR

lorenzwalthert:
As far as the lost end-of-chuck line breaks goes, I think that's not related primarily to this PR, but styler drop lines at the end anyways, 
``` r
reprex::reprex_info()
#> Created by the reprex package v0.1.1.9000 on 2017-10-26

styler::style_text("1++1\n")
#> 1 + +1
```
That stems from the fact that we don't have an end token. See discussions around start tokens in #218, #44 and #4. We always have a final line blank if we use functions that use `enc::transform_lines_enc()` because it does add that line for us, but here we use `style_text()`. Why do you @krlmlr think this is undesirable? If we want to ensure a blank line at the end of every code chuck, we could just add it after serialisation.

lorenzwalthert:
@jonmcalder please take note of #264 and adapt the PR accordingly.

krlmlr:
The closing backticks in a `.Rmd` need to remain on a separate line, and if the selection ends with newlines, styler should keep them.

lorenzwalthert:
Also @jonmcalder can you please rebase on master because then we can test on `.Rmd` files too. @krlmlr I can't reproduce the behaviour you describe with the current master branch. I only see blank lines removed when I use  the Addin for styling highlighted code, but the back-ticks remain on a separate line, i.e. they don't follow the code on the same line as the code ends. With highlighting code and back-ticks, I can't perform the styling as the parse data is invalid.

lorenzwalthert:
Ok @krlmlr now I get what you mean. If I recall correctly we had that problem with `style_selection()` before we used `rstudioapi::modifyRange()`. It think we used [this](https://github.com/krlmlr/styler/pull/208/files#diff-fc3f33c61709e4e0dec6f2567f50af09L36) workaround to adjust the range for this case. However, we could probably do that in a slightly more elegant way by adding a line break at the end of the styled expression before inserting it again.

jonmcalder:
Ok documentation added/updated - I think this should be ready for review now @lorenzwalthert.

My understanding is that the above conversation around refining the usage of `rstudioapi::modifyRange()` within `style_selection()` and/or updating `style_text()` to handle Rmd content is out of scope and this will need to be addressed in separate PR's.

So just to summarize the content of this PR again:
- update `style_active_file()` to allow for styling of untitled/unsaved files
- rename `style_active_region()` to `style_selection()`
- add option to automatically save after styling via these addins (using an environment variable)
- update function documentation

R/addins.R
```diff
@@ -1,25 +1,50 @@
-#' Style the active file
+
+#' Stylers for RStudio Addins
+#'
+#' Helper functions for styling via RStudio Addins.
+#'
+#' @section Auto-Save Option:
+#' By default, both of the RStudio Addins will apply styling to the (selected)
+#' file contents without saving changes. Automatic saving can be enabled by
+#' setting the environment variable \code{save_after_styling} to \code{TRUE}.
```
lorenzwalthert: We use markdown roxygen, so can you please change
```r
\code{...}
```
to 
```r
`...`
```
here and elsewhere in the PR?

R/addins.R
```diff
@@ -1,25 +1,50 @@
...
 }
 
 
-#' Style the highlighted region
-#'
-#' Helper function for RStudio Addin. This function is complicated because of
-#' one thing: You can highlight also just parts of lines.
+
+#' @describeIn styler_addins Styles the highlighted region
 #' @importFrom rlang seq2
```
lorenzwalthert: I think the `rlang` import is redundant. Can you remove as part of this PR?

lorenzwalthert:
Ok, thanks @jonmcalder. For me it's ok to solve the problem with the back-ticks brought up by @krlmlr later. I already solved it in [some](https://github.com/lorenzwalthert/styler/tree/jmc_style-active-file) branch, so I will add the commits after we merged this PR.

One last thing: Would you mind updating `R/styler.R` to contain a reference to to the RStudio Addin help file (line 10) just as for the other stylers?

R/addins.R
jonmcalder: Oops! Sorry, old habits...

R/addins.R
jonmcalder: Sure.

R/styler.R
```diff
@@ -7,7 +7,8 @@
 #'  * [style_file()] to style a single .R file.
 #'  * [style_dir()] to style all .R files in a directory.
 #'  * [style_pkg()] to style the source files of an R package.
-#'  * An RStudio Addin to style the active file .R file and the current package.
+#'  * [styler_addins] (RStudio Addins) to style either selected code or the
```
lorenzwalthert: Can you add two braces after `styler_addins` (as it is done with `style_pkg()` etc. above) to not just make a reference, but also use code font markup?

jonmcalder: I was deliberately trying to avoid making the reference look like a function since it's not. Does that distinction not really matter here?

lorenzwalthert: Ok, that actually makes sense. You are right.

lorenzwalthert:
Thanks

Issue #238, opened 2017-10-12
Move to r-lib?


hadley:
I think the main thing that you're missing is travis tests for R 3.1 on

lorenzwalthert:
styler now works with R>=3.1.

hadley:
I'm happy to accept this in R-lib if you want to move it :)  (you'll retain admin rights)

krlmlr:
Done.

PR #249, opened 2017-10-23
Test R 3.1 and R 3.2

krlmlr:
@lorenzwalthert: Our suggested data.tree imports DiagrammeR which requires R >= 3.2. Could you please look into making R CMD check work if data.tree is not available? You can push to this branch if you like.

Required for #238.


krlmlr:
Link to travis build: https://travis-ci.org/krlmlr/styler/jobs/291431915#L3122

krlmlr:
We can just use backports to fix the trimws error: https://travis-ci.org/krlmlr/styler/jobs/293558005#L1786

lorenzwalthert:
`trimws()` was only introduced in R3.2 and we need it to drop spaces at the end of comments. Should we implement regex-based function that does something similar?

lorenzwalthert:
Ok, great.

lorenzwalthert:
Ok, now I made it work. `test_collection()` has a new default `write_tree = NA`, in which case it is determined via `getRversion()` whether a tree should be written or not. If set to `TRUE` under R < 3.2. an informative error will be thrown. In addition, I had to replace some vignette content with rendered markdown. `create_node_from_nested_root()` is the workhorse tree creation function that throws the same error if a tree should be produced with R < 3.2.

R/nested_to_tree.R
```diff
@@ -17,11 +17,14 @@ create_tree <- function(text) {
...
 #' @return An object of class "Node" and "R6".
 #' @examples
+#' if (getRversion() >= 3.2) {
 #' code <- "a <- function(x) { if(x > 1) { 1+1 } else {x} }"
 #' nested_pd <- styler:::compute_parse_data_nested(code)
 #' initialized <- styler:::pre_visit(nested_pd, c(styler:::initialize_attributes))
 #' styler:::create_node_from_nested_root(initialized)
+#' }
 create_node_from_nested_root <- function(pd_nested) {
+  if (getRversion() < 3.2) stop_insufficient_r_version()
```
krlmlr: Can we just check if the data.tree package is available here? They might start supporting R 3.1, or it still may be missing because it's Suggests only.

vignettes/customizing_styler.Rmd
```diff
@@ -221,7 +221,33 @@ position of line breaks and spaces. Let's first create the nested parse table.
 
 ```{r}
 code <- c("add_one <- function(x) { x + 1 }")
+```
+
+``` r
 styler:::create_tree(code)
+```
+
+    ##                                              levelName
```
krlmlr: I wonder if this still renders correctly in the vignette.

krlmlr: Oh, this is a code block, too. I was just looking for the triple backticks.

Issue #250, opened 2017-10-23
Add style_dir() to RStudio add-in


jarodmeng:
Currently the add-in provides 3 options to style active region, active file, or the whole package. It would be awesome to add another option to style all files in the working directory. There's already a `style_dir()` function that can be bound to the option.

lorenzwalthert:
Thanks for bringing that up. I am not sure whether we should add that though. `styler` already has three RStudio Addins and I think the RStudio Addin Menu should not be "cluttered" with too many Addins from one package. In particular, styling of the active file and styling of a highlighted region can be achieved more efficiently via an Addin compared to the command line, so that's why we added the Addin. Both styling of the package and styling of the working directory can be achieved via the command line fairly quickly. Maybe we could also change the "Style package" Addin to something like "Style RStudio Project" to cover both packages and other projects? @krlmlr any thought on that?

jarodmeng:
The use case I have is when I write multiple R/Rmd scripts in a project that's not organized as a package. I agree that it can be achieved by calling `styler::style_dir()`. But I find myself running that command often enough that having an add-in option really saves time. I, for one, can tolerate a slightly longer add-in menu. :)

lorenzwalthert:
Reference: #233. One idea was to extend `style_dir()` and friends to `.Rmd` files, which currently only works with `style_file()`.

krlmlr:
A `style_project()` function feels helpful, should look at both `.R` and `.Rmd` files and work for packages and script projects. I'd prefer keeping the add-in menu lean, if we really need more options, a small Shiny app would give a better experience IMO.

lorenzwalthert:
So how is `style_project()` then different from `style_dir()`? If we add `.Rmd` support (which is not too much of work from where we are since `style_file()` can already handle `.Rmd`), we probably want the user to be able to specify which extensions to style, right? So `style_project()` could be an alias for `style_dir(..., include_rmd = TRUE)` whereas `style_dir` gains arguments `include_r` that defaults to `TRUE` and `include_rmd` that defaults to `FALSE`?

PR #251, opened 2017-10-23
Add option to style working directory in RStudio addin

jarodmeng:
Related to #250


lorenzwalthert:
To be closed for now after discussion in https://github.com/r-lib/styler/issues/250#issuecomment-341636504.

PR #260, opened 2017-10-25
Indention of function declaration and closing braces

### LINKED ISSUE #253: Function declarations
``` r
styler::style_text("function (a = 1,\nb = 2\n){\nNULL\n}", strict = FALSE, scope = "indention")
#> function (a = 1,
#>          b = 2
#>        ) {
#>   NULL
#> }
```

- [ ] The `b` argument needs to be aligned with the opening paren
- [ ] The closing paren and the opening brace should follow `b = 2`, or at least align with the opening paren

lorenzwalthert:
Closes #253. Moving the closing brace all the way to the left if it can't be moved one line up seems a fairly rare case and it can't be implemented easily because the most left expression is not in the same nest when the indention is referenced.
I.e. this
```r
a <- function(x,
              y #
            )
```
Remains as is and does not become
```r
a <- function(x,
              y #
)
```
Because of the reference issue and also not
```r
a <- function(x,
              y #
              )
```
Because of the brace unindention mechanism that applies here. I think we can leave that as is for now since it should be fairly rare.


tests/testthat/line_breaks_and_other/line_break_fun_dec-in.R
```diff
@@ -0,0 +1,17 @@
...
+
+a <- function(x, #
+              y) #
+{
+  x
+}
+
+a <- function(x, #
+              y #
+            ) {
```
krlmlr: I think we can still add a FIXME here, because this isn't what we'd truly like to see.

PR #261, opened 2017-10-25
Allow empty {} without line break.

### LINKED ISSUE #256: Allow empty blocks
``` r
styler::style_text("list(function(){})", strict = FALSE, scope = "line_breaks")
#> list(function() {
#> })
```

We don't need to add a new line if the block is truly empty. This happens occasionally with R6 classes.

``` r
styler::style_text("{}", strict = FALSE, scope = "line_breaks")
#> {
#> }
```

Blocks also occur naked sometimes.

lorenzwalthert:
* Minor simplification of code based on the insight that curly braces expressions exhibit symmetry, which in turn
* Facilitates checking whether or not a line break should be added before and after an opening / closing curly brace respectively.

Closes #256.

### DIFF
--- a/R/rules-line_break.R
+++ b/R/rules-line_break.R
@@ -5,26 +5,19 @@ remove_line_break_before_curly_opening <- function(pd) {
   pd
 }
 
-set_line_break_around_curly <- function(pd) {
+style_line_break_around_curly <- function(strict, pd) {
   if (is_curly_expr(pd) && nrow(pd) > 2) {
     closing_before <- pd$token == "'}'"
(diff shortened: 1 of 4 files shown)


lorenzwalthert:
We could even go one step further and add the argument `strict` to the a function `style_line_break_curly()` (which would replace `set_line_break_curly()` and `add_line_break_curly()`) to handle both strict and non-strict case, as it was done with `style_space_around_math_token()`.

krlmlr:
Yeah, why not? That would slightly add complexity, but reduce redundant code.

lorenzwalthert:
We now need 22 lines less code to achieve the same as before and also fix #256.

PR #262, opened 2017-10-25
avoid checking for hard-coded dot

### LINKED ISSUE #81: Avoid checking for hard-coded "." in token text
https://github.com/krlmlr/styler/pull/80/files#diff-abef16e00aa68f1dab0e93325dd84b74R3

``` r
styler:::create_tree("a %>% b")
#> 1 ROOT (token: short_text [lag_newlines/spaces] {id})
#> 2  °--expr:  [0/0] {7}                               
#> 3      ¦--expr:  [0/1] {3}                           
#> 4      ¦   °--SYMBOL: a [0/0] {1}                    
#> 5      ¦--SPECIAL-PIPE: %>% [0/1] {2}                
#> 6      °--expr:  [0/0] {6}                           
#> 7          °--SYMBOL: b [0/0] {4}
styler:::create_tree("a %>% b()")
#>                                              levelName
#> 1  ROOT (token: short_text [lag_newlines/spaces] {id})
#> 2   °--expr:  [0/0] {11}                              
#> 3       ¦--expr:  [0/1] {3}                           
#> 4       ¦   °--SYMBOL: a [0/0] {1}                    
#> 5       ¦--SPECIAL-PIPE: %>% [0/1] {2}                
#> 6       °--expr:  [0/0] {9}                           
#> 7           ¦--expr:  [0/0] {6}                       
#> 8           ¦   °--SYMBOL_FUNCTION_CALL: b [0/0] {4}  
#> 9           ¦--'(': ( [0/0] {5}                       
#> 10          °--')': ) [0/0] {7}
```

The first needs to be transformed to the second example, can be detected by working on the level of the nested parse data that contains the `SPECIAL-PIPE` symbol and looking into the children (one level deep).

lorenzwalthert:
Can do without checking for "." as suggested in #81 by @krlmlr, but does not necessarily make things simpler, mainly because we used to check at a different nest, that is, we checked in the nest where `token_before` is "SPECIAL-PIPE", so we did not use the `reduce` approach since these nests only had one row. Closes #81.


R/rules-other.R
```diff
@@ -1,17 +1,27 @@
...
-    (pd$token == "SYMBOL") & (pd$text != ".")
-  if (!any(has_no_brackets)) return(pd)
-  new_pos_ids <- create_pos_ids(pd, 1, after = TRUE, n = 2)
-  new_pd <- create_tokens(
-    tokens = c("'('", "')'"), texts = c("(", ")"), pos_ids = new_pos_ids,
-    lag_newlines = rep(0, 2), parents = create_parent_id(pd))
-  pd <- bind_rows(pd, new_pd)
+  is_pipe <- pd$token == "SPECIAL-PIPE"
+  if (any(is_pipe, na.rm = TRUE)) {
+    pd <- reduce(which(is_pipe), add_brackets_in_pipe_one, .init = pd)
```
krlmlr: Would the `reduce()` call also work if `which(is_pipe)` has length zero?

R/utils.R
```diff
@@ -49,3 +49,13 @@ assert_text <- function(text) {
   }
   text
 }
+
+#' Find the index of the next non-comment in a parse table
+#' @param pd A parse table.
+#' @param pos The position of the token to start the search from.
+#' @importFrom rlang seq2
+next_non_comment <- function(pd, pos) {
+  if (pos >= nrow(pd)) return(integer(0))
```
krlmlr: This statement seems redundant.

R/rules-other.R
lorenzwalthert: Yes, thanks.

R/utils.R
lorenzwalthert: Yes, in the current use context it seems, so I can remove it. Just wanted to add safety.

lorenzwalthert: Yes, in the current use context it seems, so I can remove it. Just wanted to add safety.

Issue #257, opened 2017-10-24
Code in braces added to "if" not indented properly


krlmlr:
``` r
styler::style_text("if (TRUE)\na")
#> if (TRUE) {
#>   a
#> }
styler::style_text("if (TRUE)\na <- b")
#> if (TRUE) {
#>   a <- b
#>   }
```

``` r
styler::style_text("if (TRUE)\ncall(\narg\n)")
#> if (TRUE) {
#>   call(
#>   arg
#>   )
#> }
```

Maybe we need to add braces before doing other transformations?

lorenzwalthert:
I think the problem is that `+` (and other tokens) causes indention for all subsequent tokens in the nest. This was fine until now because braces were never on the same nest as those indention-causing operators.
``` r
reprex::reprex_info()
#> Created by the reprex package v0.1.1.9000 on 2017-10-25

styler:::create_tree("
if (x >= 52) {
  2 +1 \n } else {\n 5
}")
#>                                              levelName
#> 1  ROOT (token: short_text [lag_newlines/spaces] {id})
#> 2   °--expr:  [0/0] {41}                              
#> 3       ¦--IF: if [0/1] {3}                           
#> 4       ¦--'(': ( [0/0] {4}                           
#> 5       ¦--expr:  [0/0] {11}                          
#> 6       ¦   ¦--expr:  [0/1] {7}                       
#> 7       ¦   ¦   °--SYMBOL: x [0/0] {5}                
#> 8       ¦   ¦--GE: >= [0/1] {6}                       
#> 9       ¦   °--expr:  [0/0] {9}                       
#> 10      ¦       °--NUM_CONST: 52 [0/0] {8}            
#> 11      ¦--')': ) [0/1] {10}                          
#> 12      ¦--expr:  [0/1] {26}                          
#> 13      ¦   ¦--'{': { [0/2] {13}                      
#> 14      ¦   ¦--expr:  [1/1] {21}                      
#> 15      ¦   ¦   ¦--expr:  [0/1] {16}                  
#> 16      ¦   ¦   ¦   °--NUM_CONST: 2 [0/0] {15}        
#> 17      ¦   ¦   ¦--'+': + [0/0] {17}                  
#> 18      ¦   ¦   °--expr:  [0/0] {19}                  
#> 19      ¦   ¦       °--NUM_CONST: 1 [0/0] {18}        
#> 20      ¦   °--'}': } [1/0] {24}                      
[...]

PR #263, opened 2017-10-25
Wrap expr in expr before enclosing with curly braces

lorenzwalthert:
Found a way around it as outlined in #257. However, the problem that `token_before` and `token_after` get outdated after tokens were inserted persists. It's kind of a chicken and egg problem because if we first insert the tokens, we need to already have the line breaks set for which we need `token_before` and `token_after` (at least in the current code, and it might be necessary to do so to access tokens from neighbouring nests). Anyways I think we should merge that quickly since it seems to be substantial bug.

The column `parent` in the nested parse table also get's outdated with this approach, but I think we can consider removing it after we computed the nested parse data since it appears to me to be not of any further use.


lorenzwalthert:
See #77 for an earlier (short) discussion on the ordering of the transformations.

lorenzwalthert:
Yes, `parent` and `id` removed altogether worked out.

PR #264, opened 2017-10-26
Account for dependency renaming (utf8 changed to enc)

lorenzwalthert:
Reference: https://github.com/krlmlr/enc/issues/17#issuecomment-339588117

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -10,7 +10,7 @@ Imports:
     rlang,
     rprojroot,
     tibble,
-    utf8,
+    enc,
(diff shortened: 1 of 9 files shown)


DESCRIPTION
```diff
@@ -10,7 +10,7 @@ Imports:
     rlang,
     rprojroot,
     tibble,
-    utf8,
+    enc,
```
krlmlr: Can you please keep sort order?

lorenzwalthert: Sure

PR #273, opened 2017-11-05
style_files -> vectorized style_file

### LINKED ISSUE #219: Vectorising style_file()
``` r
style_file(rep("test.R", 2))
#> Styling 2 files:
#> test.R. [DONE]
#> test.R. [DONE]
#> No files changed.
#> Warning message:
#> In if (!grepl("\\.[Rr]$", path)) stop(path, " is not a .R file") :
#>   the condition has length > 1 and only the first element will be used
```

lorenzwalthert:
Closes #219. Should we depreciate `style_file()` in favour of `style_files()` or should we rename `style_files()` to `style_file()` and only use the current implementation of `style_file()` internally (e.g. as `style_file_one()`)?


krlmlr:
I prefer the latter option, keep API and rename `style_file()` -> `style_file_one()` .

tests/testthat/test-public_api.R
```diff
@@ -10,12 +10,17 @@ test_that("styler can style directory", {
   expect_false(all(style_dir(testthat_file("public-api", "xyzdir"))))
 })
 
-test_that("styler can style file", {
+test_that("styler can style files", {
```
krlmlr: Do we need to test `character(0)` as argument, or a `TRUE` return value if contents change?

lorenzwalthert: Not sure if I understand you correctly. Well, the idea was just to check whether calling the function causes an error or not. Since the files were already styled, we don't expect any changes. So you wanted to test whether the output of `style_file()` is `character(0)`?

krlmlr: Seems that `style_file()` returns a logical vector, but the test only tests `FALSE` return values. Is there a test for a `TRUE` return value? If not, we might need one.

Likewise, empty input should lead to `logical(0)` .

lorenzwalthert: Yes we have never explicitly tested for a return value other than `TRUE` because the majority of our tests works with `test_collection()` that transforms an input with a transformer function (in our case `style_text()`) and compares the output to a reference. 
Do you think we need a test for  it? If yes, we'd probably best write a function that takes as input the path of a mal-styled file, copies it to a temp directory, styles it, saves the direct return value of `style_file()`and then discards the files in the temp directory. I just hoped to get around that because it may bear additional problems with paths for `covr` or `testthat` for limited gain, but if you want, I can implement that.

krlmlr: No, that's not a priority, I was just wondering when looking at the tests.

PR #276, opened 2017-11-08
Fix eq assign parsing

### LINKED ISSUE #259: if-else with '=' assignment, wrong curly placement
Hi, I didn't find a previous issue about this.

I have this (working) code: 
```
x = 5

if(x >= 5)
  y = TRUE else 
    y = FALSE
```
With `style_active_region()` producing: 
```
if (x >= 5) {
  y
} <- TRUE else
  y <- FALSE
```
It breaks my code here.

lorenzwalthert:
Closes #259. As outlined in #259, we

* Walk all nests
* Within one nest, we split the nest into sub-nests that contain one assignment expression (e.g. `a = b = c`) at most.
* Within every such sub-nest, we relocate the assignment expression.

However, we only do that if there is an `EQ_ASSIGN` token in the flat parse data since the whole procedure is rather costly and we can expect the probability that there is an `EQ_ASSIGN` in the parse data to be small.


R/relevel.R
```diff
@@ -78,3 +81,128 @@ wrap_expr_in_expr <- function(pd) {
...
+
+#   ____________________________________________________________________________
+#   Relocate EQ_ASSIGN                                                      ####
+
+#' Relocate the expressions containing the token `EQ_ASSIGN` within the nested
+#' parse table
+#'
+#' Although syntactically identical, [utils::getParseData()] does not produce
+#' the same hierarchy of the parse table (parent and id relationship) for `<-`
+#' and `=` (See 'Examles').
```
krlmlr: Examples

R/relevel.R
```diff
@@ -78,3 +81,128 @@ wrap_expr_in_expr <- function(pd) {
...
+#' to the `EQ_ASSIGN` token occuring before them, except the token right before
+#' `EQ_ASSING` already belongs to the `EQ_ASSING` after it.
+#' @param pd A parse table.
+find_block_id <- function(pd) {
+  is_eq_assign <- which(pd$token == "EQ_ASSIGN")
+  eq_belongs_to_block <-
+    cumsum((is_eq_assign - lag(is_eq_assign, default = is_eq_assign[1])) > 2)
+  empty_seq <- rep(0, nrow(pd))
+  empty_seq[lead(pd$token == "EQ_ASSIGN", default = FALSE)] <- eq_belongs_to_block
+  block_id <- cumsum(empty_seq)
```
krlmlr: Is this the value this function returns?

R/relevel.R
```diff
@@ -78,3 +81,128 @@ wrap_expr_in_expr <- function(pd) {
...
+    add_line_col_to_wrapped_expr() %>%
+    remove_attributes(c(
+      "multi_line", "indention_ref_pos_id",
+      "newlines", "indent", "spaces", "lag_newlines"
+    ))
+  eq_expr$id <- NA
+  eq_expr$parent <- NA
+  non_eq_expr <- pd[-eq_ind,]
+  pd <- bind_rows(eq_expr, non_eq_expr) %>%
+    arrange(pos_id)
```
krlmlr: Same here?

R/relevel.R
```diff
@@ -78,3 +81,128 @@ wrap_expr_in_expr <- function(pd) {
...
+#'
+#' Two assignment tokens `EQ_ASSIGN` belong to the same block if they are not
+#' separated by more than one token. Token between `EQ_ASSIGN` tokens belong
+#' to the `EQ_ASSIGN` token occuring before them, except the token right before
+#' `EQ_ASSING` already belongs to the `EQ_ASSING` after it.
+#' @param pd A parse table.
+find_block_id <- function(pd) {
+  is_eq_assign <- which(pd$token == "EQ_ASSIGN")
+  eq_belongs_to_block <-
+    cumsum((is_eq_assign - lag(is_eq_assign, default = is_eq_assign[1])) > 2)
```
krlmlr: `x - lag(x)` isn't much different from `diff(x)`, and `cumsum(diff(x))` is about the same as `x`. Does this need to be that complicated?

R/relevel.R
lorenzwalthert: Yup. Lets add 
```r
block_id
```
So it is explicit.

R/relevel.R
lorenzwalthert: See above

R/relevel.R
lorenzwalthert: Well I checked `diff(x)` from a previous comment from you but the thing is we also need 0 to be the first element. Because with `diff()`, you get `n - 1` elements back, not `n`. So we could do
```r
c(0, diff(x)) # instad of 
x - lag(x, default = x[1])
```
Then, `cumsum(c(0, diff(x)) > 2)` can be simplified by pushing the zero out of `cumsum`, i.e.
`c(0, cumsum(diff(is_eq_assign)) > 2)`
Which is, I admit, simpler.

R/relevel.R
lorenzwalthert: Uups.

R/relevel.R
lorenzwalthert: Or can you even simplify it further? Cause it's not immediately clear to me how one could do that.

R/relevel.R
```diff
@@ -78,3 +81,130 @@ wrap_expr_in_expr <- function(pd) {
...
+#' Find the block to which a token belongs
+#'
+#' Two assignment tokens `EQ_ASSIGN` belong to the same block if they are not
+#' separated by more than one token. Token between `EQ_ASSIGN` tokens belong
+#' to the `EQ_ASSIGN` token occuring before them, except the token right before
+#' `EQ_ASSING` already belongs to the `EQ_ASSING` after it.
+#' @param pd A parse table.
+find_block_id <- function(pd) {
+  is_eq_assign <- which(pd$token == "EQ_ASSIGN")
+  eq_belongs_to_block <- c(0, cumsum(diff(is_eq_assign)) > 2)
```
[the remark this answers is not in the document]
lorenzwalthert: uups yes.

R/relevel.R
```diff
@@ -78,3 +81,130 @@ wrap_expr_in_expr <- function(pd) {
...
+
+  empty_seq <- rep(0, nrow(pd))
+  empty_seq[lead(pd$token == "EQ_ASSIGN", default = FALSE)] <- eq_belongs_to_block
```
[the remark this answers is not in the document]
lorenzwalthert: it's a minus 😄 .

PR #279, opened 2017-11-18
Fix token insertion / comment interaction

### LINKED ISSUE #278: Comments in if clause without braces not recognised correctly
``` r
reprex::reprex_info()
#> Created by the reprex package v0.1.1.9000 on 2017-11-08

  styler::style_text(
"if (!inherits(geom, 'Geom')) # for geoms that return more than one thing
  a()
")
#> Error in if (has_if_without_curly) {: missing value where TRUE/FALSE needed
```
Need to make consistent use of `next_non_comment()` to find correct positions.

lorenzwalthert:
Intended to fix #278. 

* `wrap_if_else_multi_line_in_curly()` now simply forwards to `wrap_if_else_multi_line_in_curly_since_is_if_expr()` if the expression contains an if-clause. This should decrease complexity for the programmer.
* `wrap_if_else_multi_line_in_curly_since_is_if_expr()` calls `wrap_subexpr_in_curly()` to wrap the relevant sub-expression into curly braces for both the if-clause and the else or else-if clause.

@krlmlr can you maybe add other test cases if I missed any?


R/indent.R
```diff
@@ -78,18 +78,21 @@ indent_without_paren_for_while_fun <- function(pd, indent_by) {
 #' @describeIn update_indention Is used to indent if and if-else statements.
 #' @importFrom rlang seq2
 indent_without_paren_if_else <- function(pd, indent_by) {
+  expr_after_if_while <- next_non_comment(pd, 4)
   has_if_without_curly <-
-    pd$token[1] %in% c("IF", "WHILE") && pd$child[[5]]$token[1] != "'{'"
+    pd$token[1] %in% c("IF", "WHILE") && pd$child[[expr_after_if_while]]$token[1] != "'{'"
```
krlmlr: The `"WHILE"` here seems inconsistent with the function name.

R/rules-other.R
```diff
@@ -26,26 +26,71 @@ add_brackets_in_pipe_one <- function(pd, pos) {
...
-    !is_curly_expr(pd$child[[5]])) {
-    pd$lag_newlines[5] <- 0L
+  if (is_cond_expr(pd)) {
+    pd <- wrap_if_else_multi_line_in_curly_since_is_if_expr(pd, indent_by)
+  }
+    pd
+}
+
+
+wrap_if_else_multi_line_in_curly_since_is_if_expr <- function(pd, indent_by = 2) {
```
krlmlr: I didn't understand the function name when skimming through it. Maybe extract `wrap_if_multi_line_in_curly()` and `wrap_else_multi_line_in_curly()`and then write

```r
wrap_if_else_multi_line_in_curly <- function(pd, indent_by = 2) {
  if (is_cond_expr(pd)) {
    pd <- wrap_if_multi_line_in_curly(pd, indent_by)
    pd <- wrap_else_multi_line_in_curly(pd, indent_by)
  }
  pd
}
```

krlmlr:
This generates a full set of test cases. Do we need newlines to make sure the transformer is triggered?

``` r
gen <- function(x) {
  if (length(x) == 0) ""
  else {
    c(
      paste0(x[1], gen(x[-1])),
      paste0(x[1], " # comment\n", paste(x[-1], collapse = ""))
    )
  }
}

collapse <- function(x) paste(x, collapse = "\n\n")

cat(collapse(gen(c("if", "(", "TRUE", ")", "NULL"))))
#> if(TRUE)NULL
#> 
#> if(TRUE)NULL # comment
#> 
#> 
#> if(TRUE) # comment
#> NULL
#> 
#> if(TRUE # comment
#> )NULL
#> 
#> if( # comment
#> TRUE)NULL
#> 
#> if # comment
#> (TRUE)NULL
cat(collapse(gen(c("if", "(", "TRUE", ")", "NULL", " else", " NULL"))))
#> if(TRUE)NULL else NULL
#> 
#> if(TRUE)NULL else NULL # comment
#> 
#> 
#> if(TRUE)NULL else # comment
#>  NULL
#> 
#> if(TRUE)NULL # comment
#>  else NULL
#> 
#> if(TRUE) # comment
#> NULL else NULL
#> 
#> if(TRUE # comment
#> )NULL else NULL
#> 
#> if( # comment
#> TRUE)NULL else NULL
#> 
#> if # comment
#> (TRUE)NULL else NULL
cat(collapse(gen(c("if", "(", "TRUE", ")", "NULL", " else", " if", "(", "FALSE", ")", "NULL", " else", " NULL"))))
#> if(TRUE)NULL else if(FALSE)NULL else NULL
#> 
#> if(TRUE)NULL else if(FALSE)NULL else NULL # comment
#> 
#> 
#> if(TRUE)NULL else if(FALSE)NULL else # comment
#>  NULL
#> 
#> if(TRUE)NULL else if(FALSE)NULL # comment
#>  else NULL
#> 
#> if(TRUE)NULL else if(FALSE) # comment
#> NULL else NULL
#> 
#> if(TRUE)NULL else if(FALSE # comment
#> )NULL else NULL
#> 
#> if(TRUE)NULL else if( # comment
#> FALSE)NULL else NULL
#> 
#> if(TRUE)NULL else if # comment
#> (FALSE)NULL else NULL
#> 
#> if(TRUE)NULL else # comment
#>  if(FALSE)NULL else NULL
#> 
#> if(TRUE)NULL # comment
#>  else if(FALSE)NULL else NULL
#> 
#> if(TRUE) # comment
#> NULL else if(FALSE)NULL else NULL
#> 
#> if(TRUE # comment
#> )NULL else if(FALSE)NULL else NULL
#> 
#> if( # comment
#> TRUE)NULL else if(FALSE)NULL else NULL
#> 
#> if # comment
#> (TRUE)NULL else if(FALSE)NULL else NULL
```

lorenzwalthert:
Ok, thanks for generating the test cases. I will incorporate them in a new test and try to ensure they work.

lorenzwalthert:
Ok, now all valid test cases seem to work. There were a few that are invalid, I commented them out. I just c/p your code above, probably should do some documentation at some point.

R/rules-other.R
lorenzwalthert: Yes, that makes sense. It's just that I thought it was simple but it turned out to be quite complicated with all the edge cases... -.-

R/indent.R
```diff
@@ -78,18 +78,21 @@ indent_without_paren_for_while_fun <- function(pd, indent_by) {
 #' @describeIn update_indention Is used to indent if and if-else statements.
 #' @importFrom rlang seq2
 indent_without_paren_if_else <- function(pd, indent_by) {
+  expr_after_if_while <- next_non_comment(pd, which(pd$token == "')'")[1])
```
krlmlr: So, why is the function called `indent_without_paren_if_else()`, and the code is looking for `"IF"` and `"WHILE"` tokens?

lorenzwalthert: Yes, that's inconsistent. I think it's no longer needed since the function `indent_without_paren_for_while_fun()` handles the `"WHILE"` case.

lorenzwalthert:
Maybe we also need to test for `strict = FALSE`?

R/indent.R
lorenzwalthert: Related: Since we now change 
```r
if (TRUE)
  return(call(a))
```
Into 
```r
if (TRUE) {
  return(call(a))
}
```
We should probably also change 
```r
while (FALSE)
  xyz(v)
```
Into
```r
while (FALSE) {
  xyz(v)
}
```
The same applies to `for`. Should we open an issue and postpone it?

krlmlr:
If you expect the output to be different, this sounds like a good idea.

lorenzwalthert:
Yes it should be quite different since no braces should be added.

R/indent.R
krlmlr: New issue and postponing sounds great. Do we fix the inconsistency in this PR?

krlmlr: Oh, just saw the commit. Thanks!

lorenzwalthert:
Ok, tests added

R/utils.R
```diff
@@ -55,6 +55,21 @@ assert_text <- function(text) {
...
   setdiff(candidates, which(pd$token == "COMMENT"))[1]
 }
+
+#' Find the index of the last comment in the sequence of comments-only tokens
+#' after the token that has position `pos` in `pd`.
+#' @param pd A parse table.
+#' @param pos The position of the token to start the search from.
+extend_if_comment <- function(pd, pos) {
+  if (pos == nrow(pd)) return(pos)
+  if (pd[pos + 1,]$token == "COMMENT") {
```
krlmlr: Probably not important here, but in general `pd$token[pos + 1]` requires a lot less work. Maybe we can search later for `,]` and `, ]` to find this and other instances?

lorenzwalthert:
Not sure whether you spotted that but because of the structure of the nested parse table, end-of-expression comments are located one level up in the nest, so it's impossible to include them within the braces I think.
Hence
```r
if (TRUE)
  4 else 
  44 # comment
```
Becomes
```r
if (TRUE) {
  4 
  } else {
  44 
} # comment
```
Instead of

```r
if (TRUE) {
  4 
  } else {
  44 # comment
} 
```

Here you see why:
``` r
reprex::reprex_info()
#> Created by the reprex package v0.1.1.9000 on 2017-11-21

code <- 
"if (TRUE)
    4 else 
44 # "
styler:::create_tree(code)
#>                                                  levelName
#> 1  ROOT (token: short_text [lag_newlines/spaces] {pos_id})
#> 2   ¦--expr:  [0/1] {1}                                   
#> 3   ¦   ¦--IF: if [0/1] {2}                               
#> 4   ¦   ¦--'(': ( [0/0] {3}                               
#> 5   ¦   ¦--expr:  [0/0] {5}                               
#> 6   ¦   ¦   °--NUM_CONST: TRUE [0/0] {4}                  
#> 7   ¦   ¦--')': ) [0/4] {6}                               
#> 8   ¦   ¦--expr:  [1/1] {8}                               
#> 9   ¦   ¦   °--NUM_CONST: 4 [0/0] {7}                     
#> 10  ¦   ¦--ELSE: else [0/0] {9}                           
#> 11  ¦   °--expr:  [1/0] {11}                              
#> 12  ¦       °--NUM_CONST: 44 [0/0] {10}                   
#> 13  °--COMMENT: #  [0/0] {12}
styler::style_text(code)
#> if (TRUE) {
#>   4
#> } else {
#>   44
#> }  #
```

R/utils.R
lorenzwalthert: So that's for speed?
``` r
reprex::reprex_info()
#> Created by the reprex package v0.1.1.9000 on 2017-11-21

library("tidyverse")
#> Loading tidyverse: ggplot2
#> Loading tidyverse: tibble
#> Loading tidyverse: tidyr
#> Loading tidyverse: readr
#> Loading tidyverse: purrr
#> Loading tidyverse: dplyr
#> Conflicts with tidy packages ----------------------------------------------
#> filter(): dplyr, stats
#> lag():    dplyr, stats
tbl <- as_tibble(mtcars)
microbenchmark::microbenchmark(tbl$cyl[1], tbl[1, ]$cyl)
#> Unit: microseconds
#>          expr     min      lq      mean   median       uq       max neval
#>    tbl$cyl[1]  47.232  50.143  62.46642  60.7610  63.6915   190.264   100
#>  tbl[1, ]$cyl 254.001 298.076 569.41824 318.3655 338.8025 13161.350   100
```

lorenzwalthert: I just checked and I found `, ]$` and `,]$` only twice and I changed it as part of this PR to your suggested style.

krlmlr:
I'm fine with putting the comment after the terminal brace.

PR #285, opened 2017-11-19
Clarify labelling strategy

### LINKED ISSUE #284: Clarify labelling strategy
In particular, issues that are labeled with "STATUS: POSTPONED" are closed but not resolved.

lorenzwalthert:
Closes #284. @krlmlr I created [saamwerk](https://github.com/lorenzwalthert/saamwerk) and described the labelling strategy there since I think it might be useful in other places too. I hope that's ok with you. I think it would be nice to integrate these recommendations in some existing open source and I tried https://opensource.guide but they don't accept self nominations. Any ideas where I could try to submit?

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -9,3 +9,4 @@
 ^\.travis\.yml$
 ^docs$
 ^_pkgdown\.yml$
+CONTRIBUTING.md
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
Yes, that's right. Yes we may want to change the policy after the CRAN release.

Issue #140, opened 2017-08-20
Adding roundtrip for nested processing


lorenzwalthert:
Might be implemented similar to depreciated flat roundtrip
```r
#' Verify parse data modifications
#'
#' @description Check whether serializing the parse data results in the same
#' number of lines as the initial data that should be styled.
#' @param pd_flat A parse table.
#' @param text A character vector with the initial text to compare against.
#' @return If the verification is successful, `pd` is returned, with empty
#'   lines at the end of `text` stripped. \cr
#'   Otherwise, an error is thrown.
#' @seealso [serialize_parse_data_flat()]
verify_roundtrip <- function(pd_flat, text) {
  roundtrip <- serialize_parse_data_flat(pd_flat)

  if (length(roundtrip) < length(text)) {
    stopifnot(text[-seq_along(roundtrip)] == "")
    text <- text[seq_along(roundtrip)]
  }

  stopifnot(identical(text, roundtrip))
  text
}
```

lorenzwalthert:
Reference: #85.

lorenzwalthert:
Related: #216.

lorenzwalthert:
I wonder how that should be implemented. Because with reference to #85, I think it would make sense to not just test parsing and serialising (because that almost certainly works now), but also whether the code is still valid after tokens were inserted, compare #259. So a simple parse statement of the output could check that.

lorenzwalthert:
Can we postpone this?

krlmlr:
Sure.

lorenzwalthert:
Reference: #280.

krlmlr:
Have we implemented roundtrip checking by now?

lorenzwalthert:
In #287, we implemented verifying whenever possible (`can_verify_roundtrip`), which seems to be the case whenever we don't have any token manipulators.

lorenzwalthert:
Reference: #368.

PR #287, opened 2017-11-21
Add roundtrip

### LINKED ISSUE #280: check parse data does not change after styling for scope < "tokens"
See https://github.com/r-lib/styler/issues/278#issuecomment-345463340
``` r
    identical(parse(text = "a+1"), parse(text = "a + 1"))
    #> [1] TRUE
```

lorenzwalthert:
Closes #280. @krlmlr do you see a better way to obtain scope than via checking whether `transformers$token` is `NULL`?


R/transform-files.R
```diff
@@ -154,5 +154,43 @@ apply_transformers <- function(pd_nested, transformers) {
...
+#' styler:::verify_roundtrip(
+#'   "a+1",
+#'   "a + 1 # comments are dropped",
+#'   tidyverse_style(scope = "line_breaks")
+#' )
+#' \dontrun{
+#' styler:::verify_roundtrip("a+1", "b - 3", tidyverse_style(scope = "line_breaks"))
+#' }
+verify_roundtrip <- function(old_text, new_text, transformers) {
+  if (is.null(transformers$token)) {
```
krlmlr: I don't see a better way, but we could make this explicit by extracting this to a function:

```r
  if (can_verify_roundtrip(transformers)) {
    ...
  }
```

lorenzwalthert:
You mean yielding a warning (or a message?) if scope > "line_breaks"? Just for consistency, there is "carefully inspect the results" message for `style_dir()` and friends (R/transform_files#25).

krlmlr:
We might want to skip the "Carefully inspect..." message if scope <= "line_breaks".

lorenzwalthert:
Yes. And also, we don't want to give a message if nothing has changed. The best solution would be to only show the message if the parsed expressions are also not the same. However, this is tricky because we still want so show whether or not a file has changed at all (with the star). So I suggest to show the message whenever the code has changed (in terms of exact comparison) *and* the scope is > "line_breaks".

krlmlr:
Comparing the parsed output in addition is a good idea. Can we show

- two stars if the parsed expressions have changed,
- one star if the file has changed (but not the parsed expressions) and
- no star if the file is unchanged?

lorenzwalthert:
We could I guess... It adds complexity and needs testing though.

lorenzwalthert:
I agree that it would be a nice feature though, I just wonder if I find time to implement it before the CRAN release....

lorenzwalthert:
As for today, let me push what I have.

lorenzwalthert:
See branch [roundtrip](https://github.com/lorenzwalthert/styler/tree/roundtrip).

lorenzwalthert:
The problem is that we can't easily achieve what was outlined here because  `enc::transform_lines_enc()` does only return one boolean that indicates whether the content has changed (exact comparison). We'd have to read and write ourself and do error handling to also check whether the expression remains the same. I think the gain is limited compared to the effort required. So I propose to stick with the simpler solution outlined above.

krlmlr:
Works for me.

lorenzwalthert:
Reference: #140.

lorenzwalthert:
Reference: #368.

PR #289, opened 2017-11-25
Roundtrip and CLI

### LINKED ISSUE #288: Better CLI
I propose:
- Use the cli package to create nice ticks next to the styled files instead of the red, aggressive [DONE] if the styling did not change, an info icon if the styling changed and an warning icon if the styling failed. At the end of the styling, we can add a summary. Just like this:
```
Styling  1  files:
 test-start_line.R ✔ 
────────────────────────────────────────
Status	Count	Legend 
✔ 	1	File unchanged.
ℹ 	0	File changed.
✖ 	0	Styling threw an eror.
```
- Ideally we would also add one of the circling dots but I have not figured out how to do that quickly. Maybe you @krlmlr know how that works form testthat?
- Instead of a logical vector, `style_file()` and friends should (invisibly) return a tibble that contains the paths to the files as well.
```r
return_value
#> # A tibble: 1 x 2
#>                               file changed
#>                              <chr>   <lgl>
#> 1 tests/testthat/test-start_line.R   FALSE
```

Any further suggestions @krlmlr?

lorenzwalthert:
Closes #288, implements https://github.com/r-lib/styler/pull/287#issuecomment-346693621.


R/communicate.R
```diff
@@ -0,0 +1,27 @@
+#' Communiate a warning if necessary
+#'
+#' If roundtrip verification was not possible, issue a warning to review the
+#' changes carefully.
+#' @param changed Boolean with indicating for each file whether or not it has
+#'   been changed.
+#' @inheritParams can_verify_roundtrip
+communicate_warning <- function(changed, transformers) {
+  if (any(changed, na.rm = TRUE) & !can_verify_roundtrip(transformers)) {
+    cat("Please review the changes carefully!")
```
lorenzwalthert: Is it ok to use `cat()` or do you prefer `warning()`. I just felt red is not the right color.

lorenzwalthert:
Any ideas why the appveoyr build fails at the point where it compares the printed console output 
```
Styling  1  files:
 test-start_line.R ✔ 
────────────────────────────────────────
Status	Count	Legend 
✔ 	1	File unchanged.
ℹ 	0	File changed.
✖ 	0	Styling threw an eror.
```
With a reference value? I guess it has to do with the encoding. But using class `utf8` does not help.

R/transform-files.R
```diff
@@ -200,3 +202,14 @@ verify_roundtrip <- function(old_text, new_text) {
     stop(msg, call. = FALSE)
   }
 }
+
+#' Check whether two expressions are identical
+#'
+#' @param old_text The initial expression in its character representation.
+#' @param new_text The styled expression in its character representation.
+expressions_are_identical <- function(old_text, new_text) {
+    identical(
```
krlmlr: Indent?

R/communicate.R
krlmlr: We can add color with crayon if necessary.

R/transform-files.R
```diff
@@ -11,19 +11,15 @@
 transform_files <- function(files, transformers) {
   transformer <- make_transformer(transformers)
   max_char <- min(max(nchar(files), 0), 80)
-  if (length(files) > 0) {
-    message("Styling ", length(files), " files:")
+  if (length(files) < 1) {
```
krlmlr: Please check the comparison operator.

R/ui.R
```diff
@@ -142,7 +142,10 @@ style_file <- function(path,
                         ...,
                         style = tidyverse_style,
                         transformers = style(...)) {
-  map_lgl(path, style_file_one, ..., style = style, transformers = transformers)
+  changed <- map_lgl(path,
+    style_file_one, ..., style = style, transformers = transformers
+  )
+  invisible(tibble(file = path, changed = changed))
```
krlmlr: Is this return value used anywhere?

R/communicate.R
```diff
@@ -0,0 +1,27 @@
+#' Communiate a warning if necessary
+#'
+#' If roundtrip verification was not possible, issue a warning to review the
+#' changes carefully.
+#' @param changed Boolean with indicating for each file whether or not it has
+#'   been changed.
+#' @inheritParams can_verify_roundtrip
+communicate_warning <- function(changed, transformers) {
+  if (any(changed, na.rm = TRUE) & !can_verify_roundtrip(transformers)) {
```
krlmlr: `&&` better communicates intent here.

krlmlr:
Would you like to change the default scope to something weaker than `"token"`?

lorenzwalthert:
I am not sure. Would you? It took us quite a while to implement everything as it is now and if we lower the default scope, I think a lot of people will not take the time to set it manually to `"tokens"`. Also, we have the warning and many tests, so I feel like that should be sufficient.

R/ui.R
lorenzwalthert: No, I just found `style_file_one()` is redundant anyways with the current implementation of `transform_files()`. Must have overlooked that at some point. I will make the changes in `transform_files()`

krlmlr:
I value the effort that it took to patch the parse data. However, if we don't change tokens, we can automatically verify correctness. This gives us an extra safety net.

I'm fine either way, please decide.

krlmlr:
AppVeyor: Checking Unicode symbols against known output on Windows is a pain. It's ok to just skip these tests on Windows.

lorenzwalthert:
I am wondering about how good of an idea it is to set the default to `"line_breaks"` and change it later to `"tokens"` if we styled more tidyverse packages and gained certainty that everything works fine. On the one hand, that would add safety, but the downside is API stability. So I propose to keep it as is and make sure we can address potential bug reports quickly.

R/transform-files.R
lorenzwalthert: uups, thanks.

R/communicate.R
lorenzwalthert: Right...

lorenzwalthert:
Reference: #368.

Issue #291, opened 2017-11-25
allow rmd styling also via style_dir() and style_pkg()


lorenzwalthert:
`style_dir()` and `style_pkg()` should gain a new argument `extension` (that defaults to `c("R")`) to indicate which types of extensions should be styled. It should handle:
* just R files (`extension = "R"` or `extension = c("R")`).
* just Rmd files (`extension = "Rmd"` or `extension = c("Rmd")`).
* both (`extension = c("Rmd", "R")`).

The case should be ignored, i.e. `extension = c("rMd")` should be accepted. Nice to have: a trailing dot is removed (i.e. `extension = c(".Rmd")` also works).

krlmlr:
Do you want to include this in the first CRAN release?

krlmlr:
Oh, I missed the open PR. Fine with me.

krlmlr:
Closed by #292.

PR #292, opened 2017-11-25
Update style_dir()

jonmcalder:
Extension to `style_dir()` to address the first part of #291


R/ui.R
```diff
@@ -100,10 +104,11 @@ style_dir <- function(path = ".",
                       ...,
                       style = tidyverse_style,
                       transformers = style(...),
+                      filetype = "R",
```
lorenzwalthert: can we change that to `c("r")` also?

R/ui.R
```diff
@@ -36,6 +39,7 @@ style_pkg <- function(pkg = ".",
                       ...,
                       style = tidyverse_style,
                       transformers = style(...),
+                      filetype = c("r"),
```
lorenzwalthert: I don't see how this argument does affect the function call, i.e. it is not really used further downstream if I see that correctly.

lorenzwalthert: So was your intend to wait with merging anyways until #293 is resolved?

R/ui.R
jonmcalder: Sorry did the other merge from my phone and I thought you'd changed it already. Happy to change this too.

R/ui.R
jonmcalder: Yeah this was just because the other functions (style_dir & prettify_any) were already inheriting params from here and I figured we'd be adding this to style_pkg later anyway.

R/ui.R
lorenzwalthert: @krlmlr do you prefer to stick with `"R"` everywhere instead of `"r"`? Its kind of a philosophical debate I guess...

R/ui.R
lorenzwalthert: Ok, lets wait with merging then.

R/ui.R
krlmlr: Can we use `"R"` everywhere? It doesn't seem to affect the code, but the default file extension is `.R` and not `.r`, and we should be consistent here.

jonmcalder: I agree that the default here and elsewhere should be uppercase. I think @lorenzwalthert wanted to emphasize that this argument can be/is a vector so wanted to use c("R") instead. I think the case inconsistency was incidental/accidental.

krlmlr: I thought `c("R")` was an artifact... I feel it's enough if we just mention that the argument also accepts vectors in the documentation.

lorenzwalthert: ok, fine.

jonmcalder:
Ok, this should be ready for further review now.

R/ui.R
```diff
@@ -18,6 +18,9 @@ NULL
 #' @param transformers A set of transformer functions. This argument is most
 #'   conveniently constructed via the `style` argument and `...`. See
 #'   'Examples'.
+#' @param filetype Vector of file extensions indicating which filetypes should
+#'   be styled. Case is ignored, and the `.` is optional, e.g. c(".R", ".Rmd")
```
krlmlr: Does the code need backticks?

R/set-assert-args.R
```diff
@@ -0,0 +1,73 @@
...
+#'
+#' Check whether one or more tokens exist and have a unique token-text mapping
+#' @param tokens Tokens to check.
+assert_tokens <- function(tokens) {
+  invalid_tokens <- tokens[!(tokens %in% lookup_tokens()$token)]
+  if (length(invalid_tokens) > 0) {
+    stop(
+      "Token(s) ", paste0(invalid_tokens, collapse = ", "), " are invalid. ",
+      "You can lookup all valid tokens and their text ",
+      "with styler:::looup_tokens(). Make sure you supply the values of ",
```
krlmlr: looup -> lookup

R/ui.R
```diff
@@ -37,24 +40,47 @@ style_pkg <- function(pkg = ".",
...
+    )
+  }
 
-  files <- setdiff(c(r_files, test_files), exclude_files)
+  if ("\\.rmd" %in% filetype) {
+    vignette_files <- dir(
+      path = "vignettes", pattern = "\\.rmd$", ignore.case = TRUE,
+      recursive = TRUE, full.names = TRUE
+    )
+    readme <- dir(pattern = "readme\\.rmd", ignore.case = TRUE)
```
krlmlr: Needs `^` and `$` ?

R/ui.R
```diff
@@ -37,24 +40,47 @@ style_pkg <- function(pkg = ".",
...
-  r_files <- grep("/RcppExports[.]R$", r_files, invert = TRUE, value = TRUE)
-  test_files <- dir(
-    path = "tests/testthat", pattern = "[.][rR]$",
-    recursive = TRUE, full.names = TRUE
-  )
+  filetype <- set_and_assert_arg_filetype(filetype)
+  r_files <- test_files <- data_raw_files <- vignette_files <- readme <- NULL
+
+  if ("\\.r" %in% filetype) {
+    r_files <- dir(
```
krlmlr: The `paths` argument to `dir()` seems to be vectorized, not sure about older versions of R.

R/ui.R
```diff
@@ -37,24 +40,47 @@ style_pkg <- function(pkg = ".",
...
+    prettify_local(transformers, filetype, exclude_files)
+  )
   invisible(changed)
 }
 
-prettify_local <- function(transformers, exclude_files) {
-  r_files <- dir(
-    path = "R", pattern = "[.][rR]$", recursive = TRUE, full.names = TRUE
-  )
+prettify_local <- function(transformers, filetype, exclude_files) {
```
krlmlr: I'm not sure the function name still matches the intent.

tests/testthat/public-api/xyz-r-and-rmd-dir/subdir/random-rmd-script-in-subdir.Rmd
```diff
@@ -0,0 +1,30 @@
+---
+title: "random-rmd-script"
+author: "Lorenz Walthert"
+date: "11/25/2017"
+output: html_document
+---
+
+```{r setup, include=FALSE}
+1+1
```
krlmlr: Do the tests actually test that styling is applied?

jonmcalder:
Thanks very much for the review @krlmlr! Unfortunately I won't be able to get back to this until this evening (i.e. outside office hours). I hope I'm not holding up your CRAN release?

krlmlr:
No, that's totally fine.

krlmlr:
Thanks for working on this!

tests/testthat/public-api/xyz-r-and-rmd-dir/subdir/random-rmd-script-in-subdir.Rmd
lorenzwalthert: I think we just check whether the file was named in the message about changed files. Should we test for the actual change?

R/ui.R
jonmcalder: Thanks - I had overlooked this. Based on [this source code from 2000](https://github.com/wch/r-source/blob/92142731348d937ec347a56a8b68d34e9dc1d172/src/library/base/man/list.files.Rd), I think it's safe to assume that the paths argument has been vectorized since pre `R 2.0` days, so I'll go ahead and combine these.

tests/testthat/public-api/xyz-r-and-rmd-dir/subdir/random-rmd-script-in-subdir.Rmd
jonmcalder: They don't check whether styling is applied, just which files are being named in the changed files message.  @lorenzwalthert suggested these tests so I kinda assumed we had our bases covered with actual styling tests elsewhere.

jonmcalder:
Hi @krlmlr! Sorry I should have requested a follow up review from you on Wednesday eve after pushing some changes.

I think I have addressed all your comments. The drop in testing coverage is very marginal - but in any case highlights the question of whether or not you feel there is sufficient testing? (please refer to the comments above)

krlmlr:
Thanks again!

jonmcalder:
Happy to help out - another good learning opportunity for me. Thanks @krlmlr!

PR #300, opened 2017-12-12
Update communicate.R

devSJR:
fix typo


lorenzwalthert:
Thanks @devSJR. There are also tests that check the API, so we need to update those as wel. Do you want to do that? Just run
```r
devtools::test(f = "api")
```
on the package and commit the changed reference objects. in `tests/testthat/...`.

devSJR:
Thanks for the offer. Due to time-constraints I won't be able to help you there. Maybe later. Kind regards

lorenzwalthert:
Ok, thanks @devSJR, I created a PR to your `patch-1`branch that fixes the problem. Can you "rebase and merge" this? Then, I will merge this PR (which will contain my commits) into master. This way, we can ensure the master build will continue to pass.

PR #304, opened 2017-12-26
don't re-indent function declarations with scope < "indention"

### LINKED ISSUE #303: Function declarations should not be re-indented for scope < "indention"
`update_indention_ref_fun_dec()` should not be an `indention_modifier` for scope < "indention.

lorenzwalthert:
Closes #303


R/style_guides.R
```diff
@@ -142,8 +142,8 @@ tidyverse_style <- function(scope = "tokens",
 
   indention_modifier <-
     c(
-      update_indention_ref_fun_dec,
-      NULL
+      if (scope >= "indention") update_indention_ref_fun_dec,
+      identity
```
krlmlr: I'm not in love with the use of `identity` here, this sure doesn't help performance. Maybe allow `NULL` and filter out later with `compact()`? Need to use `lst()` instead of `c()`.

The name for the list elements will be the full quoted expression, maybe change l. 138 above to `wrap_if_else_multi_line_in_curly = if (strict) wrap_if_else_multi_line_in_curly else identity` and likewise in other places?

lorenzwalthert: Ok, I like the idea with `purrr::compact()`, `lst()` and naming the arguments here and in other places.

API
```diff
@@ -2,7 +2,7 @@
 
 ## Exported functions
 
-create_style_guide(initialize = default_style_guide_attributes, line_break = NULL, space = NULL, token = NULL, indention = NULL, use_raw_indention = FALSE, reindention = tidyverse_reindention())
+create_style_guide(initialize = lst(default_style_guide_attributes), line_break = NULL, space = NULL, token = NULL, indention = NULL, use_raw_indention = FALSE, reindention = tidyverse_reindention())
```
lorenzwalthert: Should we use `lst()` as a default in an exported function at all?

R/style_guides.R
```diff
@@ -206,7 +204,8 @@ create_style_guide <- function(initialize = default_style_guide_attributes,
     # transformer options
     use_raw_indention,
     reindention
-  )
+  ) %>%
+    map(compact)
```
lorenzwalthert: Since we changed the default values of `create_style_guide`, now all elements of the list are lists and we can map over them with compact. However, this seems a breaking API change to me. If we keep `initialize = default_style_guide_attributes`, we can't map over them (at least not all of them).

lorenzwalthert:
Unfortunately, there was an inconsistency in `create_style_guide()`, which is that all inputs are list of transformers, except `initialize`, which is just a function (and not a list of functions). Should we make a breaking API change here so we can map conveniently and have a more consistent API or not? I suggest we internally convert to a list and keep the API as is. Will push soon.

krlmlr:
It looks more natural to accept a function for the `initialize` argument, but there may be use cases where a list of functions is helpful. Agree to keep the API and use `lst()` internally (and not in the default argument).

lorenzwalthert:
Ok, agree.

PR #308, opened 2017-12-30
Avoid spaces around ^

krlmlr:
following https://github.com/tidyverse/style/pull/46.


lorenzwalthert:
Aren't we changing the defaults in `specify_math_token_spacing()`? Because currently, all the defaults correspond to the tidyverse style guide. Or do you consider that API change as too disruptive?

krlmlr:
I meant to change the default, the tidyverse style guide now contains the rules for `^`. What am I missing?

lorenzwalthert:
I thought we change the default value in the function declaration of `specify_math_token_spacing()`, not in the function call within `tidyverse_style()`.

tests/testthat/math_token_spacing/non_strict_math_spacing_all-out.R
```diff
@@ -1 +1 @@
-1 + +1   -       3 / 23 * 3 ^ 4
+1 + +1   -       3 / 23 * 3 ^4
```
krlmlr: I have no idea why this happens:

``` r
styler::style_text("1 - 3 * 4 ^ 5", strict = FALSE)
#> 1 - 3 * 4 ^ 5
```

lorenzwalthert: That's actually what I'd expect because `strict = FALSE` means no strict implementation, right? Or do you prefer to set the spaces to exactly zero around `^` even if `strict = FALSE`?

lorenzwalthert: I think the reason this might seem inconsistent with the other test is that there is no space around `^` in any of the other tests, but I think that is because in *-in.R files, we always have `x^k`(so no space in between) and hence the tests actually don't test whether `x ^ k` is turned into `x^k`.

krlmlr:
I've updated the code, because we're removing spaces for `:` :

``` r
styler::style_text("1 : 10", strict = FALSE)
#> 1:10
styler::style_text("1 ^ 10", strict = FALSE)
#> 1^10
```

> Created on 2018-02-23 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

lorenzwalthert:
Great. It payed off to reduce some duplication and encapsule common functionality in `style_space_around_token()`.

PR #313, opened 2018-01-04
Only consider R code as code to format

### LINKED ISSUE #312: ignore knitr chunks that use a non-R engine such as sql
styler is choking on a chunk that contains a sql code block.  I would suggest that styler ignore non-R code.

lorenzwalthert:
Closes #312. @krlmlr should we use a package option to allow specifying the regex pattern for the engine ([L69](https://github.com/r-lib/styler/compare/master...lorenzwalthert:fun-dec?expand=1#diff-fe3ee4fefd992c62bde48f05ccd986fbR67))? I.e. if someone creates a custom engine that uses R (as we did for the tidyverse.org post), this code will be ignored for styling with this patch.


lorenzwalthert:
@krlmlr what do you think regarding the package option?

lorenzwalthert:
Ok, I added the engine pattern as a package option.

krlmlr:
Sorry, I missed that. I don't think this should be a package option, but adding more arguments to `style_pkg()` and `style_file()` feels awkward. Can we refactor the function that decides which engines are supported, so that this function just returns a constant for now?

lorenzwalthert:
Ok, I reset hard and implemented it the way I understand you suggested it. What do you think?

PR #318, opened 2018-01-06
Named arguments and line breaks

### LINKED ISSUE #246: Line breaks and named arguments
Line breaks should not be forced after the opening brace of a function call if the call follows a structure like
```r
map(x, c, 
  furher_arg = 3,
  another_arg = 2:4
)
```
Reference: https://github.com/tidyverse/style/issues/39

Hence, when modifying line breaks after `(`, one has to check the structure and only break the line before the first named argument.
More concretely, 
* if expression is a function call,
* and expression spans over multiple lines
* find the first named argument and add a line break before it.
* If no named argument occurs (but there is at least one argument), break after opening brace.

lorenzwalthert:
Closes #246.


lorenzwalthert:
Reference: #342.

PR #322, opened 2018-01-14
remove spaes after !! and !!!

### LINKED ISSUE #321: Styling around Bang-bang
Rule to add space before `!!` needs to be reverted. 
Reference: https://github.com/tidyverse/dplyr/pull/3299#discussion_r161287448

lorenzwalthert:
Closes #321 but does even more because it also remove the space after`!!!`. It seems not so easy to distinguish `!!` and `!!!` since they are on separate levels in the nested parse table and we only have `token_before` and `token_after`, but we also needed `token_before2` and `token_after2` which does not seem to be particularly elegant. An alternative is go inside the child to see what comes afterwards, and set the spacing before processing the child. Also not very nice. Maybe let's remove the space after both `!!` and `!!!` for now.
``` r
styler:::create_tree("call(!!!a)")
#>                                                  levelName
#> 1  ROOT (token: short_text [lag_newlines/spaces] {pos_id})
#> 2   °--expr:  [0/0] {1}                                   
#> 3       ¦--expr:  [0/0] {3}                               
#> 4       ¦   °--SYMBOL_FUNCTION_CALL: call [0/0] {2}       
#> 5       ¦--'(': ( [0/0] {4}                               
#> 6       ¦--expr:  [0/0] {5}                               
#> 7       ¦   ¦--'!': ! [0/0] {6}                           
#> 8       ¦   °--expr:  [0/0] {7}                           
#> 9       ¦       ¦--'!': ! [0/0] {8}                       
#> 10      ¦       °--expr:  [0/0] {9}                       
#> 11      ¦           ¦--'!': ! [0/0] {10}                  
#> 12      ¦           °--expr:  [0/0] {12}                  
#> 13      ¦               °--SYMBOL: a [0/0] {11}           
#> 14      °--')': ) [0/0] {13}
```

Created on 2018-01-14 by the [reprex package](http://reprex.tidyverse.org) (v0.1.1.9000).


lorenzwalthert:
@krlmlr should aim to alter the parse data before nesting it and make `!!` and `!!!` one token? That would allow us to handle `!!` and `!!!` differently (or not touch `!!!`).

lorenzwalthert:
Reference: https://github.com/tidyverse/dplyr/pull/3299#discussion_r161390932

lorenzwalthert:
Reference: https://github.com/tidyverse/style/issues/55

krlmlr:
I could live with looking a few levels into the parse tree, e.g.: `pd$child[[2]]$child[[2]]$token == `"!"` . For now, let's get rid of spaces after `!!!` for now, we can add the distinction later if it's specified in tidystyle.

lorenzwalthert:
Yes, looking into the child is one thing, but manipulating it is even more questionable. Ok, sure.

Issue #329, opened 2018-01-23
style_file() unexpectedly deletes non-ASCII characters on Windows


yutannihilation:
(Sorry, this is a "manual" reprex, as it throws errors on reading/writing to the tempfile.)

```r
tmp <- tempfile(fileext = ".R")

con <- file(tmp, encoding = "UTF-8")

# "1+1" needs spaces
writeLines("# 忍者\n1+1", con)

readLines(con)
#> [1] "# 忍者" "1+1"

styler::style_file(tmp)

#> ----------------------------------------
#> Status	Count	Legend 
#> √ 	0	File unchanged.
#> i 	1	File changed.
#> x 	0	Styling threw an eror.
#> ----------------------------------------
#> Please review the changes carefully!


# the comment "# 忍者" disappeared...
readLines(con)
# [1] ""      "1 + 1"
```

I think the root cause is this line in enc package; `from = "UTF-8"` should not set here as it seems that text is encoded as the default encoding.

https://github.com/krlmlr/enc/blob/883df89f77bb5f8646b9a8fa94cc72120cdf0c81/R/io.R#L59

yutannihilation:
Ah, sorry, I was wrong on this part:

> `from = "UTF-8"` should not set here

`?write_lines_enc` says "It always expects text in the UTF-8 encoding", but this assumption seems to fail on my Windows.

yutannihilation:
I got it. The string returned by `transformer` is not encoded in the UTF-8 encoding here: https://github.com/krlmlr/enc/blob/883df89f77bb5f8646b9a8fa94cc72120cdf0c81/R/io.R#L113

``` r
reprex::reprex_info()
#> Created by the reprex package v0.1.1.9000 on 2018-01-24

utf8_string <- enc2utf8("# 忍者")

tf <- styler:::make_transformer(styler::tidyverse_style())
Encoding(tf(utf8_string))
#> [1] "unknown"
```

To me, while `write_lines_enc()` expects UTF-8 texts, there looks nothing to ensure or check that the result of `fun()` is in the UTF-8 encoding. Is this the problem on enc's side? Or, is it styler's fault that its transformers don't follow the convention of `write_lines_enc()`?

```r
  text <- try_read_lines_enc(path, file_encoding = file_encoding, ok = ok, skipNul = skipNul)
  sep <- detect_sep(path)
  tryCatch(
    {
      new_text <- as.character(fun(text))
[...]

PR #330, opened 2018-01-23
Use the dev version of enc package to fix handling non-ASCII characters on Windows

yutannihilation:
related: #329

### DIFF
--- a/tests/testthat/test-encoding.R
+++ b/tests/testthat/test-encoding.R
@@ -1,11 +1,10 @@
 context("non-ASCII characters are handled properly")
 
 test_that("non-ASCII characters are handled properly", {
-  # c.f. rlang::mut_latin1_locale()
-  locale <- if (.Platform$OS.type == "windows") "English_United States.1252" else "en_US.ISO8859-1"
+  skip_if(.Platform$OS.type != "windows")
 
   withr::with_locale(
-    c(LC_CTYPE = locale),
+    c(LC_CTYPE = "English_United States.1252"),
     {
       tmp <- tempfile(fileext = ".R")
       con <- file(tmp, encoding = "UTF-8")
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
Is this the behaviour of interest here? I think the source needs to be changed to 
```
latin_string <- "Gl\u00fcck + 1"
```

yutannihilation:
Thanks for the hint! Sorry, I compared the result with the wrong string... But, the source should be "ugly" text that needs to be styled, otherwise the file stays unchanged.

lorenzwalthert:
Let's wait until a new version of `enc` is on CRAN that includes https://github.com/krlmlr/enc/pull/18, rerun CI and merge afterwards.

yutannihilation:
Thanks. I wonder if it's better to add `krlmlr/enc` on `Remotes` for now if the next release will not come soon. What do you say?

yutannihilation:
I just added a tweak to skip this test on Windows as suggested on https://github.com/krlmlr/enc/pull/18#discussion_r163591997 (which I've forgot to reflect here).

lorenzwalthert:
Yes, you are right. So people installing via `remotes` or `devtools` from GitHub can get the fix. Do you want to expand this PR in this regard?

yutannihilation:
Sure, I'll include it here 👍

yutannihilation:
Confirmed the tests now pass. I believe this is ready to merge now :)

tests/testthat/test-encoding.R
```diff
@@ -0,0 +1,21 @@
+context("non-ASCII characters are handled properly")
+
+test_that("non-ASCII characters are handled properly", {
+  skip_if(.Platform$OS.type != "windows")
```
lorenzwalthert: Is there a particular reason not to test on unix? If not, I think we can reduce complexity here and test on all platforms I think.

lorenzwalthert: Ah ok, did not see https://github.com/krlmlr/enc/pull/18#discussion_r163591997

lorenzwalthert:
Thanks @yutannihilation.

yutannihilation:
Thanks!

PR #345, opened 2018-02-05
Shebang

### LINKED ISSUE #344: style_file breaks hash-bang statement at top of script
If at the top of a script, there's a line that reads "#!/usr/bin/env Rscript", style_file will change it to

    # !/usr/bin/env Rscript

with a space after the "#". This prevents the script from being executed unless that extra space is removed. Not the end of the world but really annoying.

lorenzwalthert:
Aims to close #344. @jjramsey do you think the following regex is appropriate to identify hash-bang statements? 
```
"^#![^\\s].*\\s[^\\s]+\\b.*$"
```
I first thought about including `Rscript` too but I think commands other than `Rscript` can be used in hash-bangs that can execute R, right? So it would be too specific. Alternatively, I thought about only looking at the first line of a file, but that seems to complicate things. The prosed regex will result in a few false positives though (see example `#!NEED TO REMOVE THIS`, which will be treated as hash-bang statement with the proposed regex).


lorenzwalthert:
Ok, I think we can do it differently (better and simpler). Since a shebang statement will always be on the first line and hence a top-level expression, it will have `pos_id` 1 and we can simplify the regex to 

```
^#!
```

Issue #338, opened 2018-02-03
CRAN Submission schedule for new version of styler


lorenzwalthert:
Since the initial release of styler, some important bugs were fixed: 

*  implicit `dplyr` dependency was removed (#324).
* cascading `=` assignment does no longer cause errors (#328).
* minimal version of `purrr` added to DESCRIPTION (#326).
* An annoying spelling mistake in the output sent to the console after running `style_pkg()` and friends was fixed.
* A bug that resulted in a removal of non-ASCIII characters was fixed in the dependency `enc` the patch is not included in the current CRAN version (as of February 3, 2018).

In addition, there were other enhancements / improvements contributed to styler (most notably #318, #316, #311, #306, #304), and 

Since styler reverse depends on `usethis`, which will become a `devtools` dependency with the next release of `devtools`, we should ensure that by the time of this release, the above bugfixes are included in the CRAN version of styler.

Therefore, I suggest to submit styler `1.0.1` mid February, which seems in accordance with the [CRAN Policy](https://cran.r-project.org/web/packages/policies.html#Re_002dsubmission). @krlmlr Do you agree with this timing and the version number (or should we do `1.1`)? Do you think a new release of `enc` is also required?

krlmlr:
Would be good to release enc by then. Would you mind adding a versioned dependency (>= 0.1-10) and a remote to `DESCRIPTION`?

The timing is fine, too.

krlmlr:
If the API hasn't changed at all, 1.0.1 should be fine.

lorenzwalthert:
Yes, dependency added.

PR #348, opened 2018-02-09
Minimal enc version

lorenzwalthert:


lorenzwalthert:
Reference: https://github.com/r-lib/styler/issues/338

Issue #352, opened 2018-02-20
Wrong indention level of EQ_SUB token


lorenzwalthert:
5 and b should be indented 2 spaces.
``` r
styler::style_text("call(a = \n 5, \nb)", scope = "indention")
#> call(a =
#>     5,
#>   b)
```

Created on 2018-02-21 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

lorenzwalthert:
Blocks #351

lorenzwalthert:
Solved in #353 but this is blocked by #352.

lorenzwalthert:
Solved in #353 but this is blocked by #351.

lorenzwalthert:
Closed with #353.

Issue #351, opened 2018-02-20
Inconsistent indention for EQ_SUB


martin-mfg:
The following code was styled using the latest development version of styler:

```R
a_very_long_function_name <- function(parameter_name =
                                      "a very long default value") {
  NULL
}

a_very_long_function_name(
  parameter_name =
    "a very long default value"
)
```

What confuses me here is that in the function call, `"a very long default value"` is indented by 1 more level than `parameter_name =`, but in the function definition this is not the case.
I think there is no rule in the tidyverse style guide that governs this behaviour, so in my opinion it would be reasonable to apply the same indention strategy in both cases.

lorenzwalthert:
Thanks @martin-mfg, this is indeed unexpected. I think I know where it stems from and I will look into it as soon as possible.

lorenzwalthert:
Blocked by #352.

lorenzwalthert:
Blocked by #357.

lorenzwalthert:
Solved in #353 but this is blocked by #351.

lorenzwalthert:
Closed with #353.

PR #353, opened 2018-02-23
Token-dependent indention in combination with standard indention

lorenzwalthert:
# Overview
Token-dependent indention plays a minor role in styler, since the tidyvese style guide does only permit it in one situation: Function declarations.
As of now, token-dependent indention was always absolute, i.e. no other indention rules were allowed. This lead to problems such as the one outlined in #351. After having tried to solve the problem by modifying styler infrastructure (which was a hack, and committed in this branch for reference), the ideas developed in the second part of this PR (that reversed some hacks from the first part) should provide a more refined and flexible solution via a transformer function and allowing other style guides in the future to use the feature of token dependent indention in other ways.

# Example
Let's do it with an example. Suppose we have
```r
function(a = 
b, d +
e, 
c) {
  call(1, 
    2
)
}
```
Which is incorrectly indented.

Apply indention rules for braces (both curly and round). Function declarations will get indented because of the opening brace after `function` (no other indention is possible because other braces won't live in the same nest.
```r
function(a = 
  b, d,
  e, 
  c) {
  call(1, 
    2
)
}
```

Remove all indention in function declaration headers. Note that function calls are not affected. After this step, code may look like this:
```r
function(a = 
b, d,
e, 
c) {
  call(1, 
    2
)
}
```

Apply other indention rules such as indention based on operators (`+`, `=` etc.). Again, note that function calls are not affected.
```r
function(a = 
    b, d,
  e,
c) {
  call(1, 
    2
)
}
```
Finally, we add the indention of the reference token `(` to all tokens in the header, so we end up with the desired indention.
```r
function(a = 
            b, d,
         e,
         c) {
  call(1, 
    2
)
}
```

Closes #353, #352.


tests/testthat/fun_dec/line_break_fun_dec-out.R
```diff
@@ -12,6 +12,6 @@ a <- function(x, #
 
 a <- function(x, #
               y #
-            ) { # FIXME: Move to the same indention level as a
+) {
```
krlmlr: Do we need to indent this closing paren?

tests/testthat/indention_operators/eq_formals_complex_indention-out.R
```diff
@@ -0,0 +1,33 @@
...
+         b) {}
+
+function(a,
+         b,
+         c
+) {}
+
+function(a,
+         b,
+         c) {}
```
krlmlr: Add line break if formals consume multiple lines?

tests/testthat/indention_operators/eq_formals_complex_tokens-out.R
```diff
@@ -0,0 +1,30 @@
...
+function(ss,
+         a =
+           3,
+         er =
+           4) {}
+
+function(a =
+           b,
+         f =
+           d, c =
```
krlmlr: Add newline after `,` if the argument consumes multiple lines?

tests/testthat/indention_operators/eq_sub_complex_indention-out.R
```diff
@@ -0,0 +1,8 @@
+call(a =
```
krlmlr: Add a line break if argument is split over multiple lines?

R/indent.R
```diff
@@ -144,28 +163,55 @@ compute_indent_indices <- function(pd,
...
+  multi_line_token <- pd_is_multi_line(
+    pd[row_idx_between_trigger_and_line_break, ]
+  )
+  remaining_row_idx_between_trigger_and_line_break <- setdiff(
+    row_idx_between_trigger_and_line_break,
+    potential_trigger_pos
+  )
+
+  other_trigger_on_same_line <-
+    pd[remaining_row_idx_between_trigger_and_line_break, ]$token %in%
```
krlmlr: `pd$token[...]` is slightly faster.

tests/testthat/fun_dec/line_break_fun_dec-out.R
lorenzwalthert: I think it makes sense to not indent it (this is also what happens in RStudio) because a closing brace unindents itself. Alternative to leave it as it was before is moving it to the same level as `)`, which I think looks worse plus it's more complicated 😩.

tests/testthat/indention_operators/eq_formals_complex_indention-out.R
lorenzwalthert: Line break for the closing brace? I think under [long lines](http://style.tidyverse.org/functions.html#long-lines-1), there is no line break for that case (just an example, not covered in the floating text).
Should we open an issue to make the tidyverse style guide explicit in this regard?

lorenzwalthert:
Thanks @krlmlr. I see your point but I am not sure if I can easily split it up, because some changes I made affect multiple situations. Hence, I would feel more comfortable to keep it as one PR. Also, I wonder whether the tidyverse style guide needs to be extended in order to cover some of the cases we handle here too.

tests/testthat/indention_operators/eq_formals_complex_indention-out.R
krlmlr: Never mind, it's not that important.

tests/testthat/indention_operators/eq_formals_complex_tokens-out.R
krlmlr: Multiline formals aren't covered either, because they're very rare. I think if a formal is spread over multiple lines, no other information should be contained on these lines.

tests/testthat/indention_operators/eq_formals_complex_indention-out.R
lorenzwalthert: Ok, let's leave that one for now

tests/testthat/indention_operators/eq_formals_complex_tokens-out.R
lorenzwalthert: Deal 😊

lorenzwalthert: To do this properly, we actually need #357 because the `multi-line` attribute is not set just looking for line breaks at the current nest is insufficient.

lorenzwalthert: As this is part of milestone styler 1.0.1 (which was due a few days ago), I don't think I can implement #357 reasonably quickly. Should we do it improperly and add a FIXME or leave it as is and add a FIXME?

tests/testthat/fun_dec/line_break_fun_dec-out.R
lorenzwalthert: Also, this is only relevant for cases where the closing brace cannot go directly after the last formal, because it is blocked by a comment. Otherwise the brace moves up. Debatable whether we want that or not, but I think it's a detail.

tests/testthat/indention_operators/eq_formals_complex_tokens-out.R
lorenzwalthert: Can we migrate this to a new issue? Because a merge of #353 would close #352 and #351 and all the other issues raised in #353 seem to be gone or declared as negligible.

krlmlr:
Agree to merge and migrate remaining problems to a new issue. The new output looks better than what we currently have.

Issue #357, opened 2018-02-24
Allowing line break transformers to depend on multi-line attribute


lorenzwalthert:
Line break rules may depend on the structure of the parse table. So we may want to keep curly-brace-expressions as one-liner if the input was one line:
```r
if (x) { x + 1 }
```
Note that this is not easily possible. The `multi-line` attribute is not up-to-date for the line break transformers, hence we cannot rely on it. Further, we cannot rely on `line1` and `line2`, because this attribute was removed for safety beforehand (as it's also not up to date). We can try to "pre-reconstruct" the `multi-line` attribute by looking into the children, but that's brittle. One way out would be to split the line break transformers into two groups. One not depending on an up-to-date `multi-line` attribute and one dependent on it. 
We could first apply the first transformer group, then update the `multi-line` attribute and then apply all transformers, updating the `multi-line` attribute between every transformer (which seems expensive).

lorenzwalthert:
Reference: #342.

lorenzwalthert:
This requires style guides to have a new structure: 
```r
create_style_guide(
    # transformer functions
    initialize                      = default_style_guide_attributes,
    line_break                      = line_break_manipulators,
    line_break_requiring_multi_line = new_transformers,  # new category
    space                           = space_manipulators,
    token                           = token_manipulators,
    indention                       = indention_modifier,
    # transformer options
    use_raw_indention               = use_raw_indention,
    reindention                     = reindention
  )
```

lorenzwalthert:
@krlmlr what do you think? This is needed to solve https://github.com/r-lib/styler/pull/353#discussion_r170291414 and #342 properly, which I wanted to be part of milestone styler 1.0.1 but it seems time consuming. I think especially #342 should be solved as part of styler 0.1.0, especially the case we can solve with an improper approach.

krlmlr:
I don't mind postponing #353, let's add a test for the behavior in #342 instead ;-)

lorenzwalthert:
Blocks #351.

lorenzwalthert:
Blocks #342.

lorenzwalthert:
Blocks #371.

lorenzwalthert:
Blocks #363.

krlmlr:
Does the increasing number of "blocks" increase the priority of this issue?

lorenzwalthert:
Not if I review of the issues yields that half of them are not blocked anymore 😊

lorenzwalthert:
[...]

PR #364, opened 2018-03-02
Named arguments and line breaks in function calls

### LINKED ISSUE #342: Line broken again after first named argument
The below code should not be modified by styler.
``` r
styler::style_text(
"call(
  a, b = 3
)")
#> call(
#>   a,
#>   b = 3
#> )
```

Created on 2018-02-05 by the [reprex package](http://reprex.tidyverse.org) (v0.1.1.9000).

It seems as this undesired behaviour was introduced with #318.

lorenzwalthert:
Closes #342 with a pragmatic solution, not handling multi-line expressions as suggested in #357 for now (but added a `FIXME`), which is probably very rare and it's not even clear how such a case should be styled.


lorenzwalthert:
Abandoned since #342 is closed.

PR #366, opened 2018-03-03
Prepare v1.0.1

lorenzwalthert:
This prepares for the release of `v1.0.1`. @krlmlr Could you quickly have a look at `NEWS.md`? The rest should be fine I think. Can we use markdown syntax like [show](source) for web-links etc? Also, probably we can merge master into production then.


NEWS.md
```diff
@@ -1,3 +1,33 @@
+## styler 1.0.1 (2018-03-02)
+
+This is a maintenance release without any breaking API changes.
+
+## Major & dependency related changes
+
+* Removed implicit `dplyr` dependency via `purrr:::map_dfr()` (thanks 
+  @jimhester, #324).
+* Added required minimal version dependency for purr (`>= 0.2.3`) (#338).
+* The dependency tibble was optimized for speed in `v1.4.2` so styler should run
```
krlmlr: We rely on the tibble package which ...

NEWS.md
```diff
@@ -1,3 +1,33 @@
...
+## Major & dependency related changes
+
+* Removed implicit `dplyr` dependency via `purrr:::map_dfr()` (thanks 
+  @jimhester, #324).
+* Added required minimal version dependency for purr (`>= 0.2.3`) (#338).
+* The dependency tibble was optimized for speed in `v1.4.2` so styler should run 
+  ~2x as fast [(#348)](https://github.com/tidyverse/tibble/pull/348). Hence, 
+  styler now depends on `tibble >= 1.4.2`.
+* In the dependency `enc`, a bug was fixed that removed/changed non-ASCII 
+  characters. Hence, styler now depends on `enc >= 0.1-10` (#348).
```
krlmlr: 0.2 has been built on all platforms.

NEWS.md
```diff
@@ -1,3 +1,33 @@
...
+
+## Minor changes
+
+* Recognizing and respecting of DSLs used in R comments: rplumnber (`#*`, #306),
```
krlmlr: "respecting of" doesn't sound right

R/parse.R
```diff
@@ -9,11 +9,12 @@
 #'   * A column "pos_id" for (positional id) which can be used for sorting
 #'     (because "id" cannot be used in general). Note that the nth value of this
 #'     column corresponds to n as long as no tokens are inserted.
-#'   * A column "child" that contains the nested subtibbles.
+#'   * A column "child" that contains the nested sub-tibbles.
```
krlmlr: Do we want to use the "nest" terminology here (maybe later)?

NEWS.md
lorenzwalthert: Really? I thought that is proper usage. Searching dplyr issues and PRs gives over 30 results that seem to use "respect" in the same way: https://github.com/tidyverse/dplyr/issues?utf8=✓&q=respect

NEWS.md
lorenzwalthert: So should we rather require `>=0.2`?

NEWS.md
lorenzwalthert: Ok.

R/parse.R
lorenzwalthert: Good idea.

NEWS.md
krlmlr: Maybe "We're now recognizing and respecting DSLs used in R comments. ...

NEWS.md
krlmlr: I think so.

Issue #332, opened 2018-01-26
Style for examples in roxygen or .Rd files


nfultz:
It would be cool if we could also restyle the examples in our R package, since that's code that end users might actually see.

lorenzwalthert:
Thanks @nfultz, sounds like a cool idea but probably that's not straight forward if it should be done in a safe and clean way. We'd probably have to 

* parse roxygen comments somehow from all `R/` files (maybe this is similar to #13 in some sense) by identifying the `@example(s)` header / tag and the corresponding content.
* removing the roxygen comment prefixes (`#'`) for these lines.
* styling the content.
* re-adding royxgen comment prefixes and header.
* inserting the styled expression.

Indeed this seems really similar to #13 and I think for consistency, the infrastructure introduced in #233 should be re-uesd / extended as much as possible so that we can avoid code duplication.

Any interest to get us started with a PR? If yes, please make sure to follow the contribution guide lines outlined [here](https://github.com/r-lib/styler/blob/master/CONTRIBUTING.md). I am happy to review. Otherwise, this will be postponed at least a few months due to time constraints.

lorenzwalthert:
Anyways I think we should limit our attention to roxygen source code in a first step if that's ok for you since the majority of package developers uses roxygen.

nfultz:
Another idea - there is all ready existing tools to go from `roxygen => Rd` (`roxygen2` package) and from `Rd examples => R` (`R CMD check` does so, for example `package-Ex.R`). Yihui has written a converter for `Rd => roxygen` https://yihui.name/rd2roxygen/

So the missing link could be from an examples .R => Rd, which would give a round trip of:

ugly roxygen  => ugly Rd => ugly package-Ex.R => styler package-Ex.R => pretty Rd => pretty roxygen

styler would just need to be "compatible" with the tool chain and maintain the example metadata already present in `package-Ex.R`.

You might imagine a similar flow for R markdown, by going through `knitr::spin` and `knitr::purl` to go to/from regular R scripts.

lorenzwalthert:
Ok. I think it's cleaner if we can style the source code directly instead of a derivative of it. Also, the mapping from pretty .Rd back to the source file line where the example stems from seems to make things complicated IMHO. What do you think? Did you take a look at #233? Instead of separating out code / non code from an Rmd and styling this, we want to separate `@example(s)` roxygen vs non-examples roxygen so I think there is quite some overlap.

lorenzwalthert:
[...]

PR #381, opened 2018-03-24
Style roxygen code examples

lorenzwalthert:
* This is WIP with high-level template functions and a suggestion for the hierarchical structure. 
* I think having a text and path interface for identify start and stop is convenient (`identify_start_and_stop_of_royxgen_examples_from_text()`), at least as long as we are in the experimental state. 
* I also added a test structure for the identification part and *numbered* test files, which is a friction to other tests in the repo but I think it can help to quickly identify a failing combinations of the elements we want to test. We can possibly remove the numbers eventually.
* I used long function names on purpose to be very verbose.

To do:
- [x] completing function templates.
- [x] more tests for identification of start / stop plus unit test for other functionality this PR provides.
- [x] Documentation.
- [x] namespace handling.
- [x] consider relocating functions in R/utils.R to separate files (like `string-manipulation.R`, `string-creation.R`)


R/style-roxgen-examples.R
```diff
@@ -0,0 +1,68 @@
...
+  ) %>%
+    start_stop_pairs_to_sequences()
+
+  masked_examples <- extract_selected_lines_from_text(
+    full_file_content, start_stop_sequences
+  )
+
+  plain_examples <- map(start_stop_sequences, remove_roxygen_mask,
+    text = masked_examples
+  )
```
lorenzwalthert: Oups this seems to need mapping over both sequences anybody masked examples (which is a list with each element corresponding to one example).

lorenzwalthert:
As noted in #332, we should aim for consistency with styling of Rmd files. In particular, styling has to happen inside `transform_code` because all other approaches fail to identify (and communicate to the end user) in absence of a change in the code. 
We could just call `style_text()` on the identified code within `transform_code`, which would have the advantage that we could style roxygen comments in roxygen comments recursively. This implies passing a lot of arguments downstream. Also, creating such a recursive `style_text()` call is not in the spirit of styler (so far) and hence we may want to try to do it differently.

Probably a better and more consistent approach would be to see code examples as part of the code body that is styled with transformer functions. That would imply to 

* create a new column `is_roxygen_code_example` prior to parsing that identifies roxygen code examples. 
* remove the roxygen mask including the header for all roxygen code example tokens.
* compute nested parse table.
* style all code as implemented currently (so roxygen code examples are treated as normal code).
* adding roxygen mask, basically inverting its removal.

Probably that's too complicated if not infeasible.

In addition, we should use consistent naming for file names. R/transform-code.R should have rmd in the name since it mostly deals with rmd. The first function `transform_code()` should be renamed since it is essentially `transform_file()` and `transform_file()` should be `transform_file_verbose()`.  The new `transform_file()` should be moved to R/transform-files.R.

Since this will lead to an API change, we better think twice how we name the argument and what its values can be. In consideration of extending style functionality for roxygen comments, e.g. by maximal line width, indention etc. as suggested in https://github.com/r-lib/styler/issues/332#issuecomment-375663300, we may want to call the argument `include_roxygen` to the tidyverse style guide, which can be `NULL` or a vector with arguments for elements of roxygen comments to style.

lorenzwalthert:
Further todo: 

* Handle don't run.
* Make styling of roxygen an option to top-level API.

R/transform-files.R
```diff
@@ -81,18 +81,39 @@ transform_file <- function(path,
...
+    transformed_code <- text %>%
+      parse_transform_serialize_r(transformers) %>%
+      parse_transform_serialize_roxygen(transformers)
+    transformed_code
   }
 }
 
+#' @importFrom purrr map_at flatten_chr
+parse_transform_serialize_roxygen <- function(text, transformers) {
+  roxygen_seqs <- identify_start_to_stop_of_roxygen_examples_from_text(text)
```
lorenzwalthert: Probably need to split that up into multiple functions. Also, some of it might be re-used for dontrun handling.

lorenzwalthert:
@krlmlr any idea how to handle `\dontrun{}` segments? I am playing around with one approach, but it seems rather brittle and inelegant. Maybe you have any knowledge from roxygen2 that could help us here?

krlmlr:
Maybe with `tools::parse_Rd()`?

``` r
file <- tempfile()
writeLines(
  c(
    "\\examples{",
    "\\dontrun{",
    "TRUE",
    "}",
    "}"
  ),
  file)
parsed <- tools::parse_Rd(
  file
)
str(parsed)
#> List of 2
#>  $ :List of 3
#>   ..$ : atomic [1:1] 
#> 
#>   .. ..- attr(*, "Rd_tag")= chr "RCODE"
#>   .. ..- attr(*, "srcref")=Class 'srcref'  atomic [1:6] 1 11 1 11 11 11
#>   .. .. .. ..- attr(*, "srcfile")=Class 'srcfile' <environment: 0x55eafd898290> 
#>   ..$ :List of 2
#>   .. ..$ : atomic [1:1] 
#> 
#>   .. .. ..- attr(*, "Rd_tag")= chr "VERB"
#>   .. .. ..- attr(*, "srcref")=Class 'srcref'  atomic [1:6] 2 10 2 10 10 10
#>   .. .. .. .. ..- attr(*, "srcfile")=Class 'srcfile' <environment: 0x55eafd898290> 
#>   .. ..$ : atomic [1:1] TRUE
#> 
#>   .. .. ..- attr(*, "Rd_tag")= chr "VERB"
#>   .. .. ..- attr(*, "srcref")=Class 'srcref'  atomic [1:6] 3 1 3 5 1 5
#>   .. .. .. .. ..- attr(*, "srcfile")=Class 'srcfile' <environment: 0x55eafd898290> 
#>   .. ..- attr(*, "Rd_tag")= chr "\\dontrun"
#>   .. ..- attr(*, "srcref")=Class 'srcref'  atomic [1:6] 2 1 4 1 1 1
#>   .. .. .. ..- attr(*, "srcfile")=Class 'srcfile' <environment: 0x55eafd898290> 
#>   ..$ : atomic [1:1] 
#> 
#>   .. ..- attr(*, "Rd_tag")= chr "RCODE"
#>   .. ..- attr(*, "srcref")=Class 'srcref'  atomic [1:6] 4 2 4 2 2 2
#>   .. .. .. ..- attr(*, "srcfile")=Class 'srcfile' <environment: 0x55eafd898290> 
#>   ..- attr(*, "Rd_tag")= chr "\\examples"
#>   ..- attr(*, "srcref")=Class 'srcref'  atomic [1:6] 1 1 5 1 1 1
#>   .. .. ..- attr(*, "srcfile")=Class 'srcfile' <environment: 0x55eafd898290> 
#>  $ : atomic [1:1] 
#> 
#>   ..- attr(*, "Rd_tag")= chr "TEXT"
#>   ..- attr(*, "srcref")=Class 'srcref'  atomic [1:6] 5 2 5 2 2 2
#>   .. .. ..- attr(*, "srcfile")=Class 'srcfile' <environment: 0x55eafd898290> 
#>  - attr(*, "class")= chr "Rd"
#>  - attr(*, "srcref")=Class 'srcref'  atomic [1:6] 0 0 6 1 0 1
#>   .. ..- attr(*, "srcfile")=Class 'srcfile' <environment: 0x55eafd898290> 
#>  - attr(*, "macros")=<environment: 0x55eafd71e0b0>
```

> Created on 2018-05-02 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

lorenzwalthert:
Well I am not sure if we want to replace source code with styled derivatives that are potentially out of date.

krlmlr:
`\dontrun{}` doesn't always mean outdated code. It may be an example that won't run on CRAN, or just something that's valid syntax but doesn't work. Either way, after `parse_Rd()` you should be able to leave out these parts.

lorenzwalthert:
Yes sure, not always, but it can be outdated. Let's say you update a code example in your roxygen comments and then run styler without doing devtools::document(). This will replace the source with its styled derivative and you will loose your changes. So I wondered if there is a way around taking code directly from roxygen comments, which we can do if it does not involve a dontrun. 
We could also look into taking all example code from the .Rd and require up to date documentation prior to styling when roxygen code examples are to be styled. This would mean to start all over since we now worked in a solution based in the source, that is, taking code from the roxygen comments and not from the .Rd files

krlmlr:
How about

1. look for `@examples` section in *roxygen2* comments
2. remove the `'#`
3. style
4. put back the `'#`
5. replace the `@examples` section with the styled version

?

lorenzwalthert:
This is what we have but that does not allow `\dontrun{}` segments.

krlmlr:
2a) Run `parseRd()`.

3) Style code in all elements of the returned structure.

3a) Serialize the parsed structure.

lorenzwalthert:
So I am not sure if I understand you correctly, you want the following:

1. look for `@examples` section in *roxygen2* comments
2. remove the `'#`.
3. Parse blocks.
a. wrap the extracted code in `\examples{...}`. 
b. Call ` as.character(tools::parse_Rd(...)` to get a vector that contains the individual elements of the example in useful blocks.
4. Process the blocks 
5. wrap the extracted code in `\examples{...}`. 
6. put back the `'#`
7. replace the `@examples` section with the styled version

Here an illustration of step 3 to 5:
``` r
library(magrittr)
file <- tempfile()
writeLines(
  c(
    "\\examples{",
    "hi <- 3",
    "\\dontrun{",
    "TRUE", "function(x) {", "  x", "}",
    "}",
    "x",
    "}"
  ),
  file
)
parsed <- tools::parse_Rd(
  file,
  fragment = TRUE
)
#> Warning in tools::parse_Rd(file, fragment = TRUE): /var/folders/8l/
#> fhmv3yj12kncddcjqwc19tkr0000gr/T//Rtmp6bX2B6/file20645f19d510:1: unexpected
#> section header '\examples'
char <- as.character(parsed, deparse = TRUE)
char
#>  [1] ""                "{"               "\n"             
#>  [4] "hi <- 3\n"       "\\dontrun"       "{"              
#>  [7] "\n"              "TRUE\n"          "function(x) {\n"
#> [10] "  x\n"           "}\n"             "}"              
#> [13] "\n"              "x\n"             "}"              
#> [16] "\n"


# styline the blocks in palce.
# Code here

paste0("\\examples", paste0(char, collapse = "")) %>% cat()
#> \examples{
#> hi <- 3
#> \dontrun{
#> TRUE
#> function(x) {
#>   x
#> }
#> }
#> x
#> }
```

Created on 2018-05-21 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

Problem: `tools::parse_Rd()` only takes a file as input, which means we need to write a temp file if we want to use the always-up-to-date source code instead of the `.Rd` derivative. Rather not elegant. What do you think? We could also give warnings and take derivatives if they are newer than the source file but then we need a mapping from `.Rd` to `.R`.

lorenzwalthert:
Oh we could just use (textConnection)[https://stackoverflow.com/questions/14979858/how-to-create-text-connection-from-string-variable] to feed in a string.

krlmlr:
Yes, it's a bit of an effort but I think it's worth it to have nicely styled examples.

lorenzwalthert:
Ok, I agree. I think me and @jonmcalder have already put in some time, so I think it's worth going forward this way. I think we need a lot of TDD to accommodate all edge cases.

lorenzwalthert:
We can also set up a test infrastructure on Travis to download and style random CRAN packages for testing.

lorenzwalthert:
It seems as it's done :-) Anyone interested in testing? I tested on styler and ggplot2 source code and I did not get any unexpected problems.

krlmlr:
Thanks! I'll take a look at this PR. Could you please merge/rebase so that I can test against master + these changes?

lorenzwalthert:
Thanks for the offer @krmlr. Yes, I'll do that. However, I think I want to add some unit tests for the unexported to increase coverage and review myself first. Also, I can describe on a high-level what theses functions do (either in the PR or in the source code as part of the documentation) to facilitate the review process. In addition, I am sure with a fresh view on the PR I am likely to find some stuff I can improve on before you review. I'll give you another ping once it's ready ok?

R/testing.R
```diff
@@ -68,7 +68,7 @@ test_collection <- function(test, sub_test = NULL,
 #'  "path/to/file/first-extended-in.R"))
 #' @keywords internal
 construct_out <- function(in_paths) {
-  gsub("\\-.*([.]R(?:|md))$", "\\-out\\1", in_paths)
+  gsub("\\-in([.]R(?:|md))$", "\\-out\\1", in_paths)
 }
```
lorenzwalthert: Need to adapt documentation for `test_collection()`.

lorenzwalthert: done

lorenzwalthert:
Ready for review @krlmlr. Documented broad concept in the hierarchy section of the function `parse_transform_serialize_roxygen()`, it's inherited by the `style_roxygen()_*` functions. 

There is one problem with the current solution though. Line breaks are not respected. And I think we should get that right before we merge. Key is that `tools::parseRd()` does not return a vector with one element representing one line, which is the definition otherwise used in styler. I guess we could fix it somehow, I just thought maybe you want to have a look first and maybe you see a solution while reviewing. The thing is that we use the property that `tools::parseRd()` does, for example, return the `}` of `\dontrun{` as an individual element. That is the whole point of using it. Maybe we just need to extend the post-processing somehow.

R/roxygen-examples-add-remove.R
```diff
@@ -0,0 +1,32 @@
...
+  mask <- c(
+    1L, 2L, length(roxygen), potential_pos[is_line_break_at_potential_pos]
+  ) %>% sort()
+  list(
+    code = roxygen[-mask], mask = paste(roxygen[seq2(1, 2)], collapse = "")
+  )
+}
+
+remove_blank_lines <- function(code) {
+  code[code != "\n"]
```
krlmlr: Are the terminal newlines part of the contents of a character vector? In `readLines()` (and many other functions), each vector element is a line, but the newlines aren't part of it. This design is surprising at least.

R/utils.R
```diff
@@ -13,11 +13,19 @@ two_cols_match <- function(col1, col2, data) {
 }
 
 odd <- function(x) {
-  x[seq(1L, length(x), by = 2)]
+  x[odd_index(x)]
+}
+
+odd_index <- function(x) {
+  seq(1L, length(x), by = 2)
```
krlmlr: This will throw an error if `length(x) == 0`.

R/utils.R
```diff
@@ -13,11 +13,19 @@ two_cols_match <- function(col1, col2, data) {
...
 }
 
 even <- function(x) {
-  x[seq(2L, length(x), by = 2)]
+  x[even_index(x)]
+}
+
+even_index <- function(x) {
+  seq(2L, length(x), by = 2)
```
krlmlr: This will throw an error if `length(x) < 2`.

krlmlr:
I don't understand "line breaks are not respected". Would `strsplit(., "\n", fixed = TRUE)[[1]]` help? Or is this about line breaks near the braces in `\dontrun{}` et al.?

R/roxygen-examples-add-remove.R
krlmlr: Does this help get rid of trailing newlines at a very early stage?

``` r
library(magrittr)
tools::parse_Rd(textConnection("\\dontrun{x <- 3\n\n # hi}"), fragment = TRUE)[[1]] %>% as.character()
#> [1] "x <- 3\n" "\n"       " # hi"
```

> Created on 2018-07-15 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

lorenzwalthert: Not sure because we must keep the `dontrun`. We just want the following mapping: 
``` r
library(magrittr)
# 1) parse
rd <- tools::parse_Rd(textConnection(
  "\\dontrun{x <- 3\n\n # hi}"
), fragment = TRUE) %>%
  as.character()
rd
#> [1] "\\dontrun" "{"         "x <- 3\n"  "\n"        " # hi"     "}"        
#> [7] "\n"

# 2) map to styler line definition
mapping_to_be_defined(rd)
# > [1] "\\dontrun{x <- 3" "" "# hi}"
```
Also, the problem I think is that `strsplit` does not split if the character is already the last in the string.
``` r
library(magrittr)
c("a\n", "x") %>% strsplit("\n", fixed = TRUE) %>% unlist()
#> [1] "a" "x"
# should be
# > [1] "a" "" "x
```

Created on 2018-07-15 by the [reprex
package](http://reprex.tidyverse.org) (v0.2.0).

lorenzwalthert: Not to confuse anywone, just a note for myself. A problem might be that `strsplit` does not split if the character is already the last in the string.
``` r
library(magrittr)
c("a\n", "x") %>% strsplit("\n", fixed = TRUE) %>% unlist()
#> [1] "a" "x"
# should be
# > [1] "a" "" "x
```

Created on 2018-07-15 by the [reprex
package](http://reprex.tidyverse.org) (v0.2.0).

lorenzwalthert:
I think it's good practice to leave a PR for a few days and go back afterwards. I think I got the big picture right, but there are some small things that can simplify the code (and add the feature *preserving line-breaks*).

R/utils.R
krlmlr: I think this function should just handle that case.

lorenzwalthert:
Will do a PR on top of this PR to solve various problems. -.-

lorenzwalthert:
Finally, it seems we got it right 🎉

lorenzwalthert:
We just added ~3000 lines of code to styler. Most of it is tests though, but still a few hundred in the `R`directory.


```r
library(dplyr)
log <- gitsum::parse_log_detailed_full_run()
log %>%
  arrange(desc(date)) %>%
  slice(1:63) %>%
  unnest(nested) %>%
  filter(dirname(changed_file) == "R") %>%
  summarize(sum(insertions), sum(deletions))

#> # A tibble: 1 x 2
#>  `sum(insertions)` `sum(deletions)`
#>              <int>            <int>
#> 1               488              354
```

Issue #216, opened 2017-09-21
Long character string gets truncated


lorenzwalthert:
Again, the problem seems to be the parser:
```r
txt <- paste0(
    "'", 
    paste0(sample(c(letters, rep(" ", 10)), 2000, replace = TRUE), collapse = ""), 
    "'"
  )
  getParseData(parse(text = txt))

#>  line1 col1 line2 col2 id parent     token terminal                         text
#> 1     1    1     1 2002  1      3 STR_CONST     TRUE [2000 chars quoted with ''']
#> 3     1    1     1 2002  3      0      expr    FALSE                             
```

The problem seems to be introduced with `utils::getParseData()`
```r
parse(text = txt)
[...]

PR #384, opened 2018-03-26
Only replace long strings

lorenzwalthert:
Another attempt to solve #216, by implementing suggestion 2 suggested in https://github.com/r-lib/styler/issues/216#issuecomment-373049743. That is, only replace the strings that are necessary, not all strings if there is one string in the parse data that is longer than 1'000. This does (not yet) implement line / col substitution but returns an error if there is a problem, so we are failing fast instead of silently. I think it's rather unlikely that this is going to cause a problem in the future.


lorenzwalthert:
@krlmlr there seem to be a bug in the R base parser of `R3.1` with this example. It simply does not parse all tokens. I managed to reproduce it after painfully re-install R 3.1 and all packages needed on my Mac (this took forever, I also tried on Ubuntu and docker images). Can we skip a test based on R version? In principle, we should check whether the R version is below 3.2 and if so, check whether any non-terminal has no children and if so, throw an error.

lorenzwalthert:
> In principle, we should check whether the R version is below 3.2 and if so, check whether any non-terminal has no children and if so, throw an error.

This is implemented with 7cbc6fb

PR #389, opened 2018-04-24
Separate from potential "Warning message"

katrinleinweber:
Hello :-)

I recently saw grafik and would thus like to suggest this minimum viable formatting improvement.

### DIFF
--- a/R/communicate.R
+++ b/R/communicate.R
@@ -8,7 +8,7 @@
 #' @keywords internal
 communicate_warning <- function(changed, transformers) {
   if (any(changed, na.rm = TRUE) && !can_verify_roundtrip(transformers)) {
-    cat("Please review the changes carefully!")
+    cat("Please review the changes carefully! ")
   }
 }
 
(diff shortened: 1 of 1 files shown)


lorenzwalthert:
Thanks Kathrin. Is the warning you got related to styler? Also, we could maybe even append a line break instead of a space, i.e. setting `fill = TRUE` and leaving the message string untouched. What do you think?

katrinleinweber:
I personally don't like the look of

grafik

but what you suggest is surely better code style ;-)

lorenzwalthert:
Ok ;-). No idea why the CI build fails. Probably unrelated to your changes though...

lorenzwalthert:
Unfortunately I don't have the access rights to re-trigger the build, but I think the problem might have vanished since there were no problems in #391. Can you retrigger the build by an empty ammendment of the last commit (i.e. with `git commit --amend`)? Maybe there is also a GitHub GUI way to do that.

lorenzwalthert:
Thanks @katrinleinweber. Will merge as soon as current master build is fixed. Some issues with `pkgload` being missing, most likely because we now use the `pkgdown` version from CRAN and the package is not in the corresponding `Remotes:` entry.

lorenzwalthert:
Thanks 😊

katrinleinweber:
You're welcome, and thanks for merging :-)

PR #399, opened 2018-06-18
Remove viridis helper

### LINKED ISSUE #393: Removing viridis helper
Got a notification form CRAN maintainer because we have an undeclared import in testthat:
https://github.com/r-lib/styler/blob/dd62ba7575eceef071aece62a7ab9fd461694ced/tests/testthat/helper-viridis.R#L1

Can't remember why we put that there, I think some namespace issue which might have been limited to your local machine only, or because we used r devel (or some devel packages) at some point. Can I safely delete this @krlmlr?

lorenzwalthert:
Closes #393.

### DIFF
--- a/tests/testthat/helper-viridis.R
+++ b/tests/testthat/helper-viridis.R
@@ -1,1 +1,0 @@
-suppressWarnings(requireNamespace("DiagrammeR"))


krlmlr:
I don't remember what this was for, fine with removing.

PR #407, opened 2018-07-08
Fix CI for R < 3.4

lorenzwalthert:


lorenzwalthert:
@krlmlr have you fixed anything in tic? It works now, I have fixed some test that strangely did not work in R 3.1 (only error message display). The vignette engine error was gone.

Issue #431, opened 2018-10-11
Support for .Rnw files


Hasnep:
Sorry if this is a dumb thing to ask, but are there any plans to add support for .Rnw files?

If not, how difficult would it be for me to add this as someone who is unfamiliar with developing R libraries?

lorenzwalthert:
> Sorry if this is a dumb thing to ask, but are there any plans to add support for .Rnw files?

It's not 😄. I indeed thought about it some time ago, but since I am not using `.Rnw` myself, I decided not to look into it. To be honest I am not sure how easy that is for you. @jonmcalder and I implemented that feature in #233. I think it should not be too hard to adapt the package to handle `.Rnw` too since from what I can tell, it looks as if we mainly have to change the regex that is dedicated to match `{r ` now. But if you are not familiar with R package development workflows, even a small change in the  source code can be tricky. @jonmcalder would you be interested in extending styler in that direction?

jonmcalder:
Sure - this seems feasible and I'd be happy to assist with this. @Hasnep if you'd like to work on this together let me know and we can chat further?

lorenzwalthert:
Cool Jon. Thanks for taking care of this. Looking forward to review a PR 😊

Hasnep:
Thank you both for your quick responses. I don't think my R skills are at the level where I can usefully contribute yet.

lorenzwalthert:
@Hasnep interested in testing this feature? Advised: save + commit / backup a copy of the files you apply styler first.

```r
remotes::install_github("r-lib/styler#434")
```

Then you can proceed as usual after restarting R (Cmd + Shift + F10 in RStudio).

lorenzwalthert:
Closed with #434.

PR #434, opened 2018-10-16
Add support for .Rnw files

jonmcalder:
To address #431


lorenzwalthert:
Good start :-)

R/transform-code.R
```diff
@@ -15,40 +15,78 @@ transform_code <- function(path, fun, verbose = FALSE, ...) {
...
     flatten_chr()
 }
 
+#' Transform Rmd contents
+#'
+#' Applies the supplied transformer function to code chunks identified within
+#' an Rmd file and recombines the resulting (styled) code chunks with the text
+#' chunks.
+#'
+#' @param lines A character vector of lines from an Rmd file
```
lorenzwalthert: tidyverser style requires full stops.

R/transform-code.R
```diff
@@ -60,16 +98,16 @@ separate_chunks <- function(lines) {
...
-#' Identifies raw R code chunks
+#' Identifies raw Rmd code chunks
 #'
 #' Raw in the sense that these chunks don't contain pure R code, but they
 #' contain a header and footer of markdown. Only code chunks that have an engine
 #' whose name matches `engine-pattern` are considered as R code.
 #' @inheritParams separate_chunks
 #' @param engine_pattern A regular expression that must match the engine name.
 #' @keywords internal
-identify_r_raw_chunks <- function(lines, engine_pattern = get_engine_pattern()) {
```
lorenzwalthert: For some reason we did not name them *rmd chunks*, I think because of ambiguity with other functions, but I can't exactly remember why and looking at all function names that match `rmd`
```r
# remotes::install_github("lorenzwalther/parsesum")
parsesum::ls_fun_decs() %>% dplyr::filter(grepl("rmd", .$text))
#> # A tibble: 2 x 4
#>   text          n_lines source               line1
#>   <chr>           <int> <chr>                <int>
#> 1 transform_rmd       7 ./R/transform-code.R    33
#> 2 is_rmd_file         3 ./R/utils.R             83
```
Or `identify`: 
```r
parsesum::ls_fun_decs() %>% dplyr::filter(grepl("identify", .$text))
#> # A tibble: 4 x 4
#>   text                                                 n_lines source                      line1
#>   <chr>                                                  <int> <chr>                       <int>
#> 1 identify_insufficiently_parsed_stings                      8 ./R/parse.R                   198
#> 2 identify_start_to_stop_of_roxygen_examples_from_text       6 ./R/roxygen-examples-find.R     9
#> 3 identify_start_to_stop_of_roxygen_examples                 4 ./R/roxygen-examples-find.R    16
#> 4 identify_r_raw_chunks                                     20 ./R/transform-code.R           71
```

I don't find a compelling reason anymore, so let's rename this functionality.

R/transform-code.R
```diff
@@ -110,6 +170,10 @@ get_engine_pattern <- function() {
 #'
 #' @inheritParams separate_chunks
 #' @keywords internal
-get_knitr_pattern <- function(lines) {
-  knitr::all_patterns[["md"]]
+get_knitr_pattern <- function(lines, filetype = "Rmd") {
+  if(filetype == "Rnw") {
+    knitr::all_patterns[["rnw"]]
+  } else {
```
lorenzwalthert: Here also I'd be explicit on the `else` part

R/ui.R
```diff
@@ -188,7 +188,7 @@ prettify_any <- function(transformers,
   )
 }
 
-#' Style `.R` and/or `.Rmd` files
+#' Style `.R` and/or `.Rmd` and.or `.Rnw` files
```
lorenzwalthert: typo here I think after and.

R/transform-code.R
```diff
@@ -60,16 +98,16 @@ separate_chunks <- function(lines) {
...
+#' Identifies raw Rmd code chunks
 #'
 #' Raw in the sense that these chunks don't contain pure R code, but they
 #' contain a header and footer of markdown. Only code chunks that have an engine
 #' whose name matches `engine-pattern` are considered as R code.
 #' @inheritParams separate_chunks
 #' @param engine_pattern A regular expression that must match the engine name.
 #' @keywords internal
-identify_r_raw_chunks <- function(lines, engine_pattern = get_engine_pattern()) {
-  pattern <- get_knitr_pattern(lines)
```
lorenzwalthert: We should remove the useless argument for this function.

R/transform-code.R
jonmcalder: 👍 (have applied this correction elsewhere too)

R/transform-code.R
```diff
@@ -15,40 +15,78 @@ transform_code <- function(path, fun, verbose = FALSE, ...) {
...
+#' Applies the supplied transformer function to code chunks identified within
+#' an Rmd file and recombines the resulting (styled) code chunks with the text
+#' chunks.
+#'
+#' @param lines A character vector of lines from an Rmd file
+#' @param transformer_fun A styler transformer function
+#' @importFrom purrr flatten_chr
+#' @keywords internal
+transform_rmd <- function(lines, transformer_fun) {
+  transform_mixed(lines, transformer_fun, filetype = "Rmd")
```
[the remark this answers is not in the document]
jonmcalder: Yeah I think `purrr:partial()` is more elegant. I've now reworked this since I no longer feel the explicit `transform_rmd()` and `transform_rnw()` functions are necessary.

R/transform-code.R
```diff
@@ -15,40 +15,78 @@ transform_code <- function(path, fun, verbose = FALSE, ...) {
...
 #' @param lines a character vector of lines from an Rmd file
 #' @importFrom purrr map2
 #' @importFrom rlang seq2
 #' @keywords internal
-separate_chunks <- function(lines) {
-  r_raw_chunks <- identify_r_raw_chunks(lines)
+separate_chunks <- function(lines, filetype = "Rmd") {
+  if(filetype == "Rmd") {
+    r_raw_chunks <- identify_rmd_raw_chunks(lines)
+  } else {
```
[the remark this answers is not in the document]
jonmcalder: 👍

R/transform-code.R
jonmcalder: Yeah I did wonder about this. Figured some function names just carried through from earlier implementations.  I've now merged the `Rmd` and `Rnw` cases into a single function to eliminate some of the code duplication, and have renamed this to `identify_raw_chunks()`. I think we were using 'raw' to identify the fact that the line references here encapsulate the whole R chunk as opposed to just the lines with code that get extracted by `separate_chunks()`. Happy to change this to something else if you want to though.

R/transform-code.R
jonmcalder: 👍

R/ui.R
jonmcalder: 👍

R/transform-code.R
jonmcalder: Wasn't 100% sure what you were referring to here but I think it might be redundant after the change I'm pushing now.

lorenzwalthert: The argument  `lines` I think.

jonmcalder: Oh right. Another weird code legacy 🤣

R/transform-code.R
```diff
@@ -89,6 +127,28 @@ identify_r_raw_chunks <- function(lines, engine_pattern = get_engine_pattern())
...
 }
 
+#' Identifies raw Rnw code chunks
+#'
+#' Raw in the sense that these chunks don't contain pure R code, but they
+#' contain the Rnw chunk header and footer. All Rnw code chunks are assumed
+#' to contain R code.
+#' @inheritParams separate_chunks
+#' @keywords internal
+identify_rnw_raw_chunks <- function(lines) {
```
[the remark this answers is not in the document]
jonmcalder: Yeah I didn't particularly like this duplication but couldn't think of a better way at the time - probably just a bit lazy! 😄

I think I've done it a little better this time around.

R/transform-code.R
```diff
@@ -12,43 +12,50 @@ transform_code <- function(path, fun, verbose = FALSE, ...) {
...
 #' Applies the supplied transformer function to code chunks identified within
-#' an Rmd file and recombines the resulting (styled) code chunks with the text
+#' an Rmd or Rnw file and recombines the resulting (styled) code chunks with the text
 #' chunks.
 #'
-#' @param lines A character vector of lines from an Rmd file
-#' @param transformer_fun A styler transformer function
+#' @param lines A character vector of lines from an Rmd or Rnw file.
+#' @param transformer_fun A styler transformer function.
+#' @param filetype A string indicating the filetype (Rmd or Rnw).
```
lorenzwalthert: I think here we could inherit `lines` and `filetype` from `separate_chunks()`.

tests/testthat/test-public_api.R
```diff
@@ -155,3 +155,28 @@ test_that("insufficient R version returns error", {
   expect_error(stop_insufficient_r_version())
 })
 
+context("public API - Rnw in style_file()")
+
+test_that("styler can style Rnw file", {
+  capture_output(expect_false({
```
lorenzwalthert: So for testing, I suggest we stick to the infrastructure we established before for consistency, i.e. use `test_collection()`. I think this would require to adapt `test_collection` to reckon the `.Rnw` pattern: 
https://github.com/r-lib/styler/blob/master/R/testing.R#L30

lorenzwalthert: This is used almost everywhere, e.g. here: https://github.com/r-lib/styler/blob/master/tests/testthat/test-indention_curly.R.

lorenzwalthert: Sorry, did not see the tests below. -.-

R/transform-code.R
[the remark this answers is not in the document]
lorenzwalthert: 👍

R/transform-code.R
lorenzwalthert: Thanks.

R/transform-code.R
lorenzwalthert: Ok cool.

R/transform-code.R
lorenzwalthert: Indeed.

R/transform-code.R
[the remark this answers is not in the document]
lorenzwalthert: Agree.

R/transform-code.R
jonmcalder: 👍

tests/testthat/test-public_api.R
jonmcalder: 😄

jonmcalder:
Actually I didn't really have any edge cases in mind and it's been years since I last used `Sweave`. Just figured more test coverage would be needed.

Aren't blank code chunks already covered by the styling test on [random2.Rnw](https://github.com/r-lib/styler/pull/434/files#diff-a576dc70bf12326c8e023c9dc792a988)?

R/transform-code.R
```diff
@@ -60,32 +67,39 @@ separate_chunks <- function(lines) {
...
+    ends <- even(chunks)
+    is_r_code <- grepl(
+      paste0("^[\t >]*```+\\s*\\{\\s*", engine_pattern, "[\\s\\},]"),
+      lines[starts],
+      perl = TRUE
+    )
+  } else if (filetype == "Rnw") {
+    starts <- grep(pattern$chunk.begin, lines, perl = TRUE)
+    ends <- grep(pattern$chunk.end, lines, perl = TRUE)
+    is_r_code <- rep(TRUE, length(start))
```
lorenzwalthert: I think we are missing an s here

```suggestion
    is_r_code <- rep(TRUE, length(starts))
```

Not causing any problems because
```r
start
#> function (x, ...) 
#> UseMethod("start")
#> <bytecode: 0x107af2ed8>
#> <environment: namespace:stats>
```

lorenzwalthert:
> Aren't blank code chunks already covered by the styling test on random2.Rnw?

Agree to the extent that a blank line is inserted if there is none. But this was the case already with `.Rmd` so I guess it's fine. 😜

R/transform-code.R
jonmcalder: Well spotted!

R/ui.R
```diff
@@ -19,8 +19,8 @@ NULL
 #'   conveniently constructed via the `style` argument and `...`. See
 #'   'Examples'.
 #' @param filetype Vector of file extensions indicating which file types should
-#'   be styled. Case is ignored, and the `.` is optional, e.g. `c(".R", ".Rmd")`
-#'   or `c("r", "rmd")`.
+#'   be styled. Case is ignored, and the `.` is optional, e.g. `c(".R", ".Rmd",
+#'   ".Rnw")` or `c("r", "rmd", "rnw")`.
```
lorenzwalthert: Can you make adjustments in the description of this function in line 8 too (we have not even adopted it to `.Rmd` yet)?

lorenzwalthert: Also here: https://github.com/r-lib/styler/pull/434/commits/6b56ba48d41f30fb66b6949abac73955c7263d2b#diff-4f632064d1ef846baad1bfd3b7fc8dffR149

jonmcalder: 💯

lorenzwalthert:
Thanks a lot @jonmcalder. Nice.

PR #449, opened 2018-12-04
Add yihui_style()

Robinlovelace:
A little context: https://github.com/ropensci/software-review-meta/issues/71#issuecomment-444206853

PR #451, opened 2018-12-06
Add equals_style()

Robinlovelace:
Rebased from master.


Robinlovelace:
Copied-over from #449

lorenzwalthert:
To do after merge: reorganize file style-guides into smaller files.

lorenzwalthert:
Sorry @Robinlovelace for some reason these comments were just posted now because I clicked *start a review* and never finished instead of just commenting one-off. I wrote them a few weeks ago :-(

Robinlovelace:
Hi @lorenzwalthert thanks for the comments. I also just saw these! Will aim to look this evening but not guarantees...

lorenzwalthert:
Because of https://github.com/r-lib/styler/issues/513, I am not sure if we could and should ever replace `<-` with `=`.

Robinlovelace:
Yes in edge cases were `<-` is used in a function call replacing with `=` can fail. What are your thoughts on this? In some ways `<-` to `=` is easier than `=` to `<-` because a global search and replace will sort it (assuming the code has no edge cases where objects are assigned mid function call), so automation in a function is less important: 

```r
gsub(pattern = "<-", replacement = "=", x = "x <- mean(cars$speed, na.rm = TRUE)")
```

works.

Still I think it would be great to support styling to equals assignement.

lorenzwalthert:
Well, it's rarely a good idea to work with regular expressions I think. It's rather slow, unreliable and does not fit the styler framework. We solved the problem in #513 in this way: https://github.com/pat-s/styler/pull/1/commits/5bf20ab358ba4d8c0f4a9b26f450b08bee6d616d. To use `next_terminal()`, make sure you merge r-lib/styler into your fork. Then it should work.

Robinlovelace:
Great. Fully agree a non-regex solution would be ideal. Not sure how it relates to this PR though, I've lost track of where we were and next steps I'm afaid!

Robinlovelace:
Sensible decision, would you welcome a rejuvenation of this? Would be useful for ppl using the style in their work. I've been using `styler` to tidy up my code and replace `=`  with `<-` in projects that use that, e.g.: https://github.com/ropensci/stplanr/commit/33158a5bff0f8c4191a7d317439de897e1c667d7

Would be great to be able to go in the opposite direction, e.g. here: https://github.com/ropensci/stats19/blob/master/inst/iow_example.R

lorenzwalthert:
Sorry, I did not mean you to delete the branch entirely. I should probably have asked about the status of this before closing. Sorry. Yes, you can create another PR.

Robinlovelace:
OK great. I think starting over (again!) may be the best way forward in general, but I'm sure of the specific next steps. Will see if there is scope to combine efforts with @pat-s on this. As always any pointers/thoughts welcome and no worries about closing the PR, it had definitely gone stale!

pat-s:
@Robinlovelace If you refer to the "mlr-style", @lorenzwalthert did most of the work. I essentially just tested and used it since then :) (and bother him once in a while with some upstream merge issues 😄 )

Issue #319, opened 2018-01-07
storing styler config in a yaml file


lorenzwalthert:
Repo may include a file `.styler.yaml` that contains a configuration for styling, like this:
```r
style_guide:
  source: styler
  style: tidyverse_style
  options:
    strict: TRUE
    scope: tokens
```

When `styler::style_pkg()` is run, such a config file is looked for in the root directory and if one is found, the options specified there should be given precedence over what is specified in the function call. Or something like that. WIP @ https://github.com/lorenzwalthert/styler/tree/config

lorenzwalthert:
Maybe also non-hidden `styler.yaml`. Also needs new infrastructure in root directory as #320.

pat-s:
Are there any plans to tackle this again in the future? The ideal case would be that RStudio also picks up this settings file.

lorenzwalthert:
Also, consider how other formatters have solved this problem, e.g. https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/

lorenzwalthert:
This is also relevant for the save on styling functionality provided via an environment variable (`$save_on_styling`.) without namespace or language prefix.

lorenzwalthert:
Until we make progress here (if at all), you can also use [RStudio snippets](https://support.rstudio.com/hc/en-us/articles/204463668-Code-Snippets) to store your preferred styler command, e.g. with
```r
snippet style
	styler::style_file(${1:file}, strict = TRUE, scope = "spaces", indent_by = 4)
```

You can type `style` in the RStudio console and it would expand to the command above. Just to save typing

pat-s:
Note that you can also run R code within a snippet. The placeholder inside the snippet could default to the file that is currently opened in your editor. 

Then it almost behaves like a keybinding without the need to specify the file again 🙂

eitsupi:
Hello.
As with many auto-formatters, it would be great to be able to change the behavior in the definition file.
It looks like you haven't made progress for a while, do you have something like a todo list?

Thanks for all the great work you do.

lorenzwalthert:
I think I’d prefer to solve this outside {styler}, eg with {fallback} as referenced above. I am happy to transfer the repo to someone else if they want to continue work on it, because I don’t have time.

lorenzwalthert:
Maybe {renv} would want to expose their mechanism so we don't have to re-invent the wheel? Upvotes welcome 😀  https://github.com/rstudio/renv/issues/1865

PR #463, opened 2019-01-23
Allow selecting the style used by RStudio addins.

riccardoporreca:
* Option `"styler.addins.style"` stores the selected style.
* New addin to set the style.
* Existing addins adapted to use the selected style.


R/addins.R
```diff
@@ -94,3 +96,78 @@ style_selection <- function() {
 get_rstudio_context <- function() {
   rstudioapi::getActiveDocumentContext()
 }
+
+
+# Dedicated binding for package styling addin. Simple wrapper calling style_pkg
+# with the selected addins style.
+style_package <- function() {
```
lorenzwalthert: I think we don't want another Addin, see also #250.

lorenzwalthert:
First of all, thanks @riccardoporreca for the PR. I think I need to test it myself and then I will review the PR.

riccardoporreca:
@lorenzwalthert, sure, I will also be happy to discuss if, despite the reasonable point of not cluttering the addins, the concept and utilities for setting and using a given style guide for the addins could be still useful.

lorenzwalthert:
Thanks. Just to make it clear, I see that the PR has two parts and I also think the functionality about style selection is something we'd want to support. Apart from the implementation details, one thing I am not sure about is how to deal with global variables and environment variables and default values. I created [fallback](https://github.com/lorenzwalthert/fallback) and integrated it into styler (compare #319), but I am not sure if it's the best solution.  I.e. if the default style should be set in a `yaml` file.

lorenzwalthert:
I changed my mind and we will defer the questions regarding default values and just go with R options for now. Hope to find time to review soon.

lorenzwalthert:
thanks @riccardoporreca. There is one minor thing about error messages which I will fix when working on #472. The Add-in has passed my manual test with the [oneline style guide](https://lorenzwalthert.netlify.com/post/oneliner/) too, so it's all fine 😄.

riccardoporreca:
Great, well spotted with the error message and the concise rlang alternative.

Thanks @lorenzwalthert for the nice collaboration on the PR.

lorenzwalthert:
@riccardoporreca I just realized that instead of asking for a style, we should maybe ask for transformers, because then, people could also specify arguments thereof, e.g. the scope:
```r
style_text(transformers = tidyerse_style(scope = "spaces", ...)
```

This is not currently possible because the input for the Addin is a style, i.e. 

```r
styler::tidyverse_style
```

and not transformers, e.g.
```r
styler::tidyverse_style()
```

riccardoporreca:
> @riccardoporreca I just realized that instead of asking for a style, we should maybe ask for transformers, because then, people could also specify arguments thereof

@lorenzwalthert, I agree.

I am not sure how you would describe the addin with this new approach, "transformers" is somehow a less intuitive term than "style" for the user, so maybe the addin can still be called "Set Style" and described as "Prompt for and set the **style transformers** used by all STYLER addins", or something more sensible. Similar for the prompt and error messages.

R/addins.R
lorenzwalthert: Actually thanks go @jonmcalder we have figured out that you have not registered a new Addin, just moved existing code. I overlooked that. Sorry. Will add in #500 again.

riccardoporreca: I thought this was intentional, as discussed when integrating your tweaks in https://github.com/miraisolutions/styler/pull/1#discussion_r256412827

lorenzwalthert: Well, the confusion arose because people wanted to add other Addins like #476 and #250 and in the diff, it looked as if you *added* an Addin for package styling. This was not the case, you just moved some code. So that was my mistake.

riccardoporreca: I see, no problem!

PR #414, opened 2018-08-02
Wrap long lines

lorenzwalthert:
Reference: #390.

PR #496, opened 2019-04-22
Wrap long lines (follow up)

lorenzwalthert:
Follow-up of #414.

### DIFF
--- a/R/rules-line-break.R
+++ b/R/rules-line-break.R
@@ -136,3 +136,33 @@ remove_line_break_in_empty_fun_call <- function(pd) {
   }
   pd
 }
+
+mk_add_line_break_at_pos_id <- function(ids) {
+  ids <- ids
+  
+  function(pd_flat){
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
Status of this PR: 

- worked at some point but not for all cases I experimented with.
- probably there are merge conflicts by now, has to be rebased on master. I think this PR is conceptually a good approach, but the implementation may need substantial rework.
- After the experience with caching and stylerignore feature in v1.3.0, I think we should make this an opt-in feature governed by an R option for a few releases. Also, styler will be slower when this is turned on.
- as this is considered to be invasive, it should only take effect for `strict = TRUE`.
- Not sure if we should takle the problem for different expression types individually instead of aiming for a general solution. Prime candidate: function calls. I think the overwhelming majority of long lines are function calls (potentially nested in assignment expressions). Related: #544.

I am not going to contribute code to this feature soon though.

PR #528, opened 2019-06-29
Support for curly-curly syntactic sugar

### LINKED ISSUE #526: styler fails to correctly style syntactic rlang sugar {{
Hi!
I noticed that the new `{{ }}` operator from `rlang` is not styled properly; for instance, if I style the following lines

```r
library(rlang)
name <- "Jane"
list2({{name}} := 1 + 2)
```

I get the following:

```r
library(rlang)
name <- "Jane"
list2({
  {
    name
  }
} := 1 + 2)
```

Not sure what would be a good way of handling this specific operator (maybe @lionel- has an idea), I just thought about reporting it.

Anyway, thanks for `styler`: absolutely love the package!

Alessandro

lorenzwalthert:
Closes #526. Affecting both line break rules and spacing rules. Preserve testing of previous code (mostly focussed on indention) related to using two curly braces in a sequence was preserved, as we need this to work independently of the rule for `{{`. This is implemented using a version of the tidyverse style guide that *turns off* the rule for curly-curly.


lorenzwalthert:
The problem with this PR is that it does not allow `{{` to be on a separate line in a function call: 
``` r
# this is bad when the function call name + indention is so long that we simply want the 
# first argument on a new line.
styler::style_text(
  "call(
  {{x}}
)"
)
#> call({{ x }})

# or less common
styler::style_text(
  "call(1, 2, 3, 4,
  4, 5,
  {{x}}, 
  5
)"
)
#> call(
#>   1, 2, 3, 4,
#>   4, 5, {{ x }},
#>   5
#> )
```

<sup>Created on 2019-06-29 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

lorenzwalthert:
Problem from above resolved with ba2d45d:
``` r
styler::style_text(
  "call(
  {{x}}
)"
)
#> call(
#>   {{ x }}
#> )
# or less common
styler::style_text(
  "call(1, 2, 3, 4,
  4, 5,
  {{x}}, 
  5
)"
)
#> call(
#>   1, 2, 3, 4,
#>   4, 5,
#>   {{ x }},
#>   5
#> )
```

<sup>Created on 2019-06-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

PR #536, opened 2019-08-05
Function declarations should gain braces for body if strict = TRUE

### LINKED ISSUE #535: Why is the indent different in functions without braces?
I'm using a hammer `h()` to turn off `R CMD check` warnings for functions that use NSE. Is there an easy way to make the `bad` (and then perhaps also the `use_case`) example behave?

``` r
styler::style_text("
good <- function() {
  h(
    code
  )
}
bad <- function() h(
  code
)
use_case <- function() h(~ {
  code
})
")
#> 
#> good <- function() {
#>   h(
#>     code
#>   )
#> }
#> bad <- function() h(
#>     code
#>   )
#> use_case <- function() h(~ {
#>     code
#>   })
```

<sup>Created on 2019-08-05 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lorenzwalthert:
Also add braces to multi-function declaration if missing for `strict = TRUE` (closes #535):
``` r
library(styler)
style_text("function() h(~ { ... })")
#> function() {
#>   h(~ {
#>     ...
#>   })
#> }
```

<sup>Created on 2019-08-05 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Calls of that form had too much indention, see #535.
For `strict = FALSE`, no braces are added and the following cases were fixed to remain as is:

```r
if (x) h(
  x
) 

for (i in x) h(
  x
) 

while (x) h(
  x
) 

if (x) h(
  x
) else k(
  b
)
```

They all used to have the second line twice indented.

The solution is in almost entirely in 7c26637. Simply check if a new line follows after the closing brace. Does not need to use `needs_indention()` because if a token follows on `)` on the same line, it will do the indention if necessary, which is exactly the case if it is multi-line. The other cornerstone is to lower-bound indention at 0. Because `indent_without_paren_for_while_fun()` does not know if `strict = TRUE` and curly braces will be added later, a problem is created for function declarations where the body is not wrapped in `{...}`: (1) `indent_without_paren_for_while_fun()` does not indent on `FUNCTION` because `{` follows. (2) `wrap_subexpr_in_curly()` creates a curly expression from the body and wants to unindent the body, which was not indented. This creates negative indention. (3) serialization fails due to negative indention. This is avoided by lower-bounding the indention adjustment in `wrap_subexpr_in_curly()` to 0. It seemed more robust than trying to establish at runtime in `indent_without_paren_for_while_fun()` if `wrap_subexpr_in_curly()` will be called later.


lorenzwalthert:
It seems this: 

``` r
library(styler)
style_text("function() h(~ { ... })")
#> function() {
#>   h(~ {
#>     ...
#>   })
#> }
style_text("function() h(~ { ... })", strict = FALSE)
#> function() h(~ {
#>   ...
#> })
```

<sup>Created on 2019-08-05 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #538, opened 2019-08-15
Cache styling

### LINKED ISSUE #320: Memorize which files were not changed since last styling
Could significantly speedup `style_file()` and `style_dir()` when used often as well as pre-commit hooks. ~Needs new infrastructure `.styler` folder in root directory.~

lorenzwalthert:
Closes #320. 

# Goal

The goal is to make styler remember what it styled so the pre-commit hook in
https://github.com/lorenzwalthert/pre-commit-hooks is much faster and functions
like `styler::style_pkg()` are much faster if run on already styled files.

# Requirements

The requirements are:

* must work on all plattforms and must not be error-prone.

* the user should be able to manage it (know its size and location, deactivate,
  clear, delete)

* should be as close to the functionality "once seen, by style_text(), Addin or
  style_file() remembered forever".

* The cache should be default be enabled, but the required packages should be in
  `Suggests` to keep the core installation lightweight. If `R.cache` is not
  installed, we should issue a warning message, deactivate the caching feature
  for the current R session and ask the user to install the dependencies and or
  permanently disable the feature in their .Rprofile with
  usethis::edit_r_profile().

* For this reason, the styling should also work when `R.cache` is not installed,
  which requires every `R.cache` call to be wrapped in a conditional.

* The advanced user should be able to understand how R.cache was used to
  implement the caching.

# Conceputal

In this PR we introduce caching to styler. We follow the approach outlined in
https://github.com/r-lib/styler/issues/320#issuecomment-520747381, which basically is:

* check if text to style is in cache.

* if not, style it and add styled text to cache.

The approach has the following advantages:

* Very simple to implement.

* API agnostic. Works for files, text, Addin because it operates on the text
  level, which is quite low level.

* Because approach is path and modification time independent, it can cache the
  same content in different locations, including renaming, copies in multiple
  places as well as moving files. Can cache multiple versions of a file, e.g. on
  different branches, when going back and forth in the git history.

* Cache remains very small because no actual code is cached.

The cache must be styler version specific because if not, updated styling rules
won't be applied.

# Implementation

* We use `R.cache` to power the caching.

* We use R options to manage it, with additional functional wrappers to modify
  the options.

* There must be one cache per styler version and for testing purposes, the user
  must be able to specify a cache maunally (mainly to be able to delete the
  cache of the tests without deleting the cache he uses as a user).

* To not convolute the `R.cache` caching directory, we use a `dir` under the
  caching root that corresponds to `/styler/cache_name`, where `cache_name` is
  the use specified cache name, defaulting to the installed version number from
  DESCRIPTION.

* We modify `.travis.yaml` to also test behavior if `R.cache` is not installed.

*** 
Todo: 

- [x] Should `R.cache` be in suggest and certainly fail in a vanilla installation of styler when caching is not explicitly disabled and `R.cache` is not installed? No, just warning as described above.
- [x] Think about using the cache mechanism in testing (CI and local) as well as development (local).
- [x] What is the default if the package is not loaded. Then, the option is not set.
- [x] Central place to document the options. Not needed since we use wrapper functions to manipulate.
- [x] What happens if `R.cache`  is not installed?
- [x] Which cache functions should be exported? Probably the ones we have now.
- [x] Tests for cache accessor functions.
- [x] Test for caching in conjunction with options.
- [x] Tests for caching regarding EOL blank line.
- [x] Document this PR and design choices in more detail


lorenzwalthert:
With the latest two commits, we use my fork of R.cache to store empty files for caching, as suggested in https://github.com/HenrikBengtsson/R.cache/issues/37.

krlmlr:
Is this creating one file per expression, or one file per file processed?

lorenzwalthert:
One per file. I thought we could also do one per expression, it would give an additional speed boost in case you have big expressions and modify just one. The implementation would also be more complex though. I think as of now, it's pretty slim (excluding all the cache management tools that most people will never touch). However, there is (obviously) a trade-off with file size as long as caching costs us one block, e.g. 4KB memory on macOS.

krlmlr:
Can we do one cache object per R project?

lorenzwalthert:
Do you mean [RStudio Projects](https://support.rstudio.com/hc/en-us/articles/200526207-Using-Projects) associated with a working directory?

krlmlr:
Project = package, analysis project, RStudio project -- in the sense of {here} or `usethis::proj_get()`.

lorenzwalthert:
I am not sure I understand what you mean. I don't think we should convolute the project directory with a styler cache to make the caching directory specific. What do you think would be the advantages of this approach? Advantages of a central cache: 

* Can be removed very easily, not scattered in many places.
* Not yet another file to `gitignore`, `.Rbuildignore` or whatever in every project repo.
* The same code in different places can be cached easily.

Also, I think it would not help solving the problem with the block size. Also, if possible, I'd like to use `R.cache` because: 

* CRAN policy is comparatively string on writing to anywhere on the file system.
* Offload implementation of cache to another package keeps complexity low, e.g. initialization, error handling.

DESCRIPTION
```diff
@@ -27,7 +27,8 @@ Imports:
     tibble (>= 1.4.2),
     tools,
     withr (>= 1.0.0),
-    xfun (>= 0.1)
+    xfun (>= 0.1),
+    R.cache (>= 0.13.0.9000)
```
krlmlr: Did you mean to add {R.cache} to "Suggests"? Leaving it in "Imports" seems easier.

NEWS.md
```diff
@@ -12,9 +12,16 @@
...
-  [definition for aligned function calls](https://styler.r-lib.org/articles/detect-alignment.html)
-  (#537).
+* styler caches results of styling, so applying styler to code it has styled
+  before will be instantaneous. This brings large speed boosts in many
+  situations, e.g. when `style_pkg()` is run but only a few files have changed
+  since the last styling or when using the [styler pre-commit
+  hook](https://github.com/lorenzwalthert/pre-commit-hooks). See the README for
+  details (#538).
+
+* Aligned function calls are detected and kept as is if they match the styler
```
krlmlr:
```suggestion
* Aligned function calls are detected and remain unchanged if they match the styler
```

R/ui-caching.R
```diff
@@ -0,0 +1,102 @@
+#' Clear the cache
+#'
+#' Clears the cache that stores which files' styling is up-to-date. You won't be
```
krlmlr:
```suggestion
#' Clears the cache that stores which files are already styled. You won't be
```

R/transform-files.R
```diff
@@ -72,17 +72,37 @@ transform_file <- function(path,
...
-        parse_transform_serialize_roxygen(., transformers),
-        ~.
-      )
-    transformed_code
+    is_cached <- rlang::is_installed("R.cache") && !is.null(
+      R.cache::findCache(key = hash_standardize(text), dir = cache_dir)
+    )
+    should_use_cache <- cache_is_activated()
+    use_cache <- is_cached && should_use_cache
+    if (!use_cache) {
```
krlmlr: Add an early return like `if (use_cache) return(text)` ?

R/utils-cache.R
```diff
@@ -0,0 +1,57 @@
+#' Standardize text for hashing
+#'
+#' Make sure text after styling results in the same hash as text before styling
+#' if it is indeed identical.
+#' @param text A character vector.
+#' @keywords internal
+hash_standardize <- function(text) {
+  text <- ensure_last_is_empty(text)
+  Encoding(text) <- "UTF-8"
```
krlmlr: Do you really want to override encoding?

```suggestion
  text <- enc2utf8(text)
```

R/utils-cache.R
```diff
@@ -0,0 +1,57 @@
...
+    !is.null(current_cache)
+  } else if (!is.null(current_cache)) {
+    cache_name == current_cache
+  } else {
+    FALSE
+  }
+}
+
+cache_derive_name <- function() {
+  utils::packageDescription("styler", fields = "Version")
```
krlmlr: Can we derive without looking at the package library? Maybe store in a global variable, this is evaluated during installation?

R/transform-files.R
```diff
@@ -72,17 +72,37 @@ transform_file <- function(path,
...
-        include_roxygen_examples ~
-        parse_transform_serialize_roxygen(., transformers),
-        ~.
-      )
-    transformed_code
+    is_cached <- rlang::is_installed("R.cache") && !is.null(
+      R.cache::findCache(key = hash_standardize(text), dir = cache_dir)
+    )
+    should_use_cache <- cache_is_activated()
+    use_cache <- is_cached && should_use_cache
```
krlmlr:
```suggestion
    use_cache <- should_use_cache && is_cached(text)
```

with a suitable `is_cached()` function avoids calling `hash_standardize()` if not needed.

DESCRIPTION
lorenzwalthert: I see it was only added in the very last commit `cb2dc17363dfaa39f4025a1895ffed86556e5ceb` when I added my own fork as a remote dependency. Probably caused by unintentional `usethis::use_dep()`. Do you think it's unnecessary to keep it in `Suggests`? It needs some extra handling in other places, I agree. 
``` r
setdiff(miniCRAN::pkgDep("R.cache"), miniCRAN::pkgDep("styler"))
#> [1] "R.cache"     "R.methodsS3" "R.oo"        "R.utils"
```

<sup>Created on 2019-09-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

R/utils-cache.R
lorenzwalthert: Not during installation, but `.onLoad()`. The cache is version specific because a new version of styler potentially styles code differently, so I need the version number and I don't know what a better API would be to it. Suggestions?

R/utils-cache.R
lorenzwalthert: Right... Can't remember why I did it that way. At least it does not seem to break tests when accepting your solution so let's do it.

lorenzwalthert:
Thanks for reviewing @krlmlr. Feels like the good old times are back :tada: .

R/utils-cache.R
krlmlr: If you write `v <- read.dcf("../DESCRIPTION")[...]` in an R file, I think it's evaluated when the package is built. Can you please double-check?

lorenzwalthert: `v <- read.dcf("DESCRIPTION")` seems to work locally yes. Quick Googling did not help. However, if it is brittle, I think I'd prefer the other approach, it's a detail. I implemented it in a2be25905ceee35e4f780a4f7b1cbbbfe8e3d0f7 and left the function call. Alternative would be to directly use a variable `version` instead of the call that derives it.

krlmlr: I feel it's more robust than looking at the library, because the version in your library might be different from the version loaded. Better use `version <- read.dcf(...)[, "Version"]` though.

krlmlr:
Do we need to include the style in the hash key, so that we recompute when the style changes?

lorenzwalthert:
I guess we should. The question is how. Because currently, the version of styler determines which cache to use by default. Also, if people have their own style guide they supply, we should probably also hash that function and include it in the hash, leaving the R.cache directory structure as is, i.e. 

`.R.cache/styler/:version/:hash(transformer, text)`

lorenzwalthert:
> Do we need to include the style in the hash key, so that we recompute when the style changes?

I can't get two partialized calls to be identical which is a requirement for this to work:

```r
add <- function(x, y) {
  x + y
}

add1 <- purrr::partial(add, x = 1, .env = rlang::empty_env())
add2 <- purrr::partial(add, x = 1, .env = rlang::empty_env())
identical(add1, add2)
#> FALSE
```
 https://stackoverflow.com/questions/58656033/when-are-purrrpartial-ized-functions-identical

Do you know the solution?

lorenzwalthert:
Ok, SO question solved, now we know that we need the same environment for these functions (probably pre-created and passed as `.env` to `purrr::partial()`, which does seem to have a bug and reported in https://github.com/tidyverse/purrr/issues/717).

krlmlr:
I tweaked a bit, works for me with {dm}.

- How can we communicate to the caller that a cached values is used? I think for this we need to get rid of `transform_utf8()` and intercept at an earlier stage.
    1. Read file
    1. Check if cached
    1. If not, style
    1. If changed, cache and write
- Can we map empty files to `NULL` in {R.cache}?

lorenzwalthert:
> Can we map empty files to NULL in {R.cache}?

Tracked in https://github.com/HenrikBengtsson/R.cache/issues/37.

> How can we communicate to the caller that a cached values is used?

I am not sure this is necessary because it seems quite cumbersome to implement. You mean in the console output or also in the data frame returned invisibly by `style_pkg()` and friends? For consistency, I think we'd need to do that in both or none. The reason we check if `text` is cached in `make_transformer()` is because all style API functions (like the Addins, `style_text()`, etc.) all use this function. 

Also, if you want to read in only once (i.e. `transform_utf8_one()`) and use a functional approach (i.e. not storing this information in an environment), it would require adapting the code in many places and pass back the information about whether or not the cache was used. it would have to be propagated up the chain in this order I think:

`transform_file()` <- `transform_code()` <- `transform_utf8()` <- `transform_utf8_one()` <- `make_transformer()`

R/utils-cache.R
lorenzwalthert: ok

lorenzwalthert:
I think we should not communicate to the user if the cache was used, at least not in this PR. It involves refacturing quite a few things and the PR is already getting too large.

lorenzwalthert:
Seems that with https://github.com/r-lib/styler/pull/538/commits/36606a2a52828fcf959ee025e6e17e268b1253b9, we are able to consume 0 KB instead of block size per cached value.

krlmlr:
Plus the inode, but these are cheap.

Let's see how it works without notifying the user.

lorenzwalthert:
There are two remaining issues with this PR:

- [ ] transformers should have a version, so we don't need to cache based on text of transformers (as this won't work in general for third party style guides) and revise documentation for it. This affects the mlr style guide.
- [ ] Caching could be improved and treated in a new issue.

We will merge this and open new issues for these tasks.

PR #543, opened 2019-09-17
Force brace expression in function calls be on their own line

### LINKED ISSUE #428: Braces in function calls
Can we keep the newline after `(` here?

``` r
text <- '
eventReactive(
  {
    code()
  },
  {
    code()
  }
)
'

styler::style_text(text)
#> 
#> eventReactive({
#>   code()
#> }, {
#>   code()
#> })
```

<sup>Created on 2018-09-28 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

lorenzwalthert:
The main motivation is to avoid `tryCatch()` expressions to be re-formatted compactly:
``` r
styler::style_text("tryCatch(
  {
    x
  }, error = function(e) e
)")
#> tryCatch({
#>   x
#> }, error = function(e) e)
```

<sup>Created on 2019-09-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

While at the same time, preserve formatting of `testthat()` expressions:

``` r
styler::style_text("testthat(x, {
  code(to(exec))
})")
#> testthat(x, {
#>   code(to(exec))
#> })
```

<sup>Created on 2019-09-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

The abstract rule was formulated in https://github.com/r-lib/styler/issues/428#issuecomment-473192227. 
> Function arguments that consist of a braced expression always need to start on a new line, unless it's the last argument and all other arguments fit on the line of the function call or they are named.

The last part of the sentence *or they are named.* is needed for this case:
```r
testthat(x, 
  b = 
  {
    5
})
```
 
Where we should rearrange to
```r
testthat(x, 
  b = {
  5
})
```

Closes #428.


tests/testthat/curly-curly/mixed-out.R
```diff
@@ -59,11 +59,14 @@ call({{ x }})
...
 ##  multiple                                                                ####
-call({
-  1
-}, a + b, {
-  33 / f(c)
-})
+call(
+  {
+    1
+  }, a + b,
```
krlmlr: I think the `a + b` needs to go on a separate line. Is this related to this pull request at all?

Rationale: Multiline expressions in function calls don't contain code from other arguments.

This may be missing from the style guide, but seems obvious to me: If an argument is so complex that it needs multiple lines, we don't want it interspersed with other arguments.

lorenzwalthert: In the above example, I agree. However, when we have

```r
call(
  {
    x + y
  }, {
    f(x)
    call(x = 2, f = max(n))
  }
)
```

I'd leave it as is opposed to 
```r
call(
  {
    x + y
  }, 
  {
    f(x)
    call(x = 2, f = max(n))
  }
)
```

Because indention helps seeing what belongs together and `}, {` for me just reads as a separator of two expressions (i.e. I don't read them as two braces anymore). What do you think.

krlmlr: Can we enforce the second option in `strict` mode but leave the first in `permissive` mode?

lorenzwalthert: I'd leave the first case as is, but I would not change the second into the first. So if your code has `}, \n{`, styler should not change it to `}, {`, and vice versa. Whenever there is brace + comma + not a brace, I agree to break lines in any case. The `strict` argument is already too convoluted I think. If your suggestion gets accepted in tidyverse/style, I am happy to also implement it for `strict = TRUE`, but I think it's a detail anyways.

krlmlr: Works for me. Thanks for looking into it!

lorenzwalthert: Ok no this is more complicated to implement. I'll stick to your rule.

tests/testthat/curly-curly/mixed-out.R
```diff
@@ -59,12 +59,34 @@ call({{ x }})
...
 ##  multiple                                                                ####
-call({
+call("test", {
   1
-}, a + b, {
-  33 / f(c)
 })
 
+call(
+  "test", {
```
krlmlr: So the newline isn't added here because it's the last argument? Can we easily check if the brace is on the same line as the opening parenthesis of the function call?

tests/testthat/curly-curly/mixed-out.R
```diff
@@ -59,12 +59,34 @@ call({{ x }})
...
+call("test", {
+  1
+})
+
+call(
+  {
+    1
+  },
+  a + b,
+  {
```
krlmlr: I love this! But why is the newline added here and not in the example above?

tests/testthat/curly-curly/mixed-out.R
lorenzwalthert: Not easily but we can in `eaac448d22ec000d99030a367852c5bae9b6ceb5` (pushed to master). :-)

PR #553, opened 2019-10-10
dispatch cases where \ has different meaning.

### LINKED ISSUE #551: Styler does not preserve new line characters
Long strings that have new line characters have literal new lines added after styling:

``` r
code <- "                                                              
function() {                                                           
paste('I am loooooooooooooooooooooooooong, and I have a new line: \n')
}                                                                      
"                                                                      
styler::style_text(code)                                               
#> function() {
#>   paste("I am loooooooooooooooooooooooooong, and I have a new line: 
#> ")
#> }
```

Ideally these characters would be preserved.

lorenzwalthert:
Closes #551 by implementing the suggested solution outlined in https://github.com/r-lib/styler/issues/551#issuecomment-540788608. Put that into a file and style it with styler, all `\` should be preserved, only quotes should be replaced (does not work well with reprex).
```

'here
is a string
'
'here\nis a string
'
'here\\nis a string'

```

Note that for mixing the invisible line break and \n in single quotes, we leave it as is (second example).


lorenzwalthert:
@krlmlr can you, if you don't see a better way than my solution with the case distinction, check if I handle raw objects in a reasonable way? E.g. to get a character vector from raw, I brutally `class(raw) <- "character"`.

lorenzwalthert:
Closed in favor of the more concicse, elegant and general implementation (can handle mixed case) in #554.

PR #554, opened 2019-10-11
Fix handling of newlines when converting quotes

### LINKED ISSUE #551: Styler does not preserve new line characters

krlmlr:
I remember using `deparse(parse())` to avoid edge cases like `'\\\''`. The non-obvious `gsub()` really seems like the simplest solution to ensure we always match one character after *each* backslash.

Closes #551.


lorenzwalthert:
Awesome, thanks. This is much more elegant and even more general than #553. I am always afraid of `gsub()` / `grep()` with escape characters 😨.  Maybe now is the time to overcome that? Then maybe you can tell me why do *all* three paterns match?

``` r
grepl("\n", "\n")
#> [1] TRUE
grepl("\\n", "\n")
#> [1] TRUE
grepl("\\\n", "\n")
#> [1] TRUE
```

<sup>Created on 2019-10-11 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

michaelquinn32:
Thank you!!!

krlmlr:
`"\n"` matches because it's a literal newline character.

`"\\n"` matches because backslashes are treated specially in regex -- and backslash + n happens to match a newline.

`"\\\n"` -- :face_with_head_bandage: it seems that backslash + `\n` *also* matches a newline. It's true for both `perl = TRUE` and `perl = FALSE`, so seems reliable. Also this means that we can seek for answers in a broader context.

Luckily, it breaks with the fourth backslash.

``` r
grepl("\\\\n", "\n")
#> [1] FALSE
```

<sup>Created on 2019-10-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lorenzwalthert:
Thanks Kirill.

PR #557, opened 2019-10-13
add emacs reformatter instructions

### LINKED ISSUE #517: styler in Emacs via reformatter
Just an FYI, I've written a small snippet to use styler in Emacs. 

Issue about it in `reformatter.el`
https://github.com/purcell/reformatter.el/issues/11#issuecomment-495459306
`snippet`
https://github.com/Emiller88/doom-emacs-private/blob/6aedee9bb7a5624f1dd63fcd5b6534aad400101a/config.org#reformatter

If you're reading this and you're not a Doom emacs user, `def-package!` is just a very thin wrapper around `use-package`

Emiller88:
Closes #517

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -25,7 +25,7 @@ Status](https://ci.appveyor.com/api/projects/status/github/r-lib/styler?branch=m
 [![cran
 version](https://www.r-pkg.org/badges/version/styler)](https://cran.r-project.org/package=styler)
 
- 
+
 The goal of styler is to provide non-invasive pretty-printing of R source code
 while adhering to the [tidyverse](https://style.tidyverse.org) formatting rules.
 styler can be customized to format code according to other style guides too.
@@ -40,7 +40,7 @@ install.packages("styler")
 
 Or get the development version from GitHub:
 
-```{r, eval = FALSE} 
+```{r, eval = FALSE}
 # install.packages("remotes")
 remotes::install_github("r-lib/styler")
 ```
@@ -95,7 +95,7 @@ While spaces still got styled (around `=` in `(x)`).
 
 If you wish to keep alignment as is, you can use `strict = FALSE`:
 
-```{r} 
+```{r}
 style_text(
   c(
     "first  <- 4",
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
I believe it would make more sense to link to https://github.com/Emiller88/doom-emacs-private/blob/6aedee9bb7a5624f1dd63fcd5b6534aad400101a/config.org#reformatter than to a comment in the issue don’t you think? Also is this something that can be installed now (ie an Emacs package or something similar) or is the idea to c/p the code?

Emiller88:
I was following what the ALE plugin did, which linked to the dicussion on the ALE issues. 

My hesitation with linking to my personal config, is if I decide to remove/move/etc, then it will break your link.

The package would be `reformatter`, which is a macro to create your own formatters. I'm considering a PR to [emacs-format-all-the-code](https://github.com/lassik/emacs-format-all-the-code), which would be a little less hassle for the end user. But I'll need to go read their package see how they do things differently than reformatter.

Copying and pasting is the usual emacs way. See the [reformatter installation](https://github.com/purcell/reformatter.el#installation) for reference. I could write quite a bit on it if you'd like, but it might bog down your clean README with emacs jargon that most Emacs users would be familar with the boilerplate.

lorenzwalthert:
Ok great you made the PR to 
https://github.com/lassik/emacs-format-all-the-code. So I think it the way of going about installing it from there is standard for emacs users, we won’t need a long description. You are right about the rear, maybe I have to restructure it anyway at some point and

lorenzwalthert:
Sorry closed by mistake -.- 

Ok great you made the PR to 
https://github.com/lassik/emacs-format-all-the-code. So I think if the way of going about installing it from there is standard for emacs users, we won’t need a long description. You are right about the README, maybe I have to restructure it anyway at some point and move some content to vignettes.

Emiller88:
@lorenzwalthert This has been merged https://github.com/lassik/emacs-format-all-the-code/pull/59

I also swapped it to the `emacs-format-all-the-code` link since it's a simple install of that now.

lorenzwalthert:
Thanks @Emiller88.

Issue #542, opened 2019-09-10
Support exclusion comments


michaelquinn32:
This was mentioned in separate issues, but would be useful to track separately.

There are many cases where it might be useful to disable styler for either a line or across multiple lines. This is supported in [jimhester/lintr](https://github.com/jimhester/lintr) with

* "# nolint" excludes the line
* "# nolint start" begins a block
* "# nolint end" ends a block

This is similar to functionality in google/yapf: 
https://github.com/google/yapf#why-does-yapf-destroy-my-awesome-formatting

lorenzwalthert:
This is a duplicate of https://github.com/r-lib/styler/issues/365 and https://github.com/r-lib/styler/issues/533. As noted in these issues, we are working on [detecting alignment](https://github.com/r-lib/styler/issues/258) to fix the most common case where styler falls short of the expectations.

michaelquinn32:
This should be reopened.

We are coming close to a point where calling styler can happen by default as a precommit across all of our packages. But, we are facing some problems with some jobs that parse R files, as they weren't written with styler's formatting conventions in mind.

Being able to turn styler off for these edge cases, and even better for just part of the code, would be really helpful.

lorenzwalthert:
As pointed out, I don’t see a lot of value in that. I‘d prefer to focus on concrete cases and try to improve formatting there. If you have any, please open a new issue. The formatter should be so good that there is no need for exceptions. Also, implementing this feature is non-trivial and likely to slow down styler and make it more complex - for limited gain. Instead of marking many lines with `nostyler`, why not making them compliant?

lorenzwalthert:
No let's open the original issue.

lorenzwalthert:
We have a draft :-). Feedback welcome.
```r
remotes::install_github("lorenzwalthert/styler@stylerignore")
```

lorenzwalthert:
Closed with #560.

Issue #533, opened 2019-07-10
`Ignore these lines`


russHyde:
Is there a way to tell `styler` not to style a subset of the lines in a file?
I couldn't find anything in the docs or the github issues.

lorenzwalthert:
Thanks for searching first :-) We decided against this in (agree hard to find) https://github.com/r-lib/styler/issues/365 for the reasons outlined there. Note that as far as alignment goes, there are considerations to detect certain patterns and don't apply styling, see #258.

lorenzwalthert:
If you want you can try the version of the WIP for detecting alignment, but no guarantees about future maintenance or further development. https://github.com/lorenzwalthert/styler/tree/detect-alignment. The goal is, however, to once finish this up and merge it into master. You can also subscribe to the other issue to make sure you don't miss anything.

lorenzwalthert:
@russHyde by popular request we think we might anyways implement this feature (although I don't recommend using it) and progress is tracked in #365 if you want to follow up on that.

russHyde:
Thanks. It was the introduction of the `{{` syntax in `rlang` (and the expectation that additional syntax might be added that styler wouldn't immediately handle; I appreciate that you've addressed `{{` in styler) that spurred me to ask about this. However, I've rethought my process and tried to remove 'weird new syntax' from my packages as much as possible ...

lorenzwalthert:
Ok. Yes, I suppose it would have been helpful for me to know that earlier, I could have pointed you to #528  :-). In almost any problem with styling, there is a general rule we can derive from it and which I am happy to implement (as it is obvious to me we need to handle `{{`). I'll make a CRAN submission soon as you might have seen in #556 so this should lower the barrieres to use styler again :smile: . In the meanwhile, you can just install from the master branch from GitHub.

lorenzwalthert:
We have a draft :-). Feedback welcome.
```r
remotes::install_github("lorenzwalthert/styler@stylerignore")
```

lorenzwalthert:
Closed with #560.

PR #560, opened 2019-10-22
Ignore certain lines when styling

### LINKED ISSUE #365: Conditional styling
It would be nice to be able to turn styling on and off for different sections of a file.

There are cases where the greedy behavior tries to "fix" intentional formatting. Consider this example, where many assignments are organized to emphasize the assignment value.

https://github.com/r-lib/pkgdown/blob/master/R/rd-html.R#L390

The `nocov start/nocov end` behavior of covr might be one approach. So styling of this:

```r

x <-  'style me'

# nostyle start
x <-       "intentional indent"
y <-       "another one"
# nostyle end
```

would generate:

```r
x <- "style me"

# nostyle start
x <-       "intentional indent"
y <-       "another one"
# nostyle end
```

lorenzwalthert:
Closes #365 as well as it's duplicates #533 and #542, following the https://github.com/r-lib/styler/issues/365#issuecomment-541216912. Here is a short example of how to use it, it works both for segments as well as inline:
```r
call(1 ) # styler: off
# styler: off
# also if there are more comments
test_xkj("hier", na.rm = 3, py = 43
         )

x="new" # styler: off
y=1 # none

more_calls(with(arguments))
# styler: on
1 + 1
a(!b)
```

Taken from the outline in #365, this PR implements the functionality as follows:

Here is how to make sure this code does not get styled: 

* Before parse data is nested, we find the stylerignore tags and add a column 
  stylerignore to the parse data which is false for all token except the 
  tokens that fall between two stylerignore tags.

* After code is turned into nested parse data and spaces and newlines are 
  computed (must be between terminals, not between any token as currently done with 
  `default_style_guilde_attributes()` **edit:** Because we need the position of tokens relative to other terminals, not tokens in general when serializing, we must do   `default_style_guilde_attributes()` twice in total now). We store the lag_spaces and 
  lag_newline attributes for all terminal tokens that are marked as 
  stylerignore in the environment env_current.
* We proceed as usual with appyling transformers and flattening out the parse 
  table. 
* Right before we serialize the parse table, we replace lag_newlines, lag_spaces
  and tokens, text with the initial values for all tokens that were marked as 
  stylerignore. Then we serialize.


R/stylerignore.R
```diff
@@ -0,0 +1,119 @@
+#' Add positional information of token to next terminal
```
lorenzwalthert: Say in documentation that this function modifies `env_current`.

PR #578, opened 2020-01-19
Caching top level expressions

### LINKED ISSUE #570: Cache on expression level
Smart caching would be done by top level expressions, i.e. cache every top level expression separately (but process all non-cached top-level expressions as a block since we need to run all styling separately on them otherwise).  The current implementation only helps when styling files that are already styled, e.g. it will make `style_pkg()` considerably faster if most files are already styled previously. It will also make the styler pre-commit hook run faster. However, when changing one line in a big file that was styled previously, the cache for all expressions is invalidated and there is no benefit from caching. 

There is stale WIP at https://github.com/lorenzwalthert/styler/tree/caching-top-level-expr.

lorenzwalthert:
The cache is currently invalidated if any expression is modified. If we manage to cache top level expressions separately, we can still use the cache for all non-modified expressions, which would should improve speed very often. Closes #570. 

- [x] Do only add variables block and is_cached (if at all) after than
  get_parse_data (but before parse_transform_serialize_r_block because when we
  do block processing this might be too late) to not drag along
  transformers everywhere totally unrelated. When processing blocks, make sure
  to cache individual values where they are created.

- [x] process in block, not every expression.

- [x] expressions on the same line must be processed together, i.e. when separated
  by semi-colon or comment, they belong to the same block.

- [x] think about the interaction with create_tokens and that each column should
  potentially exist on every level, not just top level.

- [x] the text attibute of a top level nest is not used to create a hash for
  caching, but is used to return text in case no styling is needed because the
  value is cached.

- [x] resolve line break problem.

- [x] Add tests. Both low-level and high-level.

- [x] `compute_parse_data_nested()` should not compute the parse data nested for expressions that are cached already because for them, we just return `text` of the top level expression.

- [x] Put all consecutive non-cached expressions into one block to speed up things. Maybe this does not help as much as expected. Need more comprehensive benchmarks.

- [x] ~nice to have: Cache after styling each expression separately, so styling a
  file with ten identical expressions will use cache for 9 of them.~ Does not seem to work because we must have blocks first.


R/initialize.R
```diff
@@ -82,6 +84,10 @@ initialize_indent <- function(pd_flat) {
   pd_flat
 }
 
+
+
+
+
```
rillig: What's the semantic difference between 1 empty line, 3 empty lines (as in line 22) and 5 empty lines?

R/nest.R
```diff
@@ -7,23 +7,112 @@
...
+  } else {
+    pd_nested$block <- rep(1, nrow(pd_nested))
+  }
+  pd_nested
+}
+
+#' Drop all children of a top level expression that are cached
+#'
+#' Note that we do cache top-level comments. Because package code has a lot of
+#' roxygen comments and each of them is a top level expresion, so checking is
```
rillig: The word "so" is wrong since the sentence starts with "Because".

R/nest.R
```diff
@@ -7,23 +7,112 @@
...
+#' it's on a line where multiple expressions sit and at least one is not styled
+#' (exception).
+#'
+#' **usual case: All other expressions in a block are cached**
+#'
+#' Cached expressiond don't need to be transformed with `transformers` in
+#' [parse_transform_serialize_r_block()], we simply return `text` for the top
+#' level token. For that
+#' reason, the nested parse table can, at the rows where these expressions are
+#' located, be shallow, i.e. it does not have to contain a children, because it
```
rillig: typo: "a children" -> "children"

R/nest.R
```diff
@@ -7,23 +7,112 @@
...
+#' will neighter be transformerd nor serialized anytime. This function drop all
```
rillig: typo: "neighter" -> "neither"
typo: "transformerd" -> "transformed"
typo: "drop" -> "drops"

R/transform-block.R
```diff
@@ -0,0 +1,97 @@
...
+  }
+  c(rep("", start_line - 1), serialized_transformed_text)
+}
+
+#' Find the groups of expressions that should be processed together
+#'
+#' Every expression is an expression itself, Expressions on same line are in
+#' same block.
+#' Multiple expressions can sit on one row, e.g. in line comment and commands
+#' seperated with ";". This creates a problem when processing each expression
```
rillig: typo: "seperated" -> "separated"

R/transform-block.R
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#' Number of lines between cache blocks
+#'
+#' This is relevant when putting expressions together into a block and preserve
+#' blank lines between them.
+#' @param pd A top level nest.
+find_blank_lines_to_next_block <- function(pd) {
+  block_boundary <- pd$block != lag(pd$block, default = 0)
+  # TODO everywhere: block is not ambiguous. use cache block since we also have
```
rillig: typo: "not ambiguous" -> "ambiguous"

R/utils-cache.R
```diff
@@ -13,11 +13,32 @@ hash_standardize <- function(text) {
     list()
 }
 
+#' Check if text is cached
+#'
+#' This boilds down to check if the hash exists at the caching dir as a file.
```
rillig: typo: "boilds" -> "boils"

lorenzwalthert:
Thanks @rillig, I will have a look.

PR #583, opened 2020-02-02
Change in R parser makes styler fail on R devel

### LINKED ISSUE #582: R devel operator precedence fix makes styler fail on CRAN
From R core:

> the R parser changed in 77722 (PR#16710), fixing the precedence of "?" with respect to "=" to be as documented (lower) and as it seems it was until 2001. The change affects these packages, which will start failing their tests now [...]

The diff is here: https://github.com/wch/r-source/commit/84bbf385f909c0223924c310af6c7c77aa810234

The change is that the token `eq_assign` was renamed to `expr_or_assign_or_help` in certain situations:

```r
# R 3.6
> getParseData(parse(text = "a = 2"), includeText = TRUE)
  line1 col1 line2 col2 id parent        token terminal  text
8     1    1     1    5  8      0 equal_assign    FALSE a = 2
1     1    1     1    1  1      3       SYMBOL     TRUE     a
3     1    1     1    1  3      8         expr    FALSE     a
2     1    3     1    3  2      8    EQ_ASSIGN     TRUE     =
4     1    5     1    5  4      5    NUM_CONST     TRUE     2
5     1    5     1    5  5      8         expr    FALSE     2
```

```r
# R devel
  line1 col1 line2 col2 id parent                  token terminal text
9     1    1     1    5  9      0 expr_or_assign_or_help    FALSE     
1     1    1     1    1  1      3                 SYMBOL     TRUE    a
3     1    1     1    1  3      9                   expr    FALSE     
2     1    3     1    3  2      9              EQ_ASSIGN     TRUE    =
5     1    5     1    5  5      6              NUM_CONST     TRUE    2
6     1    5     1    5  6      9                   expr    FALSE     
```

This means `parser_version_find()` is fooled into believing that the code origins from parser version 1, but it comes from 2 (or should we even name it 3?)

This will change the structure of the parse data because the equal assign will be relocated, adding another intermediate *nest*.

Solution: set the parser version to 3 and document it.

lorenzwalthert:
Closes #582.

### DIFF
--- a/R/environments.R
+++ b/R/environments.R
@@ -15,6 +15,7 @@
 #'
 #' * version 1: Before fix mentioned in #419.
 #' * version 2: After #419.
+#' # version 3: After #582.
 #'
 #' The following utilities are available:
 #'
@@ -42,7 +43,13 @@ parser_version_get <- function() {
 
 #' @rdname parser_version_set
 parser_version_find <- function(pd) {
-  ifelse(any(pd$token == "equal_assign"), 2, 1)
+  ifelse(any(pd$token == "equal_assign"),
+    2,
+    ifelse(any(pd$token == "expr_or_assign_or_help"),
+      3,
+      1
+    )
+  )
 }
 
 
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
I thought this one should fail but it does not: https://travis-ci.org/lorenzwalthert/styler/builds/645177749?utm_medium=notification&utm_source=github_status 

Is R-devel outdated?

PR #589, opened 2020-02-09
make R.cache an import

lorenzwalthert:
Make sure that when cache is was deactivated by the user and R.cache is installed but there is no R.cache permanent directory, the user is never asked about setting up a permanent directory.


lorenzwalthert:
Also relevant link: #578.

Issue #371, opened 2018-03-13
Add newline after multiline function declaration


krlmlr:
```r
# Good
f1 <- function(a, b) {
  code
}

f2 <- function(a, b,
               c) {

  code
}

# Bad
f3 <- function() {
  code
}
```

lorenzwalthert:
Is there a corresponding rule in the tidyverse style or do you plan to add one?

krlmlr:
I remember Hadley suggested that an empty line after a long function header helps readability. I wouldn't formalize this as a rule yet, but I've consistently seen this pattern in dplyr.

lorenzwalthert:
Ok. I think this is easy to implement. I just think for the sake of consistency, we should in general aim to keep the style guide in sync with styler.

lorenzwalthert:
For a propper implementation, we'd need #357 because we want to add a line break dependent on whether a token is multi-line, which we cannot do if the multi-line attribute is not up-to date. The multi-line attribute is not up-to-date because it is only updated *after* all line break rules were applied.

lorenzwalthert:
Blocked by #357.

lorenzwalthert:
Implemented in mlr branch: https://github.com/lorenzwalthert/styler/commit/2a3fbc6fe278fbf9a09d229a49379d778536b5ed

lorenzwalthert:
The implementation in the mlr branch now depends on the number of lines in the body.

lorenzwalthert:
> For a propper implementation, we'd need #357 because we want to add a line break dependent on whether a token is multi-line, which we cannot do if the multi-line attribute is not up-to date. The multi-line attribute is not up-to-date because it is only updated after all line break rules were applied.

Wrong. We need the exact line count, which we can compute ad-hoc within the transformer. It's as simple as in https://github.com/lorenzwalthert/styler/commit/d9f1b22f52437368446daf2c4c654481bf914784.

PR #630, opened 2020-04-11
No blank lines in function headers

### LINKED ISSUE #549: Redundant line breaks in function header not removed when formatting
After resolving #541, `tryCatch` blocks now terminate in a new line.

```
code <- "tryCatch({
  my_options <- options()
  print('hello!')
},
error = function(e) e)
"
styler::style_text(code)
#> tryCatch(
#>   {
#>    my_options <- options()
#>    print("hello!")
#>  },
#>  error = function(e) e
#> )
```

This seems like one good case from #125, but there are examples in that issue where the original formatting was considered correct.

Similarly, function signatures that used a wrapped style for arguments also end in an empty new line.

```
code <- "function(
    my_indendented_arg, and_another) {
  print('hello')
}"
styler::style_text(code)
#> function(
#>          my_indendented_arg, and_another) {
#>   print("hello")
#> }
```

These seem to be related as cases where there are empty new lines after an opening paren.

Thanks!

lorenzwalthert:
Closes #549.


krlmlr:
This might be too strict.

## With this PR

``` r
styler::style_text("c(\n  a,\n\n  b\n)")
#> c(
#>   a,
#>   b
#> )
```

<sup>Created on 2020-04-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

## CRAN version

``` r
styler::style_text("c(\n  a,\n\n  b\n)")
#> c(
#>   a,
#> 
#>   b
#> )
```

<sup>Created on 2020-04-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lorenzwalthert:
This behavior was introduced in #629. I think it’s exactly as intended. When and why would you prefer the blank line not to be removed?

krlmlr:
The empty line allows for better structuring of the code in my case, see https://github.com/krlmlr/dm/commit/155570beda515a67ceb58106207d9ef1a3ae3901. I'm really looking for

Screenshot from 2020-04-12 08-10-40

krlmlr:
Adding an empty comment as a placeholder works but adds noise. Can we be more liberal with empty lines, e.g. only remove if `strict = TRUE`?

lorenzwalthert:
It is only for `strict = FALSE`: 
``` r
styler::style_text(
  "c(
  1, 
  
  2
  )
  ",
  strict = FALSE
)
#> c(
#>   1,
#> 
#>   2
#> )
```

<sup>Created on 2020-04-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

But I wonder if there are other ways to accommodate for your use case. Can you specify conditions under which you want to allow blank lines, e.g.

- function calls with more than *k* lines.
- arguments must be named 
- arguments must be multi-line.

etc.?

krlmlr:
Indeed, missed that. I also like how the empty line is kept with `strict = FALSE` at the beginning of function code.

``` r
styler::style_text('
f <- function(a = 1,
              b = 2) {

  here("we should keep", "that empty line above")
}', strict = FALSE)
#> 
#> f <- function(a = 1,
#>               b = 2) {
#> 
#>   here("we should keep", "that empty line above")
#> }
```

<sup>Created on 2020-04-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Maybe if there's a comment, we should keep all empty lines around that comment?

lorenzwalthert:
yeah but empty lines after function header is another topic, let's discuss that in #371. Also, there is no intention to remove this empty line, rather in #371 we may even add one for longer function declarations.

> Can you specify conditions under which you want to allow blank lines, e.g.
> - function calls with more than k lines.
> - arguments must be named
> - arguments must be multi-line.
> - only for functions like ifelse(), switch() and friends.
> ?

Any insights on that from your side? While it is ok to defer to `strict = FALSE`, this also has a lot of other implications one does not want because I think `strict` is overloaded.

krlmlr:
Maybe if there's a comment, we should keep all empty lines around that comment?

lorenzwalthert:
You mean to preserve the blank lines in 
```r
utils::globalVariables(c(
   # pipe
   ".",
  
   # nesting
   "data",
   ...
))
```
for example?

This sounds reasonable.

krlmlr:
I'd rather discuss `strict` elsewhere. I think #549 ultimately was about function headers, this PR also changes formatting of long function calls. Can we disentangle?

If in doubt, user code is correct and should be kept unchanged, also with `strict = TRUE`.

My huge `switch()` in https://github.com/krlmlr/dm/commit/155570beda515a67ceb58106207d9ef1a3ae3901 should be rewritten, let's not bother here.

lorenzwalthert:
> I'd rather discuss strict elsewhere. I think #549 ultimately was about function headers, this PR also changes formatting of long function calls. Can we disentangle?

Well the problem was I forgot to update one test in the other PR and then I rebased after merge but for some reason the commit was still part of this PR so sorry for the convolution. I think function headers are not a concern so much and after resolving #630 we can focus on function calls in #632.

PR #634, opened 2020-04-17
Allow for dry run

### LINKED ISSUE #633: Dry-run mode
Is it possible to run `styler` in dry-run/check-only mode, where it simply reports what would be formatted?

Our use-case for this is to use `styler` in a pre-commit hook on the developer's side, and as part of CI/CD tests once code has been received by git server.

Technically we could solve this by using temporary directories and a lot of copying and diff-ing. However it would be much simpler if `style_dir` accepted parameter `check_only=T` or `dry_run=T`.

We would then use the output of this function inside `stopifnot` for example, such that our scripts will fail, thereby halting the commit or failing the test run.

Python's code formatter `black` has this functionality: by running `black --check`, nothing gets modified and it will exit with non-zero exit code if something would have been formatted. This makes it very easy to use in pre-commit hooks and CI/CD tests.

lorenzwalthert:
Closes #633 and implements an API change.

New argument `dry` in `style_file()` and friends to control write back and failing in case of modification. The argument takes these values:

- `off` (default): Current behavior. Write back to file as side effect. Return tibble with summary invisible (as now).
- `on`: Never modify files, also return tibble summary **as if** things were written back.
- `fail`: Same as `on`, but fail fast with error if modification would occur on write-back.

Any feedback on the design welcome, I am not so sure this is optimal. @krlmlr since this is an API change, I'd like to have your opinion on that if possible.


lorenzwalthert:
@select-id-from-users you can try this if you want. Will merge into master later.

select-id-from-users:
@lorenzwalthert Works like a charm, thank you very much!

```
tux:/tmp/renv$ Rscript -e 'styler::style_file("test.R", dry="fail")'
Styling  1  files:
 test.R Error: File `test.R` would be modified by styler and `dry` = 'fail'.
Backtrace:
     █
  1. └─styler::style_file("test.R", dry = "fail")
  2.   └─styler:::transform_files(...)
  3.     └─purrr::map_lgl(...)
  4.       └─styler:::.f(.x[[i]], ...)
  5.         └─styler:::transform_code(path, fun = fun, ..., dry = dry)
  6.           └─styler:::transform_utf8(path, fun = fun, ..., dry = dry)
  7.             ├─map_lgl(path, transform_utf8_one, fun = fun, dry = dry) %>% set_names(path)
  8.             │ └─base::eval(lhs, parent, parent)
  9.             │   └─base::eval(lhs, parent, parent)
 10.             └─purrr::map_lgl(path, transform_utf8_one, fun = fun, dry = dry)
 11.               └─styler:::.f(.x[[i]], ...)
 12.                 └─rlang::with_handlers(...)
 13.                   └─bas
Execution halted
tux:/tmp/renv$ echo $?
1
tux:/tmp/renv$ Rscript -e 'styler::style_file("test.R", dry="off")'
Styling  1  files:
 test.R ℹ 
────────────────────────────────────────
Status	Count	Legend 
✔ 	0	File unchanged.
ℹ 	1	File changed.
✖ 	0	Styling threw an error.
────────────────────────────────────────
Please review the changes carefully!
tux:/tmp/renv$ Rscript -e 'styler::style_file("test.R", dry="fail")'
Styling  1  files:
 test.R ✔ 
────────────────────────────────────────
Status	Count	Legend 
✔ 	1	File unchanged.
ℹ 	0	File changed.
✖ 	0	Styling threw an error.
────────────────────────────────────────
tux:/tmp/renv$ echo $?
0

```

lorenzwalthert:
> Error: File `test.R` would be modified by styler and `dry` = 'fail'.

Maybe we can come up with a better error:

```
Error: File `test.R` would be modified by styler and argument `dry` is set to 'fail'.
```

Also, do you find on, off and dry self-explanatory?

select-id-from-users:
I guess since the error is thrown only when dry="fail" you don't have to mention that dry is set to 'fail'? I have no strong opinion here, and would be happy either way.

I find the options self-explanatory yes. Maybe "on" could be "warn"? Although warnings can also result in non-zero exit code when you run `devtools::check(error_on="note")`. Also here I don't have strong opinion. Naming things is hard.

Issue #637, opened 2020-04-23
Add an option for suppressing all communication


michaelquinn32:
Hi Lorenz!

We're using styler as a part of some tools that read from stdin and write to stdout. In that context it's very important that all messages are turned off, and that styler only output the styled code.

We've made some small modifications to the package to support this, and added an option for enabling and disabling the feature. Is this something that you would like added to the main package?

Thanks!

P.S. And if this came up before, apologies.

lorenzwalthert:
Sure, I'd like to see that merged into upstream. Ther's already an issue for that: #375. We can discuss it further there if you like. Also note #634, a related API change not yet merged into master for dry-running styler.

michaelquinn32:
The current implementation (which can change here of course), uses an option (styler.communicate)  and doesn't require an API change. It adds some conditional logic around the points of the code that produces messages, suppressing them when enabled.

lorenzwalthert:
Ok. The advantage is probably that we don't have to pass the argument down to the innermost stacks that way. And I think usethis also has a `{packagename}.communicate` option, right? Then we should also use a subset or the same arguments this option has for consistency in r-lib/tidyverse probably.

lorenzwalthert:
Closed with #640.

PR #640, opened 2020-05-07
Add option for suppressing styler communication

michaelquinn32:
This implements FR #637


michaelquinn32:
Hi Lorenz!

Thanks for the patience. I finally got around to this. One thing. I looked a little bit, but couldn't find if or where styler's communication was tested. Mind sharing a code pointer?

Thanks!

lorenzwalthert:
Thanks. Two things to debate: 

- You implemented as a boolean option. This only allows two modes. Do we ever want to support more than two modes? I.e. moderately verbose?
- In {usethis}, the option is called `usethis.quiet`. I don't know if the tidyverse has embraced standardisation for options that manage communication. Maybe we should investigate that.

Tests for communication are here: https://github.com/r-lib/styler/blob/master/tests/testthat/test-public_api.R

michaelquinn32:
I can't think of a situation where "less verbose" makes sense. For my use case, either the messaging is useful (which it almost always is), or it is interfering with something like stdout.

I'm happy to update the flag name. Alignment with other packages is good.

NEWS.md
```diff
@@ -6,33 +6,35 @@
   when there are no comments before or after the blank line (#629, #630, #635).
 - `style_file()` and friends gain argument `dry` to control if changes should 
   be applied to files or not (#634).
+- added an option for disabling all communication when using the package
+  (styler.quiet) (#640).
```
lorenzwalthert, line 10:
```suggestion
  (`styler.quiet`) (#640).
```

lorenzwalthert:
How come that a significant part of `NEWS.md` got reformatted in 32f659c85cd8f812a9d616294cb46e33f2981de7?

michaelquinn32:
Because I have an autoformatter that needs to be manually disabled :man_facepalming:.

Sorry, it's fixed.

lorenzwalthert:
Out of curiosity: Which formatter do you use? Could we add that to {precommit} as a hook?

lorenzwalthert:
Thanks @michaelquinn32.

michaelquinn32:
A precommit would be great!

I'm using an internal formatter, but I don't think it matters all that much. prettier, for example, supports markdown: https://github.com/prettier/prettier

lorenzwalthert:
They already made a pre-commit compatible hooks themself: https://prettier.io/docs/en/precommit.html#option-3-pre-commithttpsgithubcompre-commitpre-commit

PR #643, opened 2020-06-01
Use tokens instead of token in doc

renkun-ken:


lorenzwalthert:
Thanks. You forgot to roxygenize :smile:. Can you do this and push again? We' are not (yet) using gh actions. Also, to prevent in the future, I'll add a note to CONTRIBUTING.md on how to use pre-commit as a contributor.

renkun-ken:
Actually I tried but the following error occurs:

```
> roxygen2::roxygenize()                                                                                                                                                              
Error in loadNamespace(name) : there is no package called 'pkgapi'
Backtrace:
1: stop(cond)
2: doWithOneRestart(return(expr), restart)
3: withOneRestart(expr, restarts[[1L]])
4: withRestarts(stop(cond), retry_loadNamespace = function() NULL)
5: loadNamespace(name)
6: getNamespace(ns)
7: asNamespace(ns)
8: getExportedValue(pkg, name)
9: pkgapi::api_roclet
10: eval(expr, env)
11: eval(expr, env)
12: FUN(X[[i]], ...)
13: lapply(roclets, roclet_find)
14: roxygen2::roxygenize()

> install.packages("pkgapi")                                                                                                                                                          
Installing package into '/home/renkun/R/x86_64-pc-linux-gnu-library/4.0'
(as 'lib' is unspecified)
Warning message:
package 'pkgapi' is not available (for R version 4.0.0) 
```

I guess I'm missing something?

lorenzwalthert:
Jup, it's not yet on CRAN: https://github.com/r-lib/pkgapi.

renkun-ken:
Thanks! Just roxygenized.

lorenzwalthert:
Also, CI fails seem unrelated. Any idea? I opened https://github.com/r-lib/backports/issues/35 so we could import that function from {backports} maybe, but not sure it works that way.

PR #647, opened 2020-06-22
Invalidate cache on style guide argument change

### LINKED ISSUE #646: Changing styling options has no effect unless file is changed
This is a bit minor, but essentially: after running `style_*(<some file etc>)`, then running it again with a new style, e.g. `style_*(<some file etc>, transformers = tidyverse_style(indent_by = 4))` nothing will happen. If the file is edited in any way, then the second command will work correctly. 

I believe this is a cache issue, but could be wrong.

The issue is that if you run styler on a file with the wrong settings, every function must be touched by an edit to "break" the cache.

## Example

Say we had an example file `test.R` that only had the following:
```R
test_func <- function(x) {

   x + 2
}
```
(with three spaces before the `x` on the third line)

Then running `style_file("test.R")` will give (as expected):

```R
test_func <- function(x) {
  x + 2
}
```

Now we decide we want four spaces. We should run (I think) `style_file("test.R", transformers = tidyverse_style(indent_by = 4))`, but in this case nothing happens (`✔ 	0	File unchanged.`).

Then, if any edit is made to the file, even a change unrelated to indenting, the above command *does* work and correctly changes the file to:
```R
test_func <- function(x) {
    x + 2
}
```

NOTE: in testing, this "fix" **only worked for the function edited, not the whole file**. So I had to make edits to every function in the desired file to get the correct result.

## Best Guess

I guess this is a rule about the caching: if the function is unchanged after the previous `styler` run, it is reasonable to assume that no styling should be done, but of course this isn't the case if the style rules have changed. 

## Version and System

I got this on the latest CRAN release, 1.3.2. I got the same behaviour (with less testing) on the current dev version on github, `9c5d88328294d0c7a32818550bc99f6526f25b0d`.

MacOS Catalina 10.15.5

## Thanks

Thanks for the package, very useful! My apologies if the above issue is simply just misuse on my part.

lorenzwalthert:
Closes #646 by passing these arguments down to the key generation. Unsplicing with `!!` is not sufficient because `as.character(list(purrr::partial(sum, x = 2)))` and similar returns generic code that does not expose the value `x = 2`.


lorenzwalthert:
Won't fix https://github.com/r-lib/roxygen2/issues/1108, will be resolved later.

Issue #669, opened 2020-09-18
make use of new selection-based API in next release of rstudioapi?


kevinushey:
The next version of RStudio will come with support for a visual editing mode, as described in https://rstudio.github.io/visual-markdown-editing/#/. Unfortunately, many of the pre-existing `rstudioapi` APIs are not compatible with visual editing mode, since they assume a line-based editor, where cursor positions can be referenced globally via [row, column] offsets. The new visual editor is cell-based and so does not map to the same abstraction as well.

To help alleviate this, we are introducing a few new APIs that addin authors can use to ensure their addins function both in visual and source mode. Would you be receptive to a PR making that possible?

lorenzwalthert:
Sure, if you can make a PR that would be great.

kevinushey:
No longer needed; the APIs used by `styler` should "just work" in visual mode with the latest daily builds of RStudio (v1.4.890+).

lorenzwalthert:
@kevinushey it seems the styler addin does not work well with the visual mode as reported in #740. Can you have a look? Are we missing a minimal version requirement for `{rstudioapi}` or something like that?

kevinushey:
Can you please file a separate issue that makes the actual problem more clear? Particularly, that switching from Visual mode to Source mode seems to mangle the editor undo stack in some cases (with a reproducible example)?

jthomasmock:
Hi @kevinushey I originally opened the issue in `{styler}`, and have moved it to here at Lorenz' request.

https://github.com/r-lib/styler/issues/740#issue-817976625

I have copied the bulk of that post below. I can reproduce interactively, and have included a GIF of the behavior.

---

Problem: In Visual Markdown mode, styling the active selection in a code-chunk appears to correctly style the code, but malforms it on the plain-text side.

Switching back into plain-text mode, the code will be "fixed", but if you "undo" you can see there is an intermediate format that is completely malformed and has lost some characters. Undo a second time will revert back to the original, unstyled format.

Reprex below, and note this works fine in non-visual editor mode.

````
---
output: html_document
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = TRUE)
library(tidyverse)
```



Use the `styler` package to "style" the code, intentionally awkwardly formatted and of sufficient length as an example.

```{r}
mpg %>%     filter(cyl   %in%c(4 , 6,     8))%>%slice_sample(n=30) %>% mutate(hwy_mpl = hwy/2.352)
```
````

Environment:
```
[...]

PR #631, opened 2020-04-11
Depreciate env var save_after_styling

lorenzwalthert:
Closes #555. But wait until decided if default should no longer be false, but depend on rstudio < 1.3, because 1.3. introduces autosaving (only when curser is away from the editor).

PR #670, opened 2020-09-18
allow style_selection() to work in RStudio's visual mode

kevinushey:
Closes https://github.com/r-lib/styler/issues/669.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,6 +26,7 @@
 - overhaul pgkdown site: Add search (#623), group function in Reference (#625).
 - always strip trailing spaces and make cache insensitive to it (#626).
 - typos in documentation (#643, #618, #614).
+- `style_selection()` now uses RStudio's new selection-based API when available
 
 # styler 1.3.2
 
--- a/R/addins.R
+++ b/R/addins.R
@@ -91,6 +91,13 @@ style_active_pkg <- function() {
 #' @importFrom rlang abort
 #' @keywords internal
 style_selection <- function() {
+  if (rstudio_has_selection_apis())
+    style_selection_v2()
+  else
+    style_selection_v1()
+}
+
+style_selection_v1 <- function() {
   communicate_addins_style_transformers()
   context <- get_rstudio_context()
   text <- context$selection[[1]]$text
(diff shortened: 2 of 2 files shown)


NEWS.md
```diff
@@ -26,6 +26,7 @@
 - overhaul pgkdown site: Add search (#623), group function in Reference (#625).
 - always strip trailing spaces and make cache insensitive to it (#626).
 - typos in documentation (#643, #618, #614).
+- `style_selection()` now uses RStudio's new selection-based API when available
```
lorenzwalthert, line 29:
```suggestion
- `style_selection()` now uses RStudio's new selection-based API when available (#670).
```

R/addins.R
```diff
@@ -105,6 +112,31 @@ style_selection <- function() {
...
+  if (Sys.getenv("save_after_styling") == TRUE &&
+      !is.null(rstudioapi::documentPath(id))) {
+    invisible(rstudioapi::documentSave(id))
+  }
+
+}
+
+rstudio_has_selection_apis <- function() {
+
+  rstudioapi::hasFun("selectionGet") &&
```
lorenzwalthert, line 133: I am afraid this is a bit slow (given that styler is already quite slow). Is making it RStudio version dependent an option?

R/addins.R
```diff
@@ -91,6 +91,13 @@ style_active_pkg <- function() {
 #' @importFrom rlang abort
 #' @keywords internal
 style_selection <- function() {
+  if (rstudio_has_selection_apis())
+    style_selection_v2()
+  else
+    style_selection_v1()
+}
+
+style_selection_v1 <- function() {
```
lorenzwalthert, line 100: Why v1 and v2? Does this refer to the `rstudioapi` major version?

R/addins.R
```diff
@@ -105,6 +112,31 @@ style_selection <- function() {
...
 }
 
+style_selection_v2 <- function() {
+  communicate_addins_style_transformers()
+  id <- rstudioapi::documentId()
+  selection <- rstudioapi::selectionGet(id = id)
+  text <- selection$value
+  if (!nzchar(text)) abort("No code selected")
+  out <- style_text(text, transformers = get_addins_style_transformer())
+  rstudioapi::selectionSet(value = out, id = id)
```
lorenzwalthert, line 122: What we did above with `if (context$selection[[1]]$range$end[2] == 1) "")` guarantees that when your selection is 
Screenshot 2020-09-18 at 23 44 40

There is exactly one line break after the selected area (and no blank line), and if it is 
Screenshot 2020-09-18 at 23 45 06
there is also a line break after the selection and no blank line. Does your implementation guarantee this as well?

R/addins.R
```diff
@@ -105,6 +112,31 @@ style_selection <- function() {
...
+
+  if (Sys.getenv("save_after_styling") == TRUE &&
```
lorenzwalthert, line 124: Note to self: when merged, update #631.

lorenzwalthert (changes requested):
Thanks @kevinushey, I really appreciate RStudios proactive contribution. Below are a few questions. Also, these addin functions are currently mostly tested manually. If there are any ways to automate some of them, I'd be glad to hear about it (I only thought about mocking `get_rstudio_context()` so far), but nothing implemented. Maybe I can implement them in another PR so we can make sure your PR passes these tests. There are a few tricky cases with line breaks (last line blank, not blank, screenshots below)and selections that we need to get right. There are some instructions in `tests/testmanual/tests` you could try to run for now.

R/addins.R
kevinushey: Yes, that would also be suitable (`rstudioapi >= 0.12`; `RStudio >= 1.4.845`). Or, this condition could be cached and reused.

R/addins.R
kevinushey: Just arbitrary versions to indicate one is "newer" and one is "older".

R/addins.R
kevinushey: It doesn't; the new API doesn't provide any mechanism for expanding on modifying the user's current selection currently. We should probably discuss this on https://github.com/rstudio/rstudio/issues/7465.

kevinushey:
We've made some changes to the development version of RStudio that should allow `styler` to continue working as-is, without any changes required from your side. Sorry for the trouble!

lorenzwalthert:
Ok great. No worries.

PR #674, opened 2020-09-26
Add benchmarking infra

lorenzwalthert:
Using `bench::mark()` and friends, we build a continuous benchmarking system. For any PR there is, we create a boxplot to visualise the distribution for timings for the base branch (e.g. master) and the branch we want to merge into the base branch (head) in one plot.

Todo: 

- [x] Use more representative code (also analysis, Rmd etc.). We could even store this code in the styler repo, just not under`bench/` directly. We should probably run two styler commands - Some files should be partly styled to test caching of top level expressions. 
- [x] Use repo we have full control over instead of public r-lib repo.
- [x] maybe create versioning for the benchmarked code, so we can later switch and know what we are comparing. This can be done with the `name` arugment in `bench::mark()`.

### DIFF
--- a/bench/01-declarations.R
+++ b/bench/01-declarations.R
@@ -7,25 +7,30 @@ plot_against_base <- function(new_bm,
                               new_bm_label = deparse(substitute(new_bm)),
                               file = paste0("plots/", new_bm_label, ".pdf")) {
   new_bm <- bench::as_bench_mark(new_bm)
+  name <- unique(new_bm$name)
+  stopifnot(length(name) == 1)
   branches <- gert::git_branch_list()
   last_commit_base_branch <- branches[branches$name == Sys.getenv("GITHUB_BASE_REF"), "commit", drop = TRUE]
   bm <- bench::cb_read()
   commit_is_reference <- bm$commit_hash == last_commit_base_branch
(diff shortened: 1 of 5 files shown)


bench/01-declarations.R
```diff
@@ -0,0 +1,32 @@
...
+#'
+#' @param new_bm A new benchmark object.
+#' @param new_bm_label The label of the new benchmark used as a title.
+#' @param file The file path to write the plot.
+plot_against_master <- function(new_bm,
+                                new_bm_label = deparse(substitute(new_bm)),
+                                file = paste0("plots/", new_bm_label, ".pdf")) {
+  new_bm <- bench::as_bench_mark(new_bm)
+  branches <- gert::git_branch_list()
+  master <- branches[branches$name == "master", "commit", drop = TRUE]
```
lorenzwalthert, line 12: make this more generic for target branch != master.

bench/01-declarations.R
```diff
@@ -0,0 +1,32 @@
+#' Plot benchmarks against master, if benchmarks there exist
+#'
+#'
+#' @param new_bm A new benchmark object.
+#' @param new_bm_label The label of the new benchmark used as a title.
+#' @param file The file path to write the plot.
+plot_against_master <- function(new_bm,
```
lorenzwalthert, line 7: more generic, not just master (also in title).

PR #687, opened 2020-10-26
Fixes #686

### LINKED ISSUE #686: vignette "Customizing styler": mangled text in "Showcasing the development of a styling rule" section
The block quoted output of `styler:::create_tree(code)` is a mess ;-) when rendered by pkgdown in the pkg website.

Screen Shot 2020-10-26 at 14 43 21

espinielli:


lorenzwalthert:
Thanks @espinielli. If you check the devel doc page of styler at https://styler.r-lib.org/dev/, you'd actually see that the problem is already fixed there, but not so nicely (just c/p output, not evaluating `create_tree()`, so I agree we should use this PR. Sidenote: create_tree() currently only works with caching deactivated (which happens to be the case in this vignette), so I created https://github.com/r-lib/styler/pull/688.

lorenzwalthert:
actually this does not work because data.tree is required to run this function and it does not exist on R < 3.5.

PR #693, opened 2020-11-11
use is.character over inherits(character)

MichaelChirico:
Ran into this issue:

```
file.create(tmp <- tempfile())
l = xfun::file_string(tmp)
is.character(l)
# [1] TRUE
inherits(l, 'character')
# [2] FALSE
styler::style_text(l)
# Error in construct_vertical(styled_text) : 
#   inherits(x, what = c("utf8", "character", "vertical")) is not TRUE
```

After this PR:

```
styler::style_text(l)

Warning message:
Text to style did not contain any tokens. Returning empty string. 
```

I also filed this with `xfun`: https://github.com/yihui/xfun/pull/41 but I do think `is.character` is probably preferable anyway.

LMK if there's any test to add

### DIFF
--- a/R/vertical.R
+++ b/R/vertical.R
@@ -5,7 +5,7 @@
 #' @param x A character vector or an object of class "vertical".
 #' @keywords internal
 construct_vertical <- function(x) {
-  stopifnot(inherits(x, what = c("utf8", "character", "vertical")))
+  stopifnot(is.character(x) || inherits(x, what = c("utf8", "vertical")))
   structure(x, class = "vertical")
 }
 
--- a/R/vertical.R
+++ b/R/vertical.R
@@ -5,7 +5,7 @@
 #' @param x A character vector or an object of class "vertical".
 #' @keywords internal
 construct_vertical <- function(x) {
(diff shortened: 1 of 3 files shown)


lorenzwalthert:
Hi @MichaelChirico. Thanks for reporting.

R/vertical.R
```diff
@@ -5,7 +5,7 @@
 #' @param x A character vector or an object of class "vertical".
 #' @keywords internal
 construct_vertical <- function(x) {
-  stopifnot(inherits(x, what = c("utf8", "character", "vertical")))
+  stopifnot(is.character(x) || inherits(x, what = c("utf8", "vertical")))
```
lorenzwalthert, line 8:
```suggestion
  stopifnot(is.character(x))
```
Can we further simplify like this? Because 
```r
all(
  is.character(glue::glue("1")),
  is.character(styler:::construct_vertical("1")),
  is.character(enc::as_utf8('3'))
)
#> TRUE
```

MichaelChirico: Works for me

lorenzwalthert:
Thanks @MichaelChirico.

PR #729, opened 2021-02-05
Fix (hopefully) all roxygen parsing problems for dont* sequences and escape characters

### LINKED ISSUE #728: avoid turning `\n` into `\\n` in roxygen docs
As discussed [here](https://github.com/easystats/insight/pull/297#issuecomment-773340800):

As you mentioned, the issue is `\dontrun`:

```r
#' Data frame and Tables Pretty Formatting
#'
#' @examples
#' \dontrun{
#'   c("\nA yellow line", "yellow")
#' }
#' @export
NULL
```

### LINKED ISSUE #731: braces in roxygen comments cause styling error
``` r
styler::cache_deactivate()
#> Deactivated cache.
styler::style_text(
  c(
    "#' Hi",
    "#' @examples",
    "#' x <- \"'{'\"", # also with single quotes
    "#' \\dontrun{",
    "#' fu(x = 3)",
    "#' }",
    "NULL"
  )
)
#> Warning in tools::parse_Rd(., fragment = TRUE): <connection>:6: unexpected END_OF_INPUT '
#> '
#> Error: <text>:1:1: unexpected '}'
#> 1: }
#>     ^
```

<sup>Created on 2021-02-06 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

`(` and `[` apparently not affected.

lorenzwalthert:
Maybe there are other keywords in https://developer.r-project.org/parseRd.pdf we have to fix later. Closes #728. The solution approach is also outlined there. Leverage `roxygen2` infrastructure to generate Rd like output, then use `tools::parse_Rd()` (which we used wrongly on demasked roxygen comments, which created escaping problems). Also closes #731 (some speed implications, but not reflected in the benchmark). 

We also considered adding more examples to the `{touchstone}` benchmark since there was only one roxygen2 code styling example (and it was correctly formatted) but then figured that in `{styler}`, 10% of all function declarations had roxygen code examples, which would translate to ~1 example in the benchmark package. The benchmark does not include a `\dontrun{}` (which does not make it faster than not containing any, as `needs_rd_emulation()` is ran anyways) but not another escape sequence (on which it would most likely be slower).

Final benchmarks: 

* cache_applying: 0.02 -> 0.02 (5.8%)
* cache_recording: 0.76 -> 0.75 (-0.3%)
* without_cache: 2.36 -> 2.49 (5.8%)


lorenzwalthert:
Uggg, speed benchmarks: 

* cache_applying: 0.03 -> 0.03 (4.9%)
* cache_recording: 0.97 -> 1.17 (20.3%)
* without_cache: 3 -> 3.28 (9.4%)

Let's remove the expensive roxygen parsing when there are no characters to escape.

Issue #742, opened 2021-03-04
Error with code that contains the new experimental "examplesIf" tag


krlmlr:
``` r
styler::style_text("
#' @examplesIf TRUE
#' NULL
NULL
")
#> Error: <text>:1:4: unexpected numeric constant
#> 1: If TRUE
#>        ^
```

<sup>Created on 2021-03-04 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

krlmlr:
Also, it seems that if an .R file contains one single `#' @examplesIf`, it is not styled at all but a styling error is reported. Should we handle this more gracefully?

lorenzwalthert:
Thanks. Re: fail gracefully: this is tracked in #732.

lorenzwalthert:
Turns out this is more complicated than anticipated because 

* multiple `#' @examples` tags to end up in one `.Rd` file can't be processed together anymore since we must distinguish between `#' @examples` and `#' @examplesIf`.
* can't just always add a line break after `#' @examples` because condition should remain on same line.

I have some WIP on this. How likely is `#' @examplesIf` to make it beyond experimental? Because I don't want to put in 10h to see it go away in the next release...

krlmlr:
Can you treat `#' @examplesIf xxx` like `#' @examples` for now to simplify the code?

The tag is proving useful: https://github.com/search?q=org%3Acran+examplesIf&type=code. I think it's to stay.

@gaborcsardi: Are there plans to remove/replace `@examplesIf` in roxygen2?

gaborcsardi:
No, I think it will stay.

lorenzwalthert:
I agree we can treat it very similar to `@examples` now #748 is merged to close #746 and #747.

lorenzwalthert:
Are the following restrictions reasonable?

* there must be a line break between condition and code:
```r
@exampleIf {cond} 
{code}
```
* This implies that condition can be on separate line

```r
@exampleIf 
{cond} 
{code}
```
But all of it can't be on same line 
```r
@exampleIf {cond}  {code}
```
* if condition can span multiple lines
```r
@exampleIf {cond
is long} 
{code}
```

krlmlr:
I think the condition must be on the same line?

lorenzwalthert:
[...]

PR #743, opened 2021-03-04
Support for @examplesIf roxygen tag

krlmlr:
for #742. Ideally we'd add full support for `@examplesIf`, this gets us halfway. Not sure the final `//s*` is necessary.

Needs tests.
***
Here is how the current PR would change benchmark results when merged into master:

* cache_applying: 0.03 -> 0.03 (-1.3%)
* cache_recording: 1.19 -> 1.19 (-0.4%)
* without_cache: 3.44 -> 3.46 (0.6%)


lorenzwalthert:
Thanks. We should not match the whole line since you can have `#’ @examples 1+1` and thus is valid (and coveted by unit tests).

R/roxygen-examples-find.R
```diff
@@ -7,7 +7,7 @@
 #' @importFrom rlang seq2
 #' @keywords internal
 identify_start_to_stop_of_roxygen_examples_from_text <- function(text) {
-  starts <- grep("^#'\\s*@examples", text, perl = TRUE)
+  starts <- grep("^#'\\s*@examples\\s*$", text, perl = TRUE)
```
lorenzwalthert, line 10:
```suggestion
  starts <- grep("^#'\\s*@examples(\\s*|If\\s|\\t)", text, perl = TRUE)
```

lorenzwalthert, line 10:
```suggestion
  starts <- grep("^#'\\s*@examples(If\\s|\\s|\\t)", text, perl = TRUE)
```

R/roxygen-examples-find.R
```diff
@@ -7,7 +7,7 @@
 #' @importFrom rlang seq2
 #' @keywords internal
 identify_start_to_stop_of_roxygen_examples_from_text <- function(text) {
-  starts <- grep("^#'\\s*@examples", text, perl = TRUE)
+  starts <- grep("^#'\\s*@examples(If\\s|\\s|\\t)", text, perl = TRUE)
```
lorenzwalthert, line 10:
```suggestion
  starts <- grep("^#'\\s*@examples(If\\s|\\s|\\t|$)", text, perl = TRUE)
```

lorenzwalthert:
@krlmlr you can try this if you want, should work now. Maybe have a look at the test cases too and tell me if we miss anything...

R/roxygen-examples-add-remove.R
```diff
@@ -30,15 +30,29 @@ remove_roxygen_mask <- function(text) {
 #' @keywords internal
 remove_roxygen_header <- function(text) {
   text <- gsub("^\\s*@examples\\s*", "", text, perl = TRUE)
-  starts_with_blank <- text[1] == "\n"
-  c(text[1][!starts_with_blank], text[-1])
+  if (grepl("^If ", text[1])) {
```
krlmlr, line 33: This matches `@examples If` . Just sayin...

Can you use the regex from https://github.com/r-lib/styler/pull/743/files#diff-73645d1e1e747d5eeb45cbd47ddb2ef64c658ed5c585763efd004a1b08ba61f9R10, perhaps with rematch2 and a grouped match?

krlmlr (commented):
Looks good, thanks for the effort! I didn't review all the examples, will file new issues if anything breaks in practice.

R/roxygen-examples-add-remove.R
lorenzwalthert: Good catch. I didn't ever figure out how to use `{rematch2}` (and did not bother spending time), maybe I should... For now, I think using the regex you linked should be sufficient.

PR #761, opened 2021-03-25
Parse 0x consistently

### LINKED ISSUE #760: Expression not equivalent with scope = "line_breaks"
for a file in RSQLite.

``` r
styler::style_text(
  readLines("https://raw.githubusercontent.com/r-dbi/RSQLite/538b5c9708d9a6d90a855355c6b5a1e0bad49593/R/connect.R"),
  scope = "line_breaks"
)
#> Error: The expression evaluated before the styling is not the same as the expression after styling. This should not happen. Please file a bug report on GitHub (https://github.com/r-lib/styler/issues) using a reprex.
```

<sup>Created on 2021-03-25 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

lorenzwalthert:
Closes #760.

* cache_applying: 0.04 -> 0.04 (0.1%)
* cache_recording: 1.28 -> 1.28 (0.1%)
* without_cache: 3.95 -> 3.95 (0.1%)

fc31e73:
* cache_applying: 0.03 -> 0.03 (0.1%)
* cache_recording: 1.15 -> 1.16 (0.5%)
* without_cache: 3.33 -> 3.36 (0.7%)

d4449cb
* cache_applying: 0.03 -> 0.03 (-2.5%)
* cache_recording: 1 -> 1 (0%)
* without_cache: 2.91 -> 2.88 (-1%)

348298e
* cache_applying: 0.03 -> 0.03 (0%)
* cache_recording: 1.16 -> 1.16 (-0.5%)
* without_cache: 3.46 -> 3.41 (-1.6%)


R/parse.R
```diff
@@ -205,15 +207,23 @@ ensure_valid_pd <- function(pd) {
...
   is_problematic_string[is_problematic_string] <- candidate_substring == "["
   is_problematic_string
 }
 
+is_insufficiently_parsed_num_const <- function(pd) {
+  is_problematic_num_const <- pd$token == "NUM_CONST"
+  candidate_substring <- substr(
+    pd$text[is_problematic_num_const], 1L, 2L
+  )
+  is_problematic_num_const[is_problematic_num_const] <- candidate_substring == "0x"
```
krlmlr, line 224: A regex is shorter and most likely runs faster than creating and checking substrings:

```suggestion
  is_problematic_num_const[is_problematic_num_const] <- grepl("^0x", pd$text[is_problematic_num_const])
```

Same for `..._strings()` above.

R/parse.R
```diff
@@ -116,23 +116,25 @@ add_id_and_short <- function(pd) {
...
-  is_problematic_string <- identify_insufficiently_parsed_strings(pd, text)
-  problematic_strings <- pd[is_problematic_string, ]
+  is_problematic_string <- or(
+    is_insufficiently_parsed_strings(pd),
+    is_insufficiently_parsed_num_const(pd)
+  )
+  problematictext <- pd[is_problematic_string, ]
   is_parent_of_problematic_string <-
-    pd$id %in% problematic_strings$parent
+    pd$id %in% problematictext$parent
```
krlmlr, line 137: Underscores have gone missing here, is this intended?

krlmlr (commented):
Thanks for taking this on!

R/parse.R
lorenzwalthert: Right...

R/parse.R
lorenzwalthert: Nope, reverted.

R/parse.R
```diff
@@ -116,54 +116,57 @@ add_id_and_short <- function(pd) {
...
   ensure_valid_pd(pd)
-  is_problematic_string <- identify_insufficiently_parsed_strings(pd, text)
-  problematic_strings <- pd[is_problematic_string, ]
-  is_parent_of_problematic_string <-
-    pd$id %in% problematic_strings$parent
+  is_problematic_text <- or(
+    is_insufficiently_parsed_string(pd),
+    is_insufficiently_parsed_number(pd)
+  )
+  problematic_text <- pd[is_problematic_text, ]
```
krlmlr, line 135: I wonder if we could speed up the code if `is_problematic_text` is `FALSE` everywhere.

R/parse.R
```diff
@@ -205,15 +208,13 @@ ensure_valid_pd <- function(pd) {
...
-    pd$text[is_problematic_string], 1, 1
-  )
-  is_problematic_string[is_problematic_string] <- candidate_substring == "["
-  is_problematic_string
+is_insufficiently_parsed_string <- function(pd) {
+  grepl("^\\[", pd$text) & pd$token == "STR_CONST"
 }
 
+is_insufficiently_parsed_number <- function(pd) {
+  grepl("^0x", pd$text)
```
krlmlr, line 216: Do we need to check for `"STR_NUM"` here?

krlmlr:
I'm seeing with this PR:

``` r
styler::style_text(text = "a <- 0x2L", scope = "line_breaks")
#> Error in paste("Error in styler:::ensure_correct_txt().", "Please file an issue on GitHub (https://github.com/r-lib/styler/issues)", : argument is missing, with no default
```

<sup>Created on 2021-03-29 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

R/parse.R
lorenzwalthert: I first hoped we don't but since parent matches children too, it messes things up.

R/parse.R
lorenzwalthert: Probably, let's see what touchstone says...

lorenzwalthert:
Right... Now I get:

``` r
as.character(
  styler::style_text(text = "a <- 0x2L", scope = "line_breaks")
)
#> [1] "a <- 0x2L"
```

<sup>Created on 2021-03-29 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

R/parse.R
```diff
@@ -116,54 +116,60 @@ add_id_and_short <- function(pd) {
...
+  is_parent_of_problematic_string <- pd$id %in% problematic_text$parent
 
-  is_unaffected_token <- !(is_problematic_string | is_parent_of_problematic_string)
-  if (!any(is_problematic_string)) {
+  is_unaffected_token <- !or(
+    is_problematic_text, is_parent_of_problematic_string
+  )
+  if (!any(is_problematic_text)) {
     return(pd)
   }
```
krlmlr, line 146: Already checked above.

```suggestion
```

krlmlr (approved):
I think 1% less is a lot from such a small change. It's easy to get slower, it's way harder to speed up things.

lorenzwalthert:
Well I don't think it's actually making things faster, because from a theoretical point of view, it must make things slower I believe. There is some variance in the estimator but at least we have good reasons to believe it's not making things much slower.

krlmlr:
I agree if the shortcut is hit rarely. In our case, most of the code won't contain `0x2L` or the other problematic sequence, so in the majority of cases the shortcut will save time.

What is the other problem that's handled in this code?

There are chances that this problem is fixed upstream in R 4.2 or earlier. So, eventually, we may be able to get rid of that.

lorenzwalthert:
Ah sorry I was not referring to your return-early suggestion, but to the whole PR, which gets 1% faster for unseen code, which I don't think makes sense from a theoretical point of view but can be explained with uncertainty in the estimation. We'll see what https://github.com/r-lib/styler/pull/761/checks?check_run_id=2224978050 brings. The other problem that's handed in that `ensure_correct_txt()()` is very long strings which will be truncated in the parent text, which is apparently hardcoded in the R parser as noted in #216

lorenzwalthert:
Ok, with the change you proposed, it's actually unambiguous if we get faster or slower, because we return earlier also. Sorry for the confusion. Another benchmark run indicates -1.6%, so probably it's a net positive change for us in speed as well 🥳.

PR #814, opened 2021-06-25
Indent between comment and line break

### LINKED ISSUE #813: trailing comment within call breaks hanging indent
Compare the following:

```
writeLines(trimws('
list(
  a = 
    1,
  b = # comment here
    2
)'), tmp <- tempfile(fileext = '.R'))
styler::style_file(tmp)
writeLines(readLines(tmp))
# list(
#   a =
#     1,
#   b = # comment here
#   2
# )
```

It appears to be an issue inside calls; the following is unaffected:

```
a = 
  1
b = # comment here
  2
```

I confirmed this issue affects back to at least 1.1.0.

lorenzwalthert:
Closes #813.


R/rules-indention.R
```diff
@@ -54,12 +54,21 @@ indent_op <- function(pd,
 indent_eq_sub <- function(pd,
                           indent_by,
                           token = c("EQ_SUB", "EQ_FORMALS")) {
-  eq_sub <- which(pd$token %in% token)
```
lorenzwalthert, line 57: @MichaelChirico changing to boolean here because I think it's slightly more efficient, but later realised we need to convert to int again for handling your exception

R/rules-indention.R
```diff
@@ -54,12 +54,21 @@ indent_op <- function(pd,
...
-  if (length(eq_sub) == 0) {
+  eq_sub <- pd$token %in% token
+  if (!any(eq_sub)) {
     return(pd)
   }
-  has_line_break <- which(pd$lag_newlines > 0)
-  indent_indices <- intersect(eq_sub + 1, has_line_break)
+  has_line_break <- pd$lag_newlines > 0 | pd$token == "COMMENT"
+  indent_indices <- which(lag(eq_sub, default = FALSE) & has_line_break)
+  if (any(pd$token[indent_indices] == "COMMENT")) {
```
lorenzwalthert, line 63: As the case in #813 is probably occurring quite infrequently, avoid an unnecessary `map()` call for performance reasons.

lorenzwalthert:
@MichaelChirico also, if you have suggestions for improvement, happy to hear them 😀.

R/rules-indention.R
MichaelChirico: Wouldn't a subassign be faster?

```
comment_idx <- pd$token[indent_indices] == "COMMENT"
indent_indices[comment_idx] <- next_non_comment(pd, indent_indices[comment_idx])
```

(or a map on the RHS of `<-` if `next_non_comment` isn't vectorized in arg 2)

R/rules-indention.R
MichaelChirico: Thanks for fixing. How did you know to look in `indent_eq_sub` if I might ask?

MichaelChirico: Oh I just saw your comment on the issue, nvm.

lorenzwalthert: I just full text searched the rules in `R/rules-indention.R` With “eq_sub” to find a rule that has this in the name.

R/rules-indention.R
lorenzwalthert: It’s not vectorised. Good thing now with {touchstone} we can see what’s faster for an average Code input. I can try that. Or we can benchmark just this changed expression locally.

lorenzwalthert: apparently not, sub-assignment and sub-setting is also expensive:
```r
# devtools::load_all()
# debug(indent_eq_sub)
# run some code

bench::mark(
  first = {
    indent_indices <- which(lag(eq_sub, default = FALSE) & has_line_break)
    x <- purrr::map_int(indent_indices, function(idx) {
    if (pd$token[idx] == "COMMENT") {
      next_non_comment(pd, idx)
    } else {
      idx
    }
  })
  x
  },
  second = {
    indent_indices <- which(lag(eq_sub, default = FALSE) & has_line_break)
    comment_idx <- pd$token[indent_indices] == "COMMENT"
    indent_indices[comment_idx] <- purrr::map_int(indent_indices[comment_idx], function(idx) {
      next_non_comment(pd, idx)
    })
    indent_indices
    }
)

#>  # A tibble: 2 x 13
#>  expression     min  median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory  time   gc    
#>  <bch:expr> <bch:t> <bch:t>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>  <list> <list>
#> 1 first         88µs  95.4µs     8973.        0B     2.03  4420     1      493ms <int … <Rprof… <benc… <tibb…
#> 2 second      89.8µs  95.2µs     9172.        0B     4.26  4310     2      470ms <int … <Rprof… <benc… <tibb…

```

Issue #480, opened 2019-03-08
Support .Rmarkdown as .Rmd


lorenzwalthert:
For blogdown, you can use .Rmarkdown instead of .Rmd to generate tocs and other things. Reference: https://bookdown.org/yihui/blogdown/output-format.html. This format should be treated as equivalent to .Rmd by styler.

lorenzwalthert:
Not sure if this file format should just be treated as Rmd. This would avoid some overhead in the code but make it impossible for the user to distinguish between Rmd and Rmarkdown.

lorenzwalthert:
Not sure this has a high priority anymore, since {hugodown}, the successor of {blogdown}, does not use `.Rmarkdown`.

PR #824, opened 2021-07-15
Treat .Rmarkdown files as .Rmd files

JohannesNE:
.Rmarkdown files are like .Rmd files, except they render to .md instead of .html (https://bookdown.org/yihui/blogdown/output-format.html)

Unfortunately they are rejected by styler (`Error: Can only style .R, .Rmd and .Rnw files.`)


JohannesNE:
If you would rather implement this differently, of if I've overlooked something, feel free to just reject at implement it properly.

lorenzwalthert:
Thanks @JohannesNE. Closes: https://github.com/r-lib/styler/issues/480. Before a merge, we should:

* add tests: `tests/testthat/test-public-api.R`. Look for `Rmd` and add one test for every API function (`style_file()`, `style_dir()`, `style_pkg()`). Just look for 'Rmd' in that file (starting around line 134).
* Add a news bullet: Files with `.Rmarkdown` extensions are now recognised as an R markdown files in `style_file()` and friends (#824).
* In the documentation, state explicitly that the new extension is treated as R markdown, i.e. adapt the documentation of the parameter `filetype` in `prettify_pkg()`.

Do you want to give this a try?

JohannesNE:
My experience is limited, but I have given it a try.

NEWS.md
```diff
@@ -2,6 +2,8 @@
 
 * Don't break line before comments in pipes (#822).
 
+* Files with .Rmarkdown extensions are now recognised as an R markdown files in `style_file()` and friends (#824)
```
lorenzwalthert, line 5:
```suggestion
* Files with `.Rmarkdown` extension are now recognised as an R markdown files in `style_file()` and friends (#824)
```

lorenzwalthert:
@JohannesNE looks great overall, thanks a lot. Let's keep our fingers crossed for the tests to pass 🙏.

JohannesNE:
Sorry. I actually meant to make a comment that i was unsure about committing the updated reference files, but forgot about it.

lorenzwalthert:
There is something strange going on with some unrelated reference files I think. When I locally run the tests, I get modified files (so tests fail), but on CI/CD, the same reference files pass. I reverted to the old ref files in 2e64b07, let's hope we can make tests pass. Maybe I'll open a separate issue. The problem persists for some time already.

lorenzwalthert:
Are you working on macOS as well?

JohannesNE:
I'm on Ubuntu 21.04. I am afraid won't be of much help with that problem. I usually "fix" encoding problems by trial-and-error :man_shrugging:

lorenzwalthert:
> I'm on Ubuntu 21.04. I am afraid won't be of much help with that problem. I usually "fix" encoding problems by trial-and-error 🤷‍♂️

Ok. Seems not a problem in CI/CD, neither on CRAN, so I guess we can leave it for now.

lorenzwalthert:
Thanks @JohannesNE 🥳

PR #849, opened 2021-10-08
Also recognize lintr ignore markers

### LINKED ISSUE #848: {lintr} ignore comments should be recognized as instructions to refrain stying
Lintr has `# nolint` and similar (https://github.com/jimhester/lintr).

lorenzwalthert:
@MichaelChirico what do you think about this? From NEWS.md:

> multiple stylerignore patterns can be specified at once and lintr markers 
  `# nolint`, `# nolint start` and `# nolint end` are now by default recognized
  in addition to `# styler: off` and `# styler: on` (#849).

Closes #848

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -46,7 +46,7 @@ VignetteBuilder:
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE, roclets = c("rd", "namespace", "collate",
     "pkgapi::api_roclet"))
-RoxygenNote: 7.1.1.9001
+RoxygenNote: 7.1.2
 Collate: 
     'addins.R'
     'communicate.R'
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 # styler 1.6.2.9000 (Development version)
 
 
+* multiple stylerignore patterns can be specified at once and lintr markers 
+  `# nolint`, `# nolint start` and `# nolint end` are now by default recognized
+  in addition to `# styler: off` and `# styler: on` (#849).
+
 # styler 1.6.2
 
 * clean up cache files older than one week (#842).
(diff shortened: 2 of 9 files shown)


MichaelChirico:
I'm not sure about turning it on by default, though the extension to recognize multiple is good for allowing them to be set in sync if so desired.

indeed I just wrote some code this week like

```
# nolint start: line_length_linter. styler: off.
...
# nolint end: line_length_linter. styler: on.
```

which would look cleaner if there's a single symbol.

lorenzwalthert:
stylerignore markers must be alone on a line, so you code won't work.

``` r
styler::style_text('
# nolint start styler: off
call+1
# nolint end styler: on
')
#> 
#> # nolint start styler: off
#> call + 1
#> # nolint end styler: on
```

<sup>Created on 2021-10-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

An alternative would be to make this work too by matching the regex `styler: off` in the comment instead of an exact match of the comment.

MichaelChirico:
Probably the most common `styler: off` case that I see/use myself is for allowing whitespace alignment:

```
cbind(
  c( 'a',  'b',  'c',  'd'),
  c(1.11, 1.23, 1.44, 1.61)
)
```

And many similar cases where we want to use whitespace to make the code cleaner to read. In those cases I still want (most) lints to apply... so I'm not sure what the default should be.

Is there any empirical check we can do? How reliable do we think CRAN revdeps are for giving us insight here?

Definitely think a regex match is good... or as another suggestion, an exact match of `styler: off` on the `text` of a `COMMENT` node in the AST (prevents matching inside code, like `"# styler: off"`)

lorenzwalthert:
@MichaelChirico any final thoughts?

Issue #847, opened 2021-10-04
{styler} changes unicode characters on Windows


IndrajeetPatil:
Example:

``` r
styler::style_text('suit <- "♠"')
#> suit <- "<U+2660>"
```

<sup>Created on 2021-10-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> - Session info ---------------------------------------------------------------
#>  setting  value                       
#>  version  R version 4.1.1 (2021-08-10)
#>  os       Windows 10 x64              
#>  system   x86_64, mingw32             
#>  ui       RTerm                       
#>  language (EN)                        
#>  collate  English_United Kingdom.1252 
#>  ctype    English_United Kingdom.1252 
#>  tz       Europe/Berlin               
#>  date     2021-10-04                  
#> 
#> - Packages -------------------------------------------------------------------
#>  package     * version date       lib source        
#>  backports     1.2.1   2020-12-09 [1] CRAN (R 4.1.0)
#>  cli           3.0.1   2021-07-17 [1] CRAN (R 4.1.0)
#>  crayon        1.4.1   2021-02-08 [1] CRAN (R 4.1.1)
#>  digest        0.6.28  2021-09-23 [1] CRAN (R 4.1.1)
#>  ellipsis      0.3.2   2021-04-29 [1] CRAN (R 4.1.0)
#>  evaluate      0.14    2019-05-28 [1] CRAN (R 4.1.1)
#>  fansi         0.5.0   2021-05-25 [1] CRAN (R 4.1.1)
#>  fastmap       1.1.0   2021-01-25 [1] CRAN (R 4.1.1)
#>  fs            1.5.0   2020-07-31 [1] CRAN (R 4.1.1)
#>  glue          1.4.2   2020-08-27 [1] CRAN (R 4.1.1)
#>  highr         0.9     2021-04-16 [1] CRAN (R 4.1.1)
#>  htmltools     0.5.2   2021-08-25 [1] CRAN (R 4.1.1)
#>  knitr         1.36    2021-09-29 [1] CRAN (R 4.1.1)
#>  lifecycle     1.0.1   2021-09-24 [1] CRAN (R 4.1.1)
#>  magrittr      2.0.1   2020-11-17 [1] CRAN (R 4.1.1)
[...]

Issue #783, opened 2021-04-29
Adopt to testthat 3e


lorenzwalthert:
https://testthat.r-lib.org/articles/third-edition.html

PR #856, opened 2021-10-22
test for #847

IndrajeetPatil:


IndrajeetPatil:
Another possibility is to use [snapshot tests](https://www.tidyverse.org/blog/2021/10/testthat-3-1/#snapshot-tests), with a different snap for each R version:

image

lorenzwalthert:
@IndrajeetPatil there is #783 for testthat 3e, but I don't think we need snaps for making the test R version dependent. Superseded by #883.

PR #868, opened 2021-11-28
Benchmark / diffview against CRAN release v1.10.0

lorenzwalthert:
DO NOT MERGE, this is for benchmarking only. 

**You may want to unsubscribe from this PR notifications, as any merge into main will trigger notifications.**

*** 

Comparision with previous releases:

Releae 1.7.0

* &nbsp;&nbsp;:rocket:cache_applying: 29.1ms -> 28.2ms [-5.98%, -0.35%]
* &nbsp;&nbsp;:rocket:cache_recording: 1.24s -> 1.22s [-1.91%, -0.46%]
* :exclamation::snail:without_cache: 3.21s -> 3.3s [+2.1%, +3.94%]


IndrajeetPatil:
@lorenzwalthert For the **exact same commit,** I ran the touchstone workflow twice; once the speed-up (compared to the CRAN version) was significant, while the other time it wasn't. Such drastic variation in results is to be expected?

Screenshot 2022-09-14 at 17 32 22

lorenzwalthert:
@IndrajeetPatil it's not. We can trigger this a few more times and see if things change. I.e. see if it was just once in a blue moon or happens all the time. You can also review the touchstone code and see what we do, it's a [linar model](https://github.com/lorenzwalthert/touchstone/blob/main/R/analyze.R#L146) where we regress time spent on branch.
And we interleave both branches in [blocks of two](https://lorenzwalthert.github.io/touchstone/articles/inference.html) so I don't think there is a theoretical problem with that.

IndrajeetPatil:
> @IndrajeetPatil it's not. We can trigger this a few more times and see if things change. I.e. see if it was just once in a blue moon or happens all the time. 

@lorenzwalthert Thanks for the explanation. Looking at results from triggering this a few more times, looks like this happens quite frequently? 

I will also have a look at the model we are using to draw statistical inferences here.

IndrajeetPatil:
@lorenzwalthert This should change to against 1.8.0, right?

lorenzwalthert:
@IndrajeetPatil exactly, just changed the base branch and PR name.

IndrajeetPatil:
Thanks!

IndrajeetPatil:
Nice! The minor speed-up seems to be legit.

Issue #747, opened 2021-03-10
Multiple @examples tags are squashed into one


lorenzwalthert:
``` r
styler::style_text(c(
  "#' Example",
  "#' ",
  "#' @examples",
  "#' 1",
  "#' @examples",
  "#' 1",
  "NULL"
))
#> #' Example
#> #'
#> #' @examples
#> #' 1
#> #'
#> #' 1
#> NULL
```

<sup>Created on 2021-03-10 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

Fixing this is a pre-condition to implement #742 since we can't squash `@examples` and @examplesIf`.

lorenzwalthert:
Closed with #748.

PR #880, opened 2021-12-22
Keep at most one empty line at the end of each example

krlmlr:
This is crucial for combining multiple `@examples` and `@examplesIf` elements.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -37,7 +37,7 @@
   detected (#881).
 * Add vignette on distributing style guide (#846, #861).
 * ensure a trailing blank line also if the input is cached (#867).
-* Preserve trailing blank line in roxygen examples (#880).
+* Preserve trailing blank line in roxygen examples to simplify concatenation of examples (#880).
 * Fix argument name `filetype` in Example for `style_dir()` (#855).
 * An error is now thrown on styling if input unicode characters can't be 
   correctly parsed for Windows and R < 4.2 (#883).
--- a/NEWS.md
+++ b/NEWS.md
@@ -37,6 +37,7 @@
   detected (#881).
 * Add vignette on distributing style guide (#846, #861).
 * ensure a trailing blank line also if the input is cached (#867).
+* Preserve trailing blank line in roxygen examples (#880).
 * Fix argument name `filetype` in Example for `style_dir()` (#855).
 * An error is now thrown on styling if input unicode characters can't be 
   correctly parsed for Windows and R < 4.2 (#883).
--- a/R/roxygen-examples.R
+++ b/R/roxygen-examples.R
@@ -102,6 +102,9 @@ style_roxygen_example_snippet <- function(code_snippet,
     mask <- decomposed$mask
   }
   code_snippet <- post_parse_roxygen(code_snippet)
+  append_empty <- !is_dont &&
+    length(code_snippet) > 1L &&
+    last(code_snippet) == ""
 
   cache_is_active <- cache_is_activated()
   is_cached <- is_cached(
@@ -116,9 +119,12 @@ style_roxygen_example_snippet <- function(code_snippet,
       parse_transform_serialize_r(transformers,
         base_indention = base_indention, warn_empty = FALSE
       )
-  } else {
-    code_snippet <- ensure_last_n_empty(code_snippet, n = 0)
   }
+
+  code_snippet <- ensure_last_n_empty(
+    code_snippet,
+    n = ifelse(append_empty, 1L, 0L)
+  )
 
   if (!is_cached && cache_is_active) {
     cache_write(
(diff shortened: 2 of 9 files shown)


lorenzwalthert:
I am not sure I can follow. Can you provide a new test to demonstrate the expected behavior?

lorenzwalthert:
Also, we don't squash examples anymore: https://github.com/r-lib/styler/issues/747

krlmlr:
Thanks. The "examples" are tibble and pillar -- try styling e.g. `add.R` in tibble with this PR and with the main branch. I'll add a formal test tomorrow.

lorenzwalthert:
Ok, so what's the rationale of having a blank line like between roxygen comments this?


```r

#' Add columns to a data frame
#'
#' This is a convenient way to add one or more columns to an existing data
#' frame.
#'
#' @param .data Data frame to append to.
#' @param ... <[`dynamic-dots`][rlang::dyn-dots]>
#'   Name-value pairs, passed on to [tibble()]. All values must have
#'   the same size of `.data` or size 1.
#' @param .before,.after One-based column index or column name where to add the
#'   new columns, default: after last column.
#' @inheritParams tibble
#' @family addition
#' @examples
#' # add_column ---------------------------------
#' df <- tibble(x = 1:3, y = 3:1)
#'
#' df %>% add_column(z = -1:1, w = 0)
#' df %>% add_column(z = -1:1, .before = "y")
#'
#' # You can't overwrite existing columns
#' try(df %>% add_column(x = 4:6))
#'

#' # You can't create new observations
#' try(df %>% add_column(z = 1:5))
#'
#' @export
add_column
```

lorenzwalthert:
having blocks in the docs like here?
Screenshot 2021-12-22 at 18 55 47

krlmlr:
No, the blank line is bogus and can be removed. This is about the empty line before `#' @export`, the help page contains other functions with more examples, and I need to keep that empty line to visually separate the examples from each other.

krlmlr:
I have updated the test output files, there's still trouble. Could you please take a look?

Also, there are "unknown or unitialised token column" warnings, I have seen them before with CRAN styler.

I do wonder why this change adds so many empty lines for some outputs, in particular in the presence of `\dontrun{}`.

krlmlr:
Empty lines look good to me now. Would you support this change?

krlmlr:
Regarding the tidyverse style guide: not sure, it just feels that examples would be too condensed without empty lines inbetween.

lorenzwalthert:
If I understand you correctly from your last two comments, all we need is to keep an empty line at the end of an example section **if** already one or more empty lines are present (empty as in only `#'` on that line, not completely empty). If there is none, we should not add one. Correct?

krlmlr:
Yes, that would be great.


On 23 Dec 2021 at 12:51:14 CET, Lorenz Walthert ***@***.***> wrote:

   If I understand you correctly from your last two comments, all we need is to keep an empty line at the end of an example section  if  already one or more empty lines are present (empty as in only  #'  on that line, not completely empty). If there is none, we should not add one. Correct?  —
Reply to this email directly,  view it on GitHub , or  unsubscribe .
Triage notifications on the go with GitHub Mobile for  iOS  or  Android .

lorenzwalthert:
The warning about unknown or uninitialized column seems to happen in various places in the code (as you can tell from the warnings). And it's not new. Probably it's similar to this:

``` r
x <- tibble::tibble()
x$ff[logical()] <- 1
#> Warning: Unknown or uninitialised column: `ff`.
```

<sup>Created on 2021-12-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

For some reason, `pd` has sometimes now rows and the condition for the assignment is `logical()`, which has length 0.

Also, don't modify the `.*-utf8` reference objects. I don't know why they don't fail on CRAN, but they fail for me locally too. Wird uns wohl irgendwann um die Ohren fliegen 😄

lorenzwalthert:
@krlmlr can you have a look at the new test case I added and if that's what you want, you can merge it 🥳

NEWS.md
```diff
@@ -37,6 +37,7 @@
   detected (#881).
 * Add vignette on distributing style guide (#846, #861).
 * ensure a trailing blank line also if the input is cached (#867).
+* Preserve trailing blank line in roxygen examples (#880).
```
krlmlr, line 40:
```suggestion
* Preserve trailing blank line in roxygen examples to simplify concatenation of examples (#880).
```

krlmlr:
Are you suggesting that `pd` has no columns sometimes? Even if it has no rows and the column exists, the assignment works:

``` r
x <- tibble::tibble(ff = numeric())
x$ff[logical()] <- 1
x
#> # A tibble: 0 × 1
#> # … with 1 variable: ff <dbl>
```

<sup>Created on 2021-12-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

We should normalize the data early in the process so that the structure is stable.

lorenzwalthert:
> Are you suggesting that pd has no columns sometimes? 

Well, it has columns. But I think the problem is that it does not habe the ones to which something is assigned. Comparing my example with yours, yours already has the column that is assigned to. I’ll open a nee issue.

PR #902, opened 2022-01-14
Fix touchstone

lorenzwalthert:


assignUser:
@lorenzwalthert it looks like the comment action was triggered with the .yaml from `main` https://github.com/r-lib/styler/actions/runs/1702426418/workflow 
no idea if it's possible to force that or if you need to merge this first...

lorenzwalthert:
@assignUser I think it's triggered from main - and by design. Otherwise, I could just PR to a random repo and execute random code with their credentials, no?

assignUser:
@lorenzwalthert Oh yeah, that makes sense for files with the workflow trigger!

lorenzwalthert:
Every merge to main triggers a new touchstone CI run because I want to see how *merge main into latest release* changes the performance of the released version with #868, so we'll see.

PR #949, opened 2022-05-26
Upgrade to testthat 3e

### LINKED ISSUE #783: Adopt to testthat 3e

IndrajeetPatil:
Closes #783

TODO:

- [x] Remove `context()` statements using-

```r
xfun::gsub_dir(
  'context\\(.+\\)',
  '',
  ext = "R",
  dir = "tests/"
)
```

- [x] `expect_equivalent()` -> `expect_equal()`
- [x] Replace `expect_known_value()` with `expect_snapshot_value()`
- [x] Delete unnecessary binary files
- [ ] ~~Use `expect_snapshot_error(...)` instead of `expect_error(..., regexp)`~~
- [ ] ~~Use `expect_snapshot_warning(...)` instead of `expect_warning(..., regexp)`~~
- [ ] ~~Use `expect_snapshot_message(...)` instead of `expect_message(..., regexp)`~~


lorenzwalthert:
You are on a roll today 🥳. All checks green so far. Also, some checks in `test-public_api.R` (or similar) are checking for printed output too, maybe they should also use snapshots instead. Did you see https://www.tidyverse.org/blog/2022/02/upkeep-testthat-3/ ?

lorenzwalthert:
Also note that some utf8 related tests give a different Output for me locally, but it passes CI and CRAN, so I suggest to not touch them.

IndrajeetPatil:
> maybe they should also use snapshots instead

Yes, I am now using snapshots for the relevant tests. Let's see if the tests pass now.

If I understand the previous testing infrastructure correctly, with a switch to snapshots, the binary files [here](https://github.com/r-lib/styler/tree/main/tests/testthat/public-api/xyzdir-dirty) should be deleted, right?

IndrajeetPatil:
@lorenzwalthert The following code:

```r
 cache_info[, c("n", "size", "last_modified", "activated")]
```

produces the following output on R versions > oldrel-1

```r
      # A tibble: 1 x 4
            n  size last_modified activated
        <int> <dbl> <dttm>        <lgl>    
      1     0     0 -Inf -Inf     FALSE 
```

and otherwise

```r
      # A tibble: 1 x 4
            n  size last_modified activated
        <int> <dbl> <dttm>        <lgl>    
      1     0     0 NA NA     FALSE 
```

Should this test be skipped on older versions?

lorenzwalthert:
Yes. That one can be skipped on old R versions. And the binary (but not the R files). can be deleted in that dir.

lorenzwalthert:
@IndrajeetPatil about the errors: I am not sure we want to catch exact errors that are defined outside {styler}. In the past, R core changed the parser and I had to release a new styler version just to fix these. I am happy snapshotting our own errors but probably it's better to rely on regex for f1d87bd and the like.

MichaelChirico:
> @IndrajeetPatil about the errors: I am not sure we want to catch exact errors that are defined outside {styler}. In the past, R core changed the parser and I had to release a new styler version just to fix these. I am happy snapshotting our own errors but probably it's better to rely on regex for [f1d87bd](https://github.com/r-lib/styler/commit/f1d87bdbce5300316b4acc7b5b2a270b840eb535) and the like. Is there a way to do that with the third edition?

You might also consider generating the base error message yourself and comparing to that -- should be even more stable than regexes. `data.table` does this -- see the `get_msg()` helper in our main test suite: https://raw.githubusercontent.com/Rdatatable/data.table/master/inst/tests/tests.Rraw

lorenzwalthert:
Thanks @MichaelChirico. But what's the advantage over `expect_error()` in that context. Since `expect_error()` is not (soft-)deprecated or superseded, I think we should just continue to use it for those type of non-styler messages.

IndrajeetPatil:
> but probably it's better to rely on regex for https://github.com/r-lib/styler/commit/f1d87bdbce5300316b4acc7b5b2a270b840eb535 and the like. Is there a way to do that with the third edition?

Yeah, the third edition has no issues with using regex for testing error messages. So basically nothing needs to be changed with these tests.

In this case, there is nothing more that needs to be done for this PR.

IndrajeetPatil:
@lorenzwalthert I am not sure why pre-commit GHA is failing now. Seems to do with end of files, but not sure how to fix this:

Screenshot 2022-05-31 at 08 55 35

lorenzwalthert:
Well there is some [important definition / convention](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file) that a sffile ends with one and exactly one line break. That's enforced with that hook, so here are the options to fix it:
* remove the redundant lines manually.
* install pre-commit locally and run `pre-commit run` (makes sense if you want to use that framework in the future)
* file an issue with testthat that every snapshot should contain one and exactly one empty line at EOF. That of course would be a breaking chnage to testthat snaps, but I think it's the right thing to do. Do you want to file an issue or should I?

IndrajeetPatil:
I am not sure if this is a _testthat_ issue.

This happens because the code output itself contains a line break, and then _testthat_ adds another line break.

Screenshot 2022-05-31 at 11 22 16

As a result, `utils.md` contains two line breaks:

Screenshot 2022-05-31 at 11 24 25

IndrajeetPatil:
The easiest and quickest solution here is to remove the extra line break manually.

lorenzwalthert:
Adapt `.pre-commit-config.yaml` like this
```yaml
    -   id: end-of-file-fixer
        exclude: >
          (?x)^(
          \.Rd|
          tests/testthat/exception_handling/empty_file\.R|
          tests/testthat/parse_comments/eol_eof_spaces-.*|
          tests/testthat/reference-objects/.*|
          tests/testthat/_snaps/*|
          )$
```

lorenzwalthert:
Perfect @IndrajeetPatil. Thanks a lot. I also added a NEWS entry.

IndrajeetPatil:
Thanks, Lorenz!

Btw, _precommit_ is really awesome, so thanks for that as well. Included it here:
https://github.com/IndrajeetPatil/awesome-r-pkgtools/blob/master/README.md

PR #950, opened 2022-05-31
Upgrade to roxygen2 7.2.0

IndrajeetPatil:


R/testing.R
```diff
@@ -213,19 +213,6 @@ testthat_file <- function(...) {
...
-#' Convert a serialized R object to a certain version.
-#'
-#' Needed to make [testthat::expect_known_value()] work on R < 3.6.
-#' @param path A path to an rds file.
-#' @param version The target version.
-#' @keywords internal
-rds_to_version <- function(path, version = 2) {
-  readRDS(path) %>%
-    saveRDS(path, version = version)
-}
```
IndrajeetPatil, line 225: This is no longer needed because reliance on `testthat::expect_known_value()` was removed in #949.

IndrajeetPatil:
I am not sure where this difference between my local machine and CI checks is coming from. I am using the release versions of both _roxygen2_ and _dplyr_.

Screenshot 2022-05-31 at 18 27 07

lorenzwalthert:
Well, remember that pre-commit uses virtual environments, in this case a renv. pre-commit uses hook revision `v0.2.2.9015` according to `.pre-commit-config.` Checking `renv.lock` for that tag, you can see that [{roxygen} 7.1.2](https://github.com/lorenzwalthert/precommit/blob/v0.2.2.9015/renv.lock) was used. So, I suggest to run to update the hook revision of all hook repos with `$ pre-commit autoupdate`  in {styler}.

R/testing.R
lorenzwalthert: Good catch.

R/utils-cache.R
```diff
@@ -16,8 +16,9 @@ hash_standardize <- function(text) {
 #' Check if text is cached
 #'
 #' This boils down to check if the hash exists at the caching dir as a file.
-#' @param text,transformers,more_specs Passed to [cache_make_key()] to generate
-#'   a key.
+#' @param text Passed to [cache_make_key()] to generate a key.
+#' @param transformers Passed to [cache_make_key()] to generate a key.
+#' @param more_specs Passed to [cache_make_key()] to generate a key.
```
lorenzwalthert, line 21: How is this better than documenting each individually?

IndrajeetPatil: It's not that this is better. But this is no longer supported in the recent version of roxygen2. If you inherit these docs in another function, only the first parameter will be inherited; not the rest.

IndrajeetPatil: e.g. see [this](https://github.com/r-lib/roxygen2/issues/1347) issue.

IndrajeetPatil:
> Well, remember that pre-commit uses virtual environments, in this case a renv. pre-commit uses hook revision `v0.2.2.9015` according to `.pre-commit-config.` Checking `renv.lock` for that tag, you can see that [{roxygen} 7.1.2](https://github.com/lorenzwalthert/precommit/blob/v0.2.2.9015/renv.lock) was used. So, I suggest to run to update the hook revision of all hook repos with `$ pre-commit autoupdate` in {styler}. pre-commit.ci would do that for us, but it's not (yet) activated for r-lib. @IndrajeetPatil are you using pre-commit locally? If yes, please run these commands and commit, otherwise I can do it.

Thanks! 
I locally installed precommit via condda and ran `precommit::autoupdate()`, which did increase the hook version in the `.yaml` file, but the problem seems to persist.

Feel free to push into the PR; I don't think I can solve this on my end.

PR #956, opened 2022-06-16
Avoid ANSI escapes on website

### LINKED ISSUE #954: ANSI escape codes showing on the website
For example, see [here](https://styler.r-lib.org/dev/articles/styler.html)

Screenshot 2022-06-16 at 12 45 05

IndrajeetPatil:
Closes #954

I created a local website for the package and can confirm that the suggested solution indeed solves the issue.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -32,6 +32,8 @@ editor_options:
 
 **Other**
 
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
Thank you @IndrajeetPatil, I also added a news billet for that fix here 🥳

PR #964, opened 2022-07-19
Update to roxygen2 7.2.1

IndrajeetPatil:


IndrajeetPatil:
@lorenzwalthert I think the precommit GHA failure is because it is not using the most recent CRAN version of _roxygen2_?

lorenzwalthert:
Thanks. This does not work for pre-commit. Please add `roxygen@:version` to the  `additional_dependencies` in `.pre-commit-config.yaml` as there is currently not hook revision that has this {roxygen2} version. If you are a GitHub Actions pro, you could also look into helping me to fix it with https://github.com/lorenzwalthert/precommit/issues/433.

IndrajeetPatil:
Hmm, the problem seems to be that [RSPM](https://packagemanager.rstudio.com/all/__linux__/bionic/latest/src/contrib/roxygen2_7.2.1.tar.gz) also doesn't have the latest version.

> package 'roxygen2' with version '7.2.1' not found

Issue #769, opened 2021-04-09
upcoming breaking changes


lorenzwalthert:
These are probably candidates for release 2.0:

* remove `save_after_styling` env variable.
* `more_args` must come last and have no default in `create_style_guide()`.

PR #965, opened 2022-07-20
Default to styling all supported file formats

### LINKED ISSUE #963: Default to styling all supported file types in `style_dir()` and `style_pkg()`?
I am wondering if `style_dir()` and `style_pkg()` shouldn't style all supported file formats in the directory/package by default. 

Is there any reason why the users may wish to format only code in `.R` files, but not in (e.g.) `.Rmd` files? 

Current default:

```r
filetype = c("R", "Rprofile")
```

Proposed default:

```r
filetype = c("R", "Rprofile", "Rmd", "Rmarkdown", "Rnw")
```

IndrajeetPatil:
Closes #963


IndrajeetPatil:
@lorenzwalthert All green now.

lorenzwalthert:
Thaks for the PR 😄. A few comments:
* the `API` file should also be updated. Can you install `r-lib/pkgapi` and re-run roxygenize?
* If it is a breaking change (which I am not sure yet), we should probably not merge it into main (but like `rc-2.0.0`), fix all the other issues in #769 and then release v2.0.0.

IndrajeetPatil:
The API file is already updated. precommit caught that.

IndrajeetPatil:
> If it is a breaking change (which I am not sure yet)

Okay, let me know when you come to a decision on this point. It also doesn't feel like a breaking change to me, bur rather a major change.

IndrajeetPatil:
Hmm, why did the Continuous Benchmarks GHA fail?

lorenzwalthert:
> Why did {touchstone} fail?

@assignUser can you help here? I think it's related to some non-R code that you contributed that I can't understand. Same problems occured in https://github.com/lorenzwalthert/touchstone/issues/97#issuecomment-1162140464.

assignUser:
The Commenting step failed because the main touchstone workflow failed and this correctly added a ❌ to the check suite. The actual touchstone run timed out during dependency installation, so probably an issue unrelated to the PR/touchstone.

So overall everything works as intended (except the time out 😁 ), I would try re-running/re-triggering the benchmark

lorenzwalthert:
@assignUser oh yes right. Thanks for explaining. So basically every time the first part of the action fails, this same error ocurrs, right? Guess that's just something we have to know if errors occur and then dig deeper... Anyways, I restarted the benchmark. It has an old rspm checkpoint so I am surprised it failed. Let's see what happens...

lorenzwalthert:
GitHub Actions now passing, had to update some config, see #969.

IndrajeetPatil:
Thanks, @lorenzwalthert! Yes, the builds are indeed passing now.

Btw, having slept a bit more on it, I don't think this would qualify as a breaking change. A breaking change would have been if we had stopped supporting styling a file type. Here, we continue to support everything we were supporting _and some more file types_.

NEWS.md
```diff
@@ -6,6 +6,12 @@ editor_options:
 
 # styler 1.7.0.9000 (Development version)
 
+**Breaking Changes**
```
lorenzwalthert, line 9:
```suggestion
**User-facing changes**
```

lorenzwalthert (approved):

lorenzwalthert:
yeah, you are right. I marked it as *user-facing change*.

lorenzwalthert:
Thanks 🎉

PR #967, opened 2022-07-26
better multi-line while() statement handling

### LINKED ISSUE #966: { gets bumped in compound while statement --> open curly violation
```r
styler::style_text('
while ({
  bad <- grepl("_", z)
  any(bad)
}) {
  TRUE
}
')
# while (
#   {
#     bad <- grepl("_", z)
#     any(bad)
#   })
# {
#   TRUE
# }
```

The `{` is bumped but the [style rules dictate `{` should be on the line with `)`](https://style.tidyverse.org/syntax.html#indenting):

> `{` should be the last character on the line. Related code (e.g., an `if` clause, a function declaration, a trailing comma, …) must be on the same line as the opening brace.

AFAIK clauses like `({ ... })` are generally fine, so I'm also not sure why `(` and `{` were separated...

lorenzwalthert:
Closes #966.


MichaelChirico:
Thanks for the quick fix here!!

PR #972, opened 2022-08-07
Document UI in a single `.Rd` file

### LINKED ISSUE #971: Document the styling UI in a single `.Rd` file?
Currently, these functions are [separately documented](https://styler.r-lib.org/reference/index.html) and there is a lot of duplication of the content.

Screenshot 2022-07-29 at 14 54 04

Instead, we could do this the same way `{lintr}` does it; cataloguing them [in a single place](https://lintr.r-lib.org/reference/lint.html).

Screenshot 2022-07-29 at 14 54 32

If you agree, I can make a PR.

IndrajeetPatil:
Closes #971


IndrajeetPatil:
@lorenzwalthert This should be ready for a review.

IndrajeetPatil:
*bump*

lorenzwalthert:
Hi @IndrajeetPatil. After looking at it again, I decided that I don't actually think it's a good idea to document that in the same file. Here's why:

* Bigger helpfile: Because of the way {roxygen2} formats the help file, we now have to scroll quite a bit to get to the arguments.
* Order of arguments in Usage section =! order of arguments in Arguments: The path argument comes second last, but it's the first argument in `style_file()`. There is no way of aligning Usage and Argument section in one file.
* Since we have a *See also* section, it's easy to find other function from the same family.
* The [{pkgdown} index](https://styler.r-lib.org/reference/index.html) already groups relevant functions in the online docs.
* The example section contains now examples from different functions.

Sorry to change my mind after your implementation work. But as I said, I don't think duplication is a problem at the reader side. It's actually desired (Do repeat yourself). Unlike many other doc systems, roxygen allows inheritance, so we can actually generate duplication for the end user's docs without writing docs multiple times.

lorenzwalthert:
If you want, you could change this PR to harmonizing the sections in the different help files, but I can also understand if you don't feel like it 😒

IndrajeetPatil:
No worries, @lorenzwalthert. 

If you think this leads to deterioration in the user experience while reading the docs, then, yes, it's not ideal. Maybe I shouldn't be drawing too close of a parallel with `{lintr}`, and maybe I should also discount my own experience with reading these docs since I am already quite familiar with them.

> you could change this PR to harmonizing the sections in the different help files

I can probably handle this in a separate PR.

lorenzwalthert:
ok, thanks @IndrajeetPatil for being kind with me.

IndrajeetPatil:
Btw, given that I've been contributing and will continue to contribute to this repo, do you think it makes sense to add me as a contributor to the repo, so that-

1. I don't need to work from a fork and keep it up-to-date
2. Label issues
3. Update branches when the need be
4. Re-open old issues in case of a regression, or while revisiting it in the future
5. etc.

I am already a contributor on the `{lintr}` [repo](https://lintr.r-lib.org/authors.html), and the team there can attest to the fact that I am well-behaved 😅

cc @krlmlr

lorenzwalthert:
@IndrajeetPatil I invited you to be a collaborator of {styler}. We are happy to have you and look forward to your next contributions. I plan to reduce my engagement with {styler} in the future, so it's great that you have decided to increase yours.

IndrajeetPatil:
Thanks a lot ☺️

Excited to be part of the team, and looking forward to contributing more!

PR #973, opened 2022-08-07
Quarto: Test that unexecuted blocks aren't styled

IndrajeetPatil:
- https://quarto.org/docs/output-formats/html-code.html#code-overflow

- https://quarto.org/docs/computations/execution-options.html#unexecuted-blocks


lorenzwalthert:
Thanks. I am not sure how code overflow plays a role here. Can you explain? Also I am not sure we never want to style code blocks not intended for running. Are you aware of the code option `tidy = FALSE` for Rmarkdown? styler won't style these chunks, but only if probided in the header like `{r, tidy = FALSE}`, not as a comment like `#| tidy = FALSE` or `#| tidy: False`. The ladder might be another issue and a reason why we should use proper params parsing from knitr, as discussed in https://github.com/yihui/knitr/issues/2112.

lorenzwalthert:
@IndrajeetPatil any thoughts on these remarks?

IndrajeetPatil:
I will create a separate issue for this, and we should discuss this a bit more because it's trickier than I initially thought.

PR #977, opened 2022-08-21
Tweaks to style_pkg()

krlmlr:
- Simplify description thans to new defaults
- Simplify argument documentation
- Avoid styling `R/cpp11.R` by default, even if it's harmless

Follow-up to #965.


lorenzwalthert:
Thanks Kirill. I think as we discussed in #965, changes in default arguments should probably not constitute a breaking change, in particular not excluding files, so I think we can release that with the next non-major release. 

If there are oblivious style issues that are easy to fix in the {cpp11} generated file, can we also fix the generation of that file easily? 

Also, for next time, can you add a news bullet in `NEWS.md`? I can do this time.

krlmlr:
Would you be open to using fledge in this project? It's a lot better now, and we'll never need to think about manually adding stuff to `NEWS.md` again.

The `cpp11.R` file typically looks good, until it doesn't. This might be the reason why nobody noticed or complained. Nothing to fix there AFAICT, still generated files should be excluded. (Should we watch out for a "do not edit" header in the file too?)

I came here because a change in default arguments broke my workflows 🙃

PR #978, opened 2022-08-21
testthat updates

krlmlr:
- parallel testing
- updating golden files

Will double-check locally with non-parallel testing.


lorenzwalthert:
Thanks. 

* I am not sure parallelization is a good idea for tests that assert relative speed, e.g. `tests/testthat/test-cache-high-level-api.R:150` or tests involving `n_times_faster_with_cache()`. Alternative is to run these tests in separate GitHub Actions. Or just see if and observe if parallelisation is causing problems in the future (in this PR, it seems not, but maybe the behaviour is not so stable).
* I am not sure why the trees were outdated. Can you explain why and what you did to update them? FYI I recently also removed them from the built package and reduced size of the built package by 50% (#879).

krlmlr:
We could still run serially in CI/CD?

No idea about the trees. The tests didn't fail when the trees were out of date, they just were updated silently. Could that be the reason?

lorenzwalthert:
Trees are not involved in unit tests, they are just for us to understand the code we want to style I think. So that's fine. 

> We could still run serially in CI/CD?

Let's merge and see if it causes any problems in the future. You can merge if you are read.

Issue #759, opened 2021-03-25
Understand where styler spends most of the time


krlmlr:
This is crucial for #558.

Because we use mostly recursive calls, and effectively each function can call each other function, I suspect it is difficult to see where exactly styler spends most of its time, and which parts perhaps should be optimized. Can you confirm that?

To work around, I propose processing the `.Rprof` files created from a profiler run to adapt to the situation. If we erase the right places from the call stack for each sample, we might just end up with a clean flat structure that immediately shows where the time is spent. What do you think?

Alternatively, we could (temporarily) replace the recursive calls with `later::later()` or replace recursion with a queue and a `while {}` loop. Editing the call stack sounds like less work.

lorenzwalthert:
The recursive calls are time consuming, yes. In particular, there are two important recursions:

* building the nested parse table in `compute_parse_data_nested()`
* apply the transformer rules in `apply_transformers()`

I can't remember correctly but I think they take up similar amount of time. Recently, there were two important speed improvements made (speed change in unseen example code benchmarked with the {here} package). #739 removed unnecessary sorting in `merge()` inside a recursion since we sort right afterwards anyways (-4%) and in #711, transformers that are never used are removed (-10%).

That said, I think your proposed approach sounds interesting. I can confirm that it's hard for me to understand where bottlenecks are. I already spend quite some time optimising R code, maybe not very systematically though. And I think the cost/benefit ratio is too high for me now. I think time is better spent on parallelisation at the moment. I think this release introduced speed improvements of about 25%+ with the few changes described in the [`NEWS.md`](https://styler.r-lib.org/news/index.html#major-changes), but then there were edge case fixes and new features that slowed things down again in this release, leaving us with like 15% diff. Because minor changes were always so hard to measure, I built {touchstone}, so now every PR is benchmarked and we [benchmark against the current release](https://github.com/r-lib/styler/pull/715) as well on merge in master and for all new code, I tried to write is as efficient as possible while keeping it maintainable and understandable. You can find the speed implications under the `comment PR step` in the benchmarking workflow:
![screenshot-pr-comment](https://user-images.githubusercontent.com/10477073/112444737-a7f8d480-8d4e-11eb-8666-9c3905a26573.png)

IndrajeetPatil:
@lorenzwalthert If this is still up for grabs, I would be happy to work on trying to improve performance using C++ (via `{cpp11}` package).

[...]

Issue #558, opened 2019-10-16
Styler is slow


rillig:
Formatting a [real-life example file](https://github.com/NetBSD/pkgsrc/blob/8cc89dcf48ce59d7b83412ab5b2f66d05a849415/pkgtools/R2pkg/files/R2pkg.R#L1) takes 17 seconds on my computer. That's really much because the files is just 1215 lines. Styling this file should take at most 1 second.

lorenzwalthert:
Yes, but you get the follwing styled right:

- https://github.com/yihui/formatR/issues/80#event-2718783090
- `x %>% \nc()`
- `call({{ x }})` (in dev version)
- `call(!!x)` and `call(!!! x)`.
- https://github.com/yihui/formatR/issues/44
- and more

We also have features such as: 

- formatting of roxygen comments.
- formatting of Rnw and Rmd code chunks.

We are working on various features such as: 

- detecting alignment (#258).
- caching (#538).
- turn styler off for some lines (#533).
- multicore processing (#370).

Often there is the trade-off between: 

- fast code
- readable / maintainable code  

We choose an approach that is quite elegant, flexible and customizable (read more [here](https://styler.r-lib.org/articles/customizing_styler.html)). I hope we can cache on top-level expression level in #538, which should mitigate the problem, at least for repeated styling. 

Now the ad block is over :-) 

*** 

Indeed you are right, styler is not fast. Although it used to be twice as slow (#78), it's also bothering me. To identify more bottlenecks, I think we'd need to: 

- make use of profiling (which is quite difficult because of recursive function calls, but I think @krlmlr found a way to do it). 
-  move some code to C or C++, but I am not an expert on that so I'd leave that to others.
- identify other shortcuts from looking at logic in code (for people like me who are very familiar with the codebase).

I can't say much more about that right now, unfortunately. If you want to contribute to any of the above or you have other suggestions, please let me know.

lorenzwalthert:
Anyone interested in doing speed profiling with [proffr](https://github.com/r-prof/proffer)?

lorenzwalthert:
[...]

PR #979, opened 2022-08-21
Optimize visiting code

krlmlr:
For #558 and #759.

This is a tad faster and enables cleaner stack traces, in particular recursive calls now follow each other directly, without an interleaving `map()`. We now can profile, flatten recursive calls, and get a clean picture where time is actually spent.


lorenzwalthert:
Nice, speed benchmarks verify your claim. I am sold 🥳

Issue #937, opened 2022-04-01
convert `%>%` into `|>`?


lorenzwalthert:
{formatR} offers this in https://github.com/yihui/formatR/releases/tag/v1.12. Maybe `tidyverse_style()` could have an argument for that.

sebffischer:
wouldn't that sometimes break code as the operators behave differently?

lorenzwalthert:
It would be optional. And it's aimed at being very simlar with release 2.0: https://www.tidyverse.org/blog/2020/11/magrittr-2-0-is-here/

lorenzwalthert:
Out of scope.

PR #980, opened 2022-08-21
Refactor: better stack traces for profiling

krlmlr:
This allows replacing `%>%` with `|>` which gives much much better stack traces for profiling. Also replaced one call to `map2()` which is very high in the chain, also for clearer stack traces.

For #759.


lorenzwalthert:
Looks good. Also note there is #937.  A limitation of using pre-commit with GitHub Actions is pushes by pre-commit (to fix problems) don't trigger other CI runs. To avoid that, either force push the changes afterwards, create an emtpy bump commit or install pre-commit locally so it's ran on every commit, i.e. on macos, I use ` $ brew install pre-commit` for the system installation and then `$ pre-commit install` in this package directory.

krlmlr:
We want styler to run on R < 4.1, I'm doing the substitution temporarily.

Thanks for the hints on pre-commit. It's unfortunate that automated pushes don't trigger a CI/CD rerun.

lorenzwalthert:
> Thanks for the hints on pre-commit. It's unfortunate that automated pushes don't trigger a CI/CD rerun.

yes. That’s not a pre-commit specific problem I think. And the reason why there is https://pre-commit.ci (but that integration is not (yet) activated for r-lib)

krlmlr:
I don't understand the test failures here.

lorenzwalthert:
That's a speed test that ensure caching works as expected from a performance point of view. It's an outlier, so re-runnig the test again should pass. We could also lift the threshold a bit. Up to you. Happy to merge this.

lorenzwalthert (approved):
Looks good. I can fix failing test later, they are just very specific benchmarks that we want to keep an eye on.

PR #982, opened 2022-08-21
Cleaner stack trace with visitors

krlmlr:
- Ensures that all visited functions have a name
- This name now appears in the stack trace thanks to `make_visit_one()`

We could extend this idea to implementing a `make_visit()` function, `pre_visit()` would be `visitor <- make_visit(pre = TRUE)` then `visitor(pd_nested)` . This would save a few function calls, give more compact code, and also better stack traces for `pre_visit_one()` (which would vanish). The 80-20 rule forbids to do this right now, maybe later. Also, this turns the whole visiting logic into a metaprogram, this is a slippery slope.

For #759.


lorenzwalthert:
Do we need stack traces only for understanding where styler spends more time or do you want to merge that I to the default branch? Ie maybe it’s not necessary to go the slippery slope for the production code and this version of the code can reside in a side branch?

krlmlr:
We *could* go with two parallel files, and activate only one of them depending on some switch. Do you prefer that?

Do I read the benchmark correctly: we're faster when applying the cache, but slower without cache?

lorenzwalthert:
> We could go with two parallel files, and activate only one of them depending on some switch. Do you prefer that?

Well that means more code to maintain. What's exactly the need for clean(er) stack traces than what we have now? If the analysis of the traces is a more or less  one-off thing, then I don't think we should introduce meta programming into the default branch.

lorenzwalthert:
> Do I read the benchmark correctly: we're faster when applying the cache, but slower without cache?

Yes, but you see the confidence interval contains zero, so it's not a significant change. Gray check boxes in {touchstone} just mean no significant deviation in either direction.

krlmlr:
The stack traces also affect error messages as in `rlang::last_trace()`. I think it's a win.

We're talking about the two functions `pre_visit()` and `post_visit()`, only those would be tested in a separate GHA matrix, yes. I don't mind keeping it in a branch for now, as long as the other PRs are good.

krlmlr:
"Not significant" might just mean "we need more data to be sure". I do believe that the meta code takes a tad longer overall.

lorenzwalthert:
> The stack traces also affect error messages as in rlang::last_trace(). I think it's a win.

Our end users should not see these anyways.

lorenzwalthert:
> We're talking about the two functions pre_visit() and post_visit(), only those would be tested in a separate GHA matrix, yes. I don't mind keeping it in a branch for now, as long as the other PRs are good.

Other PRs are good yes, thanks again.

lorenzwalthert:
Yes, you can run {touchstone} for more iterations, then you'll always get a statistically significant result. However, if the change is +0.1% with a confidence interval around it of < 0.001%, it's not really significant from a user experience point of view. So we tried to run for as many iterations that the two significance dimension align (so if user experience is slowed down significantly, it also creates statistical significance). I believe from a speed perspective, these changes are not problematic.

lorenzwalthert:
Seems like we never finished up here. Should we close it?

Issue #581, opened 2020-01-22
Bug 16041 in base R was fixed


lorenzwalthert:
The extra parsing in `get_parse_data()` introduced in #216 is fixed since R 3.5.2 (bug 16041, https://cran.r-project.org/doc/manuals/r-release/NEWS.html), we don't need another extra round of parsing.

krlmlr:
Is it sufficient to search the code for `getRversion()` to find all workarounds?

lorenzwalthert:
Not sure because we also at some point introduced the notion of a parser version, see `parser_version_find()`, which was probably a mistake. We could add a "unit test" using https://github.com/r-hub/rversions to determine the last [four supported versions](https://www.tidyverse.org/blog/2019/04/r-version-support/) and make sure there are no workarounds for package versions smaller than this set.

IndrajeetPatil:
> The workaround in parse_transform_serialize_r() introduced with https://github.com/r-lib/styler/commit/5523a8f1b20c8b26d06dc187d5556cf2abdd1c67 in getRversion() < 3.4 was not required in R 3.4 anymore.

I think this has been already removed, since there is no more `getRversion() < 3.4` in [this file](https://github.com/r-lib/styler/blob/main/R/transform-files.R).

PR #985, opened 2022-08-23
WIP: Bump minimum needed R version to `3.6` (after `R 4.3` release)

IndrajeetPatil:
~~Since R versions `>= 3.4` are supported by `{styler}`.~~

To be considered after `R 4.3` release.

TODO:

- [x] Bump needed R version
- [x] Remove logic related to old parsers
- [ ] Update NEWS 
- [ ] Turn on `oldrel-4` on R CMD Check GHA once `R 4.3` is out


lorenzwalthert:
Thanks @IndrajeetPatil.  Actually we only support the current (4.2), devel and 4 previous versions of R, i.e. 4.1, 4.0, 3.6, 3.5: https://www.tidyverse.org/blog/2019/04/r-version-support/. So I guess instead, the `DESCRIPTION` should be adapted and if we can simplify parser versions or other things, we should do that I think.

IndrajeetPatil:
This [blog entry](https://www.tidyverse.org/blog/2019/04/r-version-support/) mentions that-

> our general policy is to support the current version, the devel version, and four previous versions of R.

So I'd assume that the minimum needed R version is now `R >= 3.5` across `{tidyverse}` packages, but that doesn't seem to be the case.

For example:

- For meta-package `{tidyverse}`, it is `R >= 3.3`:

https://github.com/tidyverse/tidyverse/blob/3be82832a4982d6d10e25e375cedeb8c3161be35/DESCRIPTION#L17-L18

- For `{dplyr}`, it is `R >= 3.4`:

https://github.com/tidyverse/dplyr/blob/73ffdbf29bdb5763b2788c7b41c03a4813c929a8/DESCRIPTION#L30-L31

- For `{lubridate}`, it is `R >= 3.2`:

https://github.com/tidyverse/lubridate/blob/0bb49b21c88736240219dc67e7ed0eb3df15d9b1/DESCRIPTION#L32-L34

etc.

But, if you say, you want to support only `R >= 3.5`, then yes some of the parser-related logic can be simplified.

IndrajeetPatil:
@lorenzwalthert Let's keep this on hold until `R 4.3` release. 

Then we can bump the minimum needed R version to `3.6` and get rid of some logic related to parse version 1.

I have created a new PR #986 where I bump the minimum needed R version to `3.5`.

lorenzwalthert:
Ok, thanks for the analysis. Also, there are some workarounds that can be dropped depending on the R version, mentioned in #581.

R/parse.R
```diff
@@ -111,7 +111,7 @@ get_parse_data <- function(text, include_text = TRUE, ...) {
   pd <- pd %>%
     add_id_and_short()
 
-  parser_version_set(parser_version_find(pd))
+  # parser_version_set(parser_version_find(pd))
```
IndrajeetPatil, line 114: @lorenzwalthert Do we still want to create environment bindings tracking parser version and retain the functions to do so?

Once we bump to `R >= 3.6`, they are no longer used or needed anywhere, so I am wondering if it's worth retaining them.

IndrajeetPatil: WDYT, @lorenzwalthert?

lorenzwalthert: Yes, you are right. I thought we might still need to distinguish because there is also version 3 of the parser but that's not needed anymore.

lorenzwalthert: Question is if we should remove the whole parser versioning logic or not, since in the future, we might still want to distinguish between parser versions, as new changes in the parser might require dispatch.

IndrajeetPatil: Yeah, let's retain this logic even if we don't need it currently, just to be future-proof.

IndrajeetPatil:
The `main` has deviated far away from this PR, and so I would close this for now. 

I will reopen it when `R 4.3` release date is announced.

PR #987, opened 2022-08-24
Error on a `NOTE` in `R CMD Check`

IndrajeetPatil:
Especially to catch the new valid HTML5 checks in R-devel, see:
https://blog.r-project.org/2022/04/08/enhancements-to-html-documentation/


lorenzwalthert:
Interesting 😀. I am not sure I can follow, i.e. why is `_R_CHECK_RD_VALIDATE_RD2HTML_` not set anywhere? I.e. I don't understand how this workflow is any different from what we already have in R CMD Check.

IndrajeetPatil:
It's not different from R CMD Check _per se_, it's just more specialized for one thing only. 

For example, you can already see the same NOTE in R CMD Check, which must have been present for a while but has been completely ignored because the builds don't fail. This won't happen with the HTML-5 workflow.

https://github.com/r-lib/styler/runs/7980143290?check_suite_focus=true#step:6:91

lorenzwalthert:
Ok I see. So why not fail on note in general? This needs to spin up another r cmd check  VM, which costs resources.

IndrajeetPatil:
@lorenzwalthert Sounds good. Done!

lorenzwalthert:
Looks good.

PR #989, opened 2022-08-25
Remove unused internal utilities

IndrajeetPatil:


R/utils.R
```diff
@@ -48,24 +48,13 @@ two_cols_match <- function(col1, col2, data) {
   all(unlist(data[col1]) == unlist(data[col2]))
 }
 
-odd <- function(x) {
```
MichaelChirico, line 51: are odd_index/even_index still used?

IndrajeetPatil: Yes, they are.

MichaelChirico (approved):

R/utils.R
lorenzwalthert: btw I wonder if {lintr} should offer such a check, i.e. if all symbols are used. Does it already?

lorenzwalthert:
Thanks @IndrajeetPatil. Until we use {fledge}, can you please also add a new bullet for this and your previous contributions? I'll merge afterwards.

lorenzwalthert (approved):

lorenzwalthert:
Thanks @IndrajeetPatil for this contribution 🚀 .

PR #994, opened 2022-09-12
Use integer literals and avoid coercions where needed

IndrajeetPatil:
- Use integer literals and avoid coercions where needed
- Following our coding standards

As mentioned [here](https://github.com/r-lib/styler/blob/main/CONTRIBUTING.md):

```
### Control Flow

* Conditional statements should always evaluate to `TRUE` or `FALSE`, i.e. we 
  don't encourage `if (length(x))` but rather `if (length(x) > 0L)`.
* We avoid loops whenever possible and use functions like `purrr::map()` and 
  friends when possible and prefer them over R base counterparts like 
  `base::lapply()`.
```


R/compat-dplyr.R
```diff
@@ -79,10 +79,3 @@ last <- function(x) {
 slice <- function(.data, ...) {
   .data[c(...), , drop = FALSE]
 }
-
-#' @importFrom purrr as_mapper map
-map_dfr <- function(.x, .f, ..., .id = NULL) {
-  .f <- as_mapper(.f, ...)
-  res <- map(.x, .f, ...)
-  bind_rows(res, .id = .id)
-}
```
IndrajeetPatil, line 88: We can use `purrr:: map_dfr()` now.

lorenzwalthert: Does it not require dplyr anymore? That was the reason why we re-implemented it initially.

lorenzwalthert: https://purrr.tidyverse.org/reference/map.html
Ie See. #324.

lorenzwalthert (commented):
Wow thanks @IndrajeetPatil for that hard work. Will review in detail later

R/compat-dplyr.R
IndrajeetPatil: Reverted. I have also added a comment to explain why we are re-implementing this.

lorenzwalthert (approved):
nothing to add. Just a thank you.

PR #995, opened 2022-09-12
Check for perf improvements with `{fastmatch}`

IndrajeetPatil:
Just for checking with continuous benchmarking.


lorenzwalthert:
So good to have touchstone 😊. Although styler is one out of like only 5 repos on GitHub that use it -.-

IndrajeetPatil:
It's an awesome tool indeed! Thanks for both `{touchstone}` and `{precommit}`. 

Very nifty automations to have in one's toolkit.

PR #997, opened 2022-09-13
WIP: Check for performance improvements with C++

IndrajeetPatil:
Related to #558 and #759 

Not converting this to draft because I want touchstone to continuously monitor benchmarking.


lorenzwalthert:
Oh, looking forward to this one. Also, can we exclude `.cpp` files from the wordlist hook? See how other hooks use the `exclude:` key in the `.pre-commit-config.yaml`

lorenzwalthert:
@IndrajeetPatil you will always get a statistically significant result if you increase iterations. Bu the question is is it also significant in terms of domain. And I think a significant change for cache_recording is ~10% improvement. So the idea was to choose the iterations with 30 to make the statistical and domain significance coincide (roughly).

IndrajeetPatil:
At this point, I don't think this is the way to go for improving performance. We should explore all possibilities from R itself, and then move to this option.

lorenzwalthert:
👍

PR #1000, opened 2022-09-13
Get rid of lints with performance implications

IndrajeetPatil:
Some alternatives are more efficient or avoid unnecessary function calls.

### DIFF
--- a/R/detect-alignment.R
+++ b/R/detect-alignment.R
@@ -104,10 +104,10 @@ token_is_on_aligned_line <- function(pd_flat) {
   # now, pd only contains arguments separated by values, ideal for iterating
   # over columns.
   n_cols <- map_int(pd_by_line, ~ sum(.x$token == "','"))
-  previous_line <- 0
-  current_col <- 0
-  start_eval <- ifelse(alignment_col1_all_named(pd_by_line), 1, 2)
-  for (column in seq2(1, max(n_cols))) {
+  previous_line <- 0L
+  current_col <- 0L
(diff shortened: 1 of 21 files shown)


R/detect-alignment.R
```diff
@@ -104,10 +104,10 @@ token_is_on_aligned_line <- function(pd_flat) {
   # now, pd only contains arguments separated by values, ideal for iterating
   # over columns.
   n_cols <- map_int(pd_by_line, ~ sum(.x$token == "','"))
-  previous_line <- 0
-  current_col <- 0
-  start_eval <- ifelse(alignment_col1_all_named(pd_by_line), 1, 2)
-  for (column in seq2(1, max(n_cols))) {
+  previous_line <- 0L
+  current_col <- 0L
```
lorenzwalthert, line 108: I think we could even `x <- y <- 1` no?

lorenzwalthert (approved):
looks good overall, not so much speed improvement unfortunately, but cleaner code 👍 .

R/detect-alignment.R
IndrajeetPatil: Done, thanks.

IndrajeetPatil:
> not so much speed improvement unfortunately

With time, they should accumulate to make a difference 🤞

lorenzwalthert:
> With time, they should accumulate to make a difference 🤞

Yes, for cummulated effects, there is https://github.com/r-lib/styler/pull/868 to compare to last release. Already a few percentage points faster 👍

PR #1002, opened 2022-09-15
Check performance implications with tibble changes

IndrajeetPatil:
With https://github.com/tidyverse/tibble/pull/1368


IndrajeetPatil:
Won't make much of a difference for us since we are mostly using `$` for subsetting, and not `[[`.

PR #1003, opened 2022-09-19
Don't use `nrow` arg in `new_tibble()` calls

IndrajeetPatil:
We are unnecessarily repeating ourselves by doing this.


lorenzwalthert:
Thanks. I think initially you had to supply Tue length j that constructor, maybe not anymore. Is that still wip?

lorenzwalthert (approved):

IndrajeetPatil:
Ah, I marked it WIP only because I didn't see much of a performance improvement so wasn't sure if this is worth a merge.

PR #1006, opened 2022-09-20
Check for performance improvements with `if()` + `else()` instead of `ifelse()`

IndrajeetPatil:
cf. https://github.com/r-lib/styler/issues/759#issuecomment-1252287830


IndrajeetPatil:
The `cache_applying` benchmark does seem to improve - sometimes significant, and sometimes not - but it is always in the right direction here.

lorenzwalthert (approved):
That benchmark is also the least of our concern because applying the cache is already 'instant'. I am ok with the changes. They certainly do not make things worse in terms of performance and readibility is not hurt too much. And the int conversions also make sense.

PR #1007, opened 2022-09-26
Replace tibbles with data frames to improve performance

IndrajeetPatil:
https://github.com/r-lib/styler/issues/759#issuecomment-1256896174

~~Need to use continuous benchmarking, so not converting this to a draft.~~


IndrajeetPatil:
@lorenzwalthert, @krlmlr That's quite the bump in performance when switching to data frames instead of tibbles as our data structure of choice! 😮

lorenzwalthert:
Wow yes @IndrajeetPatil and without much code change even! Well done. Now only hurdle is to make it pass on old releases...

MichaelChirico:
wow, quite impressive speed improvement per LoC change!!

krlmlr (commented):
Nice speedup! Can we encapsulate the choice of data structure in helper functions? We could add e.g. `new_styler_df()` and `styler_df()` that use `vctrs::new_data_frame()` and `vctrs::data_frame()` under the hood. This means that we could later change the underlying data structure with lesser effort.

Do we still need to import tibble?

IndrajeetPatil:
> Nice speedup! Can we encapsulate the choice of data structure in helper functions? We could add e.g. `new_styler_df()` and `styler_df()` that use `vctrs::new_data_frame()` and `vctrs::data_frame()` under the hood. This means that we could later change the underlying data structure with lesser effort.

Good idea. Done!

> Do we still need to import tibble?

Only for `tibble::tribble()`, which we use in a few places. But, if we wish to get rid of `{tibble}` from imports, removing this function's usage should be easy to do. Should I do that?

krlmlr:
Let's leave the `tribble()` call for now.

R/token-define.R
```diff
@@ -1,4 +1,4 @@
-token <- tribble(
+token <- tibble::tribble(
```
lorenzwalthert, line 1: Just realised that we import all of {tibble} in the {styler} namespace. Can you check if we can drop that import and use `::` everywhere (maybe we already do so)?

R/utils.R
```diff
@@ -4,6 +4,14 @@ line_col_names <- function() {
   c("line1", "line2", "col1", "col2")
 }
 
+styler_df <- function(..., .size = NULL, .name_repair = "minimal") {
```
lorenzwalthert, line 10: I am not sure we should use these helpers. They seem to be (too) thin wrappers around {vtrcs} functions without deviating defaults. I'd suggest to just use those directly. What is the rationale to wrap?

IndrajeetPatil: It provides a single encapsulation point that we can replace with tibble once it catches up with base-R in terms of performance.

R/token-define.R
IndrajeetPatil: Done.

R/utils.R
IndrajeetPatil: This was suggested by Kirlll [here](https://github.com/r-lib/styler/pull/1007#pullrequestreview-1120582454).

lorenzwalthert: Ok, if he thinks so I won't oppose :D

lorenzwalthert:
Also, it seems removing {tibble} does not reduce recursive dependencies:

``` r
library(magrittr)
deps <- desc::desc_get_deps() %>%
  dplyr::filter(type == 'Imports') %>%
  dplyr::pull(package)

recursive_deps_before <- purrr::map(deps, ~names(renv:::renv_package_dependencies(.x))) %>%
  unlist() %>%
  unique()


deps_without_tibble <- setdiff(deps, 'tibble')

recursive_deps_after <- purrr::map(deps_without_tibble, ~names(renv:::renv_package_dependencies(.x))) %>%
  unlist() %>%
  unique()


waldo::compare(recursive_deps_before, recursive_deps_after)
#> ✔ No differences
```

<sup>Created on 2022-09-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This is because we use {rematch2}, which in turn depends on {tibble} (but that dependency was suggested to be removed in https://github.com/r-lib/rematch2/pull/14, where @krlmlr was not all in for the suggested implementation. With the additional development that happened over the last 2 years and more recursive dependencies added to tibble, I think it would be even more beneficial to remove that dependency.

lorenzwalthert (approved):
Great job.

IndrajeetPatil:
I think this is a big enough improvement to consider creating a new CRAN release?

IndrajeetPatil:
>  I think this is a big enough improvement to consider creating a new CRAN release?

Any thoughts, @lorenzwalthert and @krlmlr?

We also need to get rid of `NOTE`s in checks: https://cran.r-project.org/web/checks/check_results_styler.html
Let's not wait to get an email about this  😬

lorenzwalthert:
Yes, I agree. Do you want to m make a PR to `main` similar to https://github.com/r-lib/styler/pull/930? If not, I can do it, but not this week. Once all checks green, I can submit it.

IndrajeetPatil:
> If not, I can do it, but not this week. Once all checks green, I can submit it.

@lorenzwalthert I can wait! :)

PR #1009, opened 2022-09-27
Simplify `styler_df()` signature

IndrajeetPatil:
This is generating warnings.

Follow-up on #1007


krlmlr:
What kinds of warning are we seeing here?

IndrajeetPatil:
Strange. I can't reproduce this anymore.

I think the PR is still worth considering, though, for simplifying the wrapper function `styler_df()`. We don't need to provide any other args to it than `...`.

lorenzwalthert (approved):

Issue #1010, opened 2022-09-28
Remove rematch2 and tibble dependencies


krlmlr:
Used only once: https://github.com/r-lib/styler/search?q=rematch2

https://github.com/r-lib/styler/pull/1007#issuecomment-1259086871

PR #1011, opened 2022-09-28
Drop `{rematch2}` dependency

IndrajeetPatil:
Related to #1010

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -29,7 +29,6 @@ Imports:
     magrittr (>= 2.0.0),
     purrr (>= 0.2.3),
     R.cache (>= 0.15.0),
-    rematch2 (>= 2.0.1),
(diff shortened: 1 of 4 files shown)


lorenzwalthert:
Can you relax [the failing test](https://github.com/r-lib/styler/actions/runs/3141460241/jobs/5103922185#step:6:116) a bit? I.e. it does not have to be 5 times faster, 4x is also enough I think.

lorenzwalthert:
Yeah, licence I also checked but both are MIT, so should be fine.

IndrajeetPatil:
> Can you relax [the failing test](https://github.com/r-lib/styler/actions/runs/3141460241/jobs/5103922185#step:6:116) a bit?

Done.

krlmlr (approved):

PR #1012, opened 2022-09-28
Remove `pr_commands` workflow

### LINKED ISSUE #1008: styler always sends an e-mail when commenting on the repo
Every time I interact on the styler repo, I get a rather dull e-mail:

https://screenshot.googleplex.com/69nkbtPXwk42pNs

From poking around a little bit, it's caused by the pr-commands GHA:

https://github.com/r-lib/styler/blob/1f4437bf81c807da713eb1ec78b08e6a78002617/.github/workflows/pr-commands.yaml

(note: `name: Commands` which matches the e-mail).

I haven't quite figured out how to fix this -- should there be a placeholder job that always runs & succeeds immediately? That way when neither of the `/document` and `/style` commands are issued, at least something runs? Seems a bit silly. I haven't been able to find any related issues though.

IndrajeetPatil:
In favour of `precommit` workflow

Closes #1008


IndrajeetPatil:
@lorenzwalthert WDYT?

lorenzwalthert (approved):

PR #1013, opened 2022-09-28
Drop `{tibble}` as a hard dependency

### LINKED ISSUE #1010: Remove rematch2 and tibble dependencies

IndrajeetPatil:
Closes #1010


lorenzwalthert:
Do we need tibble installed to format `tribble()` calls? Not sure. We can parse code that contains library calls for libraries we don't have installed I think.

IndrajeetPatil:
Yeah, I think so too. I was just trying something out to fix failing tests, but it turned out to be due to stringsAsFactors, so I can definitely change this. 

Btw, I can't get precommit to succeed here if I have styler ignore tags, can I?

R/token-define.R
```diff
@@ -1,31 +1,34 @@
...
-  "-",             "math",          "'-'",
-  "*",             "math",          "'*'",
-  "/",             "math",          "'/'",
-  "^",             "math",          "'^'",
-  "~",          "formula",          "'~'",
-  "if",            "cond",           "IF",
-  "else",          "cond",         "ELSE",
-  "in",       "loop_cond",           "IN",
-  "while",    "loop_cond",        "WHILE"
+# styler: off
```
IndrajeetPatil, line 1: @lorenzwalthert This directive is tripping up `precommit`. 

Can I remove it? It will get rid of the neat alignment here, though.

R/token-define.R
```diff
@@ -1,31 +1,32 @@
...
-  "-",             "math",          "'-'",
-  "*",             "math",          "'*'",
-  "/",             "math",          "'/'",
-  "^",             "math",          "'^'",
-  "~",          "formula",          "'~'",
-  "if",            "cond",           "IF",
-  "else",          "cond",         "ELSE",
-  "in",       "loop_cond",           "IN",
-  "while",    "loop_cond",        "WHILE"
+token <- rbind.data.frame(
```
krlmlr, line 1: Another option would be `as.data.frame(matrix(ncol = 3, byrow = TRUE, c(...)))`

krlmlr (approved):

R/token-define.R
IndrajeetPatil: Thanks for the suggestion. But I think I will just retain it like this for now.

R/token-define.R
lorenzwalthert: I don’t think we should remove it without further ado. Maybe open another issue? This should work out of the box.

IndrajeetPatil: Sounds good. I will create an issue.

IndrajeetPatil: Created issue to track this: https://github.com/r-lib/styler/issues/1019

PR #1020, opened 2022-10-01
DO NOT MERGE: Add lint GHA workflow

IndrajeetPatil:


lorenzwalthert:
I am not sure we need this. We already have pre-commit to run lintr...

IndrajeetPatil:
I agree. I wasn't actually planning to merge this; this is why I marked it as a draft PR. Just wanted to check currently present lints, and take it from there. Will close it afterwards.

IndrajeetPatil:
Also, good news is that the touchstone update is working. As expected, benchmarking is not run for this PR :)

lorenzwalthert:
Do you want to re-scope this PR into adding lintr support via pre-commit and a `.lintr` file?

IndrajeetPatil:
Why not just the lint GHA? What's the benefit of including it in the precommit action?

lorenzwalthert:
We would include it in pre-commit and then it would run on commit locally and avoid:
* push by GitHub Action and waiting.
* git history convolution.
* consistency with local interactive {lintr} run that depend on `.lintr`.

PR #1022, opened 2022-10-04
Leave comments on line after `{`

### LINKED ISSUE #1004: Comment positions not preserved when preceded by a brace (`{`)
For example, in `{lintr}`, you can include directives in comments that instruct if the line of code in question should not be linted with a certain linter, etc. But `{styler}` changes positions of such comments, which means these directives will no longer work:

``` r
styler::style_text(
  "T_and_F_symbol_linter <- function() { # nolint: object_name.
  }"
)
#> T_and_F_symbol_linter <- function() {
#>   # nolint: object_name.
#> }
```

<sup>Created on 2022-09-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

lorenzwalthert:
Closes #1004.

### DIFF
--- a/R/rules-line-breaks.R
+++ b/R/rules-line-breaks.R
@@ -157,7 +157,7 @@ style_line_break_around_curly <- function(strict, pd) {
     to_break <- lag(opening_before, default = FALSE) | closing_before
     len_to_break <- sum(to_break)
     pd$lag_newlines[to_break] <- ifelse(rep(strict, len_to_break),
-      1L,
+      ifelse(pd$token[to_break] == "COMMENT", pmin(1L, pd$lag_newlines[to_break]), 1L),
       pmax(1L, pd$lag_newlines[to_break])
     )
   } else {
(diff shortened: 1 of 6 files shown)


IndrajeetPatil:
Yikes, why isn't the benchmarking working?!

Oh, IINM, YAML is case-sensitive, right? So we need to change this to `- "R/**"`?

https://github.com/r-lib/styler/blob/d8283b0232e13d9d333c4fb6f453fb09580c65e0/.github/workflows/touchstone-receive.yaml#L12

lorenzwalthert:
I would assume YAML is case sensitive and parses `"r/**"  to `"r/**"`. Question is more if GitHub Actions syntax is case sensitive.

lorenzwalthert:
@IndrajeetPatil seems to work now with {touchstone}, committed directly to main.

IndrajeetPatil (approved):
LGTM, thanks!

PR #1026, opened 2022-10-08
Use integer literals for indexing

IndrajeetPatil:
Gets rid of `lintr::implicit_integer_linter()` lints.


lorenzwalthert (approved):

lorenzwalthert:
Still amazed how many more you can find all the time 😜

PR #1031, opened 2022-10-16
Run some tests sequentially

### LINKED ISSUE #1027: Speed test fail randomly with {testthat} 3e parallel running
See https://github.com/r-lib/styler/pull/1020 and others for example. #978. Can we split the {testthat} runs up in two parts and run all but the speed tests in parallel and then the speed tests afterwards? Problems were introduced in https://github.com/r-lib/styler/pull/978#issuecomment-1221965506.

lorenzwalthert:
Closes #1027.

### DIFF
--- a/R/testing.R
+++ b/R/testing.R
@@ -380,6 +380,8 @@ test_transformers_drop <- function(transformers) {
 
 
 skip_during_parallel <- function() {
-  is_parallel <- as.logical(as.logical(toupper(Sys.getenv("STYLER_TEST_IS_TRULY_PARALLEL", TRUE))))
-  testthat::skip_if(is_parallel)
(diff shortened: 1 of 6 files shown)


R/testing.R
```diff
@@ -377,3 +377,9 @@ test_transformers_drop <- function(transformers) {
     }
   })
 }
+
+
+skip_during_parallel <- function() {
+  is_parallel <- as.logical(as.logical(toupper(Sys.getenv("STYLER_TEST_IS_TRULY_PARALLEL", TRUE))))
```
IndrajeetPatil, line 383: The outer `as.logical()` is unnecessary?

IndrajeetPatil (approved):
Elegant solution. Thanks!

PR #1037, opened 2022-10-23
Expose internals used with other style guides

### LINKED ISSUE #974: Export of non-exported functions to enable other style guides to pass CRAN checks
Please consider exporting the following functions:

- [ ] is_function_call()
- [ ] previous_non_comment()
- [ ] scope_normalize()

Context: I'm working on {[styler.equals](https://github.com/Robinlovelace/styler.equals)} based on @lorenzwalthert's very handy template: https://github.com/lorenzwalthert/styler.yours

As mentioned in #860 tests do not pass due to these lines: 

[`  if (styler:::is_function_call(pd)) {`](https://github.com/lorenzwalthert/styler.yours/blob/07b6689532aab92b1c60f106f9daf0cbb1fed6e3/R/rules-line-break.R#L11)

https://github.com/lorenzwalthert/styler.yours/blob/07b6689532aab92b1c60f106f9daf0cbb1fed6e3/R/rules-line-break.R#L14

https://github.com/lorenzwalthert/styler.yours/blob/ea4eb50e79886c5bf58dec51dff76aafbcdd4249/R/core.R#L21

@MichaelChirico advocated in https://github.com/r-lib/styler/issues/340#issuecomment-1207569513 to export those functions to make it easier to for people building alternative styles building on this great package to release them to the masses.

Robinlovelace:
Closes #974.

# Progress tracker

- [x] Export `compute_parse_data_nested()` helper
- [x] Export `is_*()` helpers
- [ ] Add examples to new exports
- [ ] Check function and argument names are okay (since these are now user-facing)
- [ ] Code clean-up
- [ ] Code review


IndrajeetPatil (changes requested):
All user-facing functions should also include examples in the documentation.

Robinlovelace:
> All user-facing functions should also include examples in the documentation.

Thanks for the review @IndrajeetPatil. Do you have examples of the kind of situations when these functions would be used? I am not as familiar with the package as you and the other developers so support here would be greatly appreciated.

IndrajeetPatil:
@Robinlovelace Sure! Will have a look either today or tomorrow.

lorenzwalthert:
Please add a roxygen family with `#' @family third-party style guide helpers` (or similar) and list these also separately in the {pkgdwn} index at the bottom with [`has_concept()`](https://pkgdown.r-lib.org/reference/build_reference.html?q=concept#topic-matching).

Robinlovelace:
Thanks for the suggestion @lorenzwalthert, PR just updated. I think I may struggle with the examples so any help on those greatly appreciated.

R/expr-is.R
```diff
@@ -31,8 +31,12 @@ is_while_expr <- function(pd) {
   pd$token[1L] == "WHILE"
 }
 
+
+#' Check whether `pd` is a function call
 #' @describeIn pd_is Checks whether `pd` is a function call.
-#' @keywords internal
+#' @keywords other
```
lorenzwalthert, line 37: I don'nt think we need the `other` keyword do we?

Robinlovelace: Probably not...

NAMESPACE
```diff
@@ -7,6 +7,10 @@ export(cache_deactivate)
 export(cache_info)
 export(create_style_guide)
 export(default_style_guide_attributes)
+export(is_function_call)
```
IndrajeetPatil, line 10: @lorenzwalthert I am not sure if it makes sense to export only `is_function_call()` without exporting its siblings, since they are all documented together and form a family:

- is_curly_expr(pd)
- is_cond_expr(pd)
- is_while_expr(pd)
- is_function_call(pd)
- is_function_dec(pd)
- is_comment(pd)

Should we export the others as well?

R/expr-is.R
```diff
@@ -31,8 +31,21 @@ is_while_expr <- function(pd) {
...
+
+#' Check whether `pd` is a function call
 #' @describeIn pd_is Checks whether `pd` is a function call.
-#' @keywords internal
+#' @family third-party style guide helpers
+#'
+#' @examples
+#' \donttest{
+#' code <- "x <- list(1:3)"
+#' pd <- styler:::compute_parse_data_nested(code)
```
IndrajeetPatil, line 42: @lorenzwalthert Also, creating user-facing examples for these helpers functions requires using `compute_parse_data_nested()`, but `:::` is going to trip R CMD check somewhere. 

Currently, I am sidestepping this with `donttest`, but I'd highly recommend that we export `compute_parse_data_nested()` as well, since it's the workhorse that all helper functions rely upon. WDYT?

NAMESPACE
lorenzwalthert: Ok.

R/expr-is.R
lorenzwalthert: Ok. We can export this as well. I don't see namespace autocompletion conflicts nor that this function would change its API in a breaking way soon.

IndrajeetPatil: Thanks. I will make the necessary changes in this PR itself.

lorenzwalthert:
Note sure we should remove `#' @keywords internal`, as this hides the functions from the package index in the help file browser (see screenshot below). I think since 99% of all users won't need the helpers we export in this PR, shouldn't we hide the functions in the index?
Screenshot 2022-10-25 at 20 51 19

Robinlovelace:
> I think since 99% of all users won't need the helpers we export in this PR, shouldn't we hide the functions in the index?

:+1: from me

IndrajeetPatil:
>  think since 99% of all users won't need the helpers we export in this PR, shouldn't we hide the functions in the index?

@lorenzwalthert I am not sure if I agree. Although they might not be useful for most users, they are exported and should appear in PDF manual, in help pages, in pkgdown website, etc. E.g., most `{dplyr}` users will never need to use `dplyr::ident()`, but docs for this function are included nonetheless. 

With `#' @keywords internal`, the docs will be hidden in the pkgdown website as well. Given that other package developers are also users and need some docs (with examples) to understand how these functions work, they should be included on the website.

We have nothing to lose and developer-users have much to gain by including these functions in the docs.

WDYT?

lorenzwalthert:
ok, thanks for the {dplyr} comparison. I don't think internals have to be hidden from {pkgdwn}, i.e. {styler} internals also have a html page (but no links there currently). I am fine with either way.

IndrajeetPatil:
Sorry, @Robinlovelace, but the merge conflicts here got too much to resolve, so I have just created a new [PR](https://github.com/r-lib/styler/pull/1043), which is cleaner and should be a bit easier to work with.

Closing this PR in favour of that one.

Robinlovelace:
Fine by me. Good luck with it @IndrajeetPatil !

PR #1038, opened 2022-10-24
Correct typos in tests

MichaelChirico:


lorenzwalthert:
Thanks 😊. Another failing speed test to move out of the parallel processing from #1027. Also, we should double check that roxyen is not in the WORDLIST.

IndrajeetPatil (approved):
LGTM, thanks!

PR #1047, opened 2022-11-02
Update contributing guidelines

IndrajeetPatil:
There might be some whitespace changes, so you use the GitHub GUI feature to hide them and focus on the content-related changes.


lorenzwalthert (approved):
looks good, thanks. We might want to also update some parts that are out of date like file structure or how to we use {fledge}, but we can merge these changes already.

PR #1054, opened 2022-11-08
Use integer literals where needed

IndrajeetPatil:
A few I missed out on in previous PRs.


lorenzwalthert:
Agree. Let's activate the pre-commit hook before we introduce more of those.

PR #1060, opened 2022-11-14
Document imports in a single file

IndrajeetPatil:
Note that the `NAMESPACE` file remains unchanged.

This is similar to the [idiom](https://github.com/r-lib/usethis/blob/b9acf67511ffe165d9c42ef7ab9fab7ff5718eac/R/usethis-package.R#L6-L11) utilized by `{usethis}` package.


lorenzwalthert:
Ok. I am not sure that’s the best way though. Just as context: at the very beginning, we imported all objects on a function basis. That’s why there are many duplicated statements. It reminds me now of the Python  import style. The good thing about the duplication is that if a function call looks unknown, you can see quickly check the roxygen comments to know where it comes from. But since we anyways mostly imported functions that developers know, and prefixed the more exotic lines with `::`, I guess this argument is less strong.

lorenzwalthert:
I guess the PR makes sense mostly though. Maybe some of the functions that are just used once and that are more exotic like `as_mapper()` could also be prefixed with `::`, although this might have minor Speedo implications.

lorenzwalthert (approved):
`write.table` and  `capture.output` could be prefixed, these are not used at runtime anyways, only for testing.

R/styler-package.R
```diff
@@ -18,15 +18,24 @@
 #' style_text("a%>%b; a", scope = "tokens")
 "_PACKAGE"
 
-if (getRversion() >= "2.15.1") {
```
IndrajeetPatil, line 21: This conditional is unnecessary since it will always be true.

IndrajeetPatil:
> Maybe some of the functions that are just used once and that are more exotic like as_mapper() could also be prefixed with ::

I have done so now. 

> although this might have minor Speedo implications

Each `::` call costs roughly 120 nanoseconds, so I don't think the performance should degrade in any significant way.

Issue #1058, opened 2022-11-10
Eliminate `ask_to_switch_to_non_default_cache_root_impl`


hadley:
This message is annoying because it crops up randomly while you're doing something else, never at a time where you want to be messing around with cache settings. Given the content of the message, I would suggest simply using `tools::R_user_dir("styler", "cache")` (or equivalent in older R) without prompting.

lorenzwalthert:
Thanks for the suggestion. Let me give a bit of context why this message even appears:
* The problem is that CRAN requires cached files to be removed after 6 days: https://github.com/r-lib/styler/issues/836. {styler} does that, by removing all files in the default caching location older than 6 days from the cache `.onLoad()`.
* if people don't want that, we gave them the option to opt out. Just set the option  `styler.cache_root` to a non-default value.
* Since I don't think most people are aware of this mechanism, we periodically prompt people to opt out of the default or confirm it.

For these reasons, I don't think we should eliminate `ask_to_switch_to_non_default_cache_root_impl`, which is only asked in interactive settings.

hadley:
I'd suggest using a different cleanup mechanism on CRAN or during `R CMD check` so that you are in technical compliance with the CRAN policy, while delivering a positive experience for users.

You have probably set the option, so I suspect you don't appreciate quite how annoying the current random warning is.

lorenzwalthert:
Although I don't agree with all what CRAN requires, the requirement that packages clean up their cache by default seems not completely unreasonable to me, although {styler} does not use disk space (just innodes) with its caching method. So I would not want to bypass this lightly (and it might even be against the CRAN policy). 

I am not sure I understand why you oppose setting the option explicitly, given that it's quite clear from the message how to silence the info. Can you elaborate a bit more on why the barrier is so high?

Also, would it help if the message was only a oneliner (to avoid cluttering the console) pointing to instructions for resolving the issue that can be found in the documentation (e.g. help file)?

hadley:
It's annoying because I always get the message when I'm in the middle of something else. I don't want to drop whatever I'm in the middle of to do something that doesn't seem very important (because "the cache literally takes zero space on your disk"). A one-liner would help, but I really think you'd be better changing the design so that no user intervention was needed at all.

It would be an even higher barrier for most users, since I don't think the average R user knows how to find/edit their `.Rprofile`.

jennybc:
[...]

PR #1063, opened 2022-11-17
Better communication if `styler.cache_root` is not set.

lorenzwalthert:
Addresses #1058.

### DIFF
--- a/R/ui-caching.R
+++ b/R/ui-caching.R
@@ -31,16 +31,16 @@ cache_clear <- function(cache_name = NULL, ask = TRUE) {
 #'
 #' To comply with the CRAN policy, \{styler\} will by default clean up cache files
 #' that are older than 6 days. This implies that you loose the benefit of the cache
-#' for the files not styled in the last 6 days. You can set the R option
-#' `styler.cache_root` explicitly, (e.g. in your `.Rprofile` with
-#' [usethis::edit_r_profile()]) to
+#' for the files not styled in the last 6 days.
 #'
-#' * confirm the default and silence the message with
-#'   `options(styler.cache_root = "styler")`.
+#' If you want to avoid this, i.e., if you want the cache to last longer, you can use the
+#' R option `styler.cache_root` to opt for an indefinitely long-lived cache by setting it to
+#' `options(styler.cache_root = "styler-perm")`.
 #'
-#' * opt for an indefinitely long-living cache by setting the option to something
-#'   else (we recommend `options(styler.cache_root = "styler-perm")`, which is
-#'   also used by pre-commit).
+#' If you are happy with the cache being cleared after 6 days, you can confirm the default and
+#' silence this message by setting it instead to `options(styler.cache_root = "styler")`.
+#'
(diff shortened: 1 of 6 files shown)


R/ui-caching.R
```diff
@@ -23,9 +23,25 @@ cache_clear <- function(cache_name = NULL, ask = TRUE) {
...
+#' for the files not styled in the last 6 days. You can set the R option
+#' `styler.cache_root` explicitly, (e.g. in your `.Rprofile` with
+#' [usethis::edit_r_profile()]) to
+#'
+#' * confirm the default and silence the message with
+#'   `options(styler.cache_root = "styler")`.
+#'
+#' * opt for an indefinitely long-living cache by setting the option to something
+#'   else (we recommend `options(styler.cache_root = "styler-perm")`, which is
+#'   also used by pre-commit).
```
IndrajeetPatil, line 43: How about the following instead?

```r
#' To comply with the CRAN policy, \{styler\} will by default clean up cache files
#' that are older than 6 days. This implies that you loose the benefit of the cache
#' for the files not styled in the last 6 days. 
#'
#' If you want to avoid this, i.e., if you want the cache to last longer, you can use the 
#' R option `styler.cache_root` to opt for an indefinitely long-lived cache by setting it to 
#' `options(styler.cache_root = "styler-perm")`. 
#'
#' If you are happy with the cache being cleared after 6 days, you can confirm the default and 
#' silence this message by setting it instead to `options(styler.cache_root = "styler")`.
#'
#' You can make this change in your `.Rprofile` using [usethis::edit_r_profile()]). 
```

lorenzwalthert: Also fine for me. Where do you see the advantage of your version?

IndrajeetPatil: I think the suggested write-up has a better _flow_: 
there is one paragraph for what to do if you don't accept default, the other paragraph if you do.

Then the tip about `{usethis}` function. No need to mention pre-commit, which most users might not be familiar with.

IndrajeetPatil (approved):
Thanks!

IndrajeetPatil:
Hmm, looks like we might need to add `{usethis}` to `Suggests` to avoid this `NOTE` (which is acceptable by CRAN, btw):

> Package unavailable to check Rd xrefs: ‘usethis’

Either that, we set the following environment variable in our GHA workflow:

```yaml
      - uses: r-lib/actions/check-r-package@v2
        with:
          error-on: '"note"'
        env:
          _R_CHECK_RD_XREFS_: false
```

Which way should we go?

lorenzwalthert:
Well we'll have to fix the note earlier or later, so I think it's fine to fail on note. Rather, we could turn the clickable reference to {usethis} with an unclickable one or add it to `Suggests:`. I think unclickable (and just code font) is fine.

IndrajeetPatil:
>  I think unclickable (and just code font) is fine.

Yes, I would personally prefer this option over acquiring a new soft dependency just so that one function can be clickable in the docs.

PR #1066, opened 2022-11-23
Replace `purrr::when()` with `if()` equivalents

IndrajeetPatil:
This function will be deprecated in the next (1.0.0) `{purrr}` release (https://github.com/tidyverse/purrr/blob/55c9a8ab8788d878ce9e8e80b867139e46d15395/R/deprec-when.R#L57), and so we should remove its usage to prevent user-facing deprecation messages.


lorenzwalthert:
Thanks. I am not sure we should port it since it requires maintenance. Can't we just use `if`, as the docs suggest? Also, when does this require a new release on CRAN to avoid a user facing warning?

IndrajeetPatil:
> Can't we just use `if`, as the docs suggest? 

Done. No longer porting over the function.

> Also, when does this require a new release on CRAN to avoid a user facing warning?

IIRC, `{purrr}` is scheduled for CRAN submission on the 15th of December.

lorenzwalthert:
> IIRC, {purrr} is scheduled for CRAN submission on the 15th of December.

Should we release a new version before that?

lorenzwalthert (approved):

IndrajeetPatil:
> Should we release a new version before that?

Yes, I think that would be a good idea. Maybe at the end of this month? 

Btw, can I upgrade myself from `ctb` to `aut` for the next release? 
I don't know if you have a particular milestone in mind when this should happen.

PR #1068, opened 2022-11-24
Check if all examples run successfully

IndrajeetPatil:
`\donttest` examples are run in [an additional check](https://www.stats.ox.ac.uk/pub/bdr/donttest/) by BDR, and probably `\dontrun` examples are also run in some other checks. 

Therefore, making sure all examples run successfully (or are never run if they are just for illustrative purposes) is a good preventive practice in general. 

This PR does that and also adds a workflow to catch such failures in the future.

### DIFF
--- a/.github/workflows/check-all-examples.yaml
+++ b/.github/workflows/check-all-examples.yaml
@@ -1,0 +1,42 @@
+# Make sure all examples run successfully, even the ones that are not supposed
+# to be run or tested on CRAN machines by default.
+#
+# The examples that fail should use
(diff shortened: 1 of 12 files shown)


lorenzwalthert (approved):
Sure why not. Never heard of these additional checks. Is there documentation for those?

IndrajeetPatil:
Yes, all additional checks can be found [here](https://www.stats.ox.ac.uk/pub/bdr/). There is `README.txt` in each folder that describes the details of that check.

PR #1081, opened 2022-12-24
no pkgdown pre-commit action with GitHub Actions

lorenzwalthert:
We already have the pkgdown action. This should be added to the template in lorenzwalthert/precommit.


IndrajeetPatil (approved):
Agreed. It's a bit redundant in our workflow.

PR #1082, opened 2022-12-26
stylerignore sequences must always be in one block

### LINKED ISSUE #1072: Combination of stylerignore directives and `local()` leads to problematic styling
## Problem

The following code 

``` r
# styler: off
skip_if_not_installed("tibble")
local({
  x <- 1
})
# styler: on
```

is styled to-

```r
# styler: off
local({
  x <- 1
})
skip_if_not_installed("tibble")
local
# styler: on
```

Note how `skip()` and `local` change their positions.

## How to reproduce

Run the example code in interactive mode.
If you are looking for a more realistic example, run `styler:::style_active_file()` on [this](https://github.com/r-lib/lintr/blob/main/tests/testthat/test-implicit_integer_linter.R) file.

## Note

It's interesting that I can reproduce the issue in the minimal example posted above in the _interactive_ mode, but not in a reprex.

Screenshot 2022-11-26 at 08 54 23

lorenzwalthert:
With this PR, styling is correct for a stylerignore sequence that contains a cached expression. The solution involves ensuring that a stylerignore sequence is always processed in one block. Closes #1072. 

``` r
full <- c(
  '# styler: off',
  'skip_if_not_installed("tibble")',
  'local({',
  'x<-1',
  '})',
  '# styler: on'
)
without_ignore <- full[c(-1, -length(full))]
styler:::local_test_setup(cache = TRUE)
styler:::style_text(without_ignore)
#> skip_if_not_installed("tibble")
#> local({
#>   x <- 1
#> })
styler:::style_text(full)
#> # styler: off
#> skip_if_not_installed("tibble")
#> local({
#> x<-1
#> })
#> # styler: on
```

<sup>Created on 2022-12-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


lorenzwalthert:
@IndrajeetPatil maybe you can have a look if you have time. I know the internals are quite complex, so I don't expect you to understand that. Maybe more just if the docs are grammatically understandable and maybe if we should add more tests and if yes which ones.

IndrajeetPatil (approved):
Thanks, @lorenzwalthert! Looks like this must have been a nightmare to debug.

I couldn't follow all the changes in the code, but the comments and the tests confirm that this fixes the original issue. 

I think the current set of tests is enough for now. I already styled a few repos to check for any regressions, and found none.

lorenzwalthert:
Great, thanks @IndrajeetPatil 🥳. To me, this was a major bug and I wonder why no one before found it. Glad we solved it.

PR #1083, opened 2022-12-26
If function declarations are double indented, leave it

### LINKED ISSUE #845: Decide and implement double indent for function definitions
https://github.com/tidyverse/style/pull/174, and in particular the addin of @lionel- to toggle styles.

lorenzwalthert:
Closes #845.


R/rules-indention.R
```diff
@@ -15,17 +15,44 @@ indent_braces <- function(pd, indent_by) {
...
+    }
   }
   pd
 }
 
+#' Is the function declaration double indented?
+#'
+#' Assumes you already checked if it's a function with
+#' `is_function_declaration`. It is double indented if the first token
+#' after the first line break that is a `"SYMBOL_FORMALS"` is
```
IndrajeetPatil, line 38:
```suggestion
#' after the first line break that is a `"SYMBOL_FORMALS"`.
```

IndrajeetPatil:
@lorenzwalthert Overall looks good to me.

I just had one comment. 

Currently, as seen in one of the test cases, the following code

```r
function(a =
           b,
         f =
           d, c =
           3, d =
           4) {

}
```

is styled to

```r
function(a =
           b,
         f =
           d, c =
           3, d =
           4) {

}
```

Shouldn't it instead be styled to?

```r
function(a =
           b,
         f =
           d, 
         c =
           3, 
         d =
           4) {

}
```

Or is the advice of one argument per line [here](https://style.tidyverse.org/functions.html#long-lines-1) relevant only when the arguments are long?

lorenzwalthert:
> Or is the advice of one argument per line [here](https://style.tidyverse.org/functions.html#long-lines-1) relevant only when the arguments are long?

I don't know. Probably it's meant to be one argument per line or all arguments on one line. But I think this has not really to do with the PR directly, so if you care about it, I suggest to open a new issue.

PR #1085, opened 2022-12-26
Work around empty example lines

### LINKED ISSUE #1084: Blank lines for `@examples` are transformed to blank roxygen2 comments
``` r
styler::style_text("
#' @export
#' @examples
#' a + b
#'
fun <- function() {
}
")
#> #' @export
#> #' @examples
#> #' a + b
#> #'
#> fun <- function() {
#> }
```

<sup>Created on 2022-12-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Introduced in 2b886c42e42108241d2f55a80705135fce807533.

krlmlr:
Closes #1084.

### DIFF
--- a/R/roxygen-examples.R
+++ b/R/roxygen-examples.R
@@ -24,6 +24,9 @@ style_roxygen_code_example <- function(example, transformers, base_indention) {
 style_roxygen_code_example_one <- function(example_one,
                                            transformers,
                                            base_indention) {
+  # Workaround for imperfect parsing of roxygen2 examples
+  example_one <- grep("^ *#+'", example_one, value = TRUE)
+
   bare <- parse_roxygen(example_one)
   one_dont <- split(bare$text, factor(cumsum(bare$text %in% dont_keywords())))
   unmasked <- map(one_dont, style_roxygen_code_example_segment,
--- a/R/roxygen-examples.R
+++ b/R/roxygen-examples.R
@@ -25,7 +25,7 @@ style_roxygen_code_example_one <- function(example_one,
                                            transformers,
                                            base_indention) {
   # Workaround for imperfect parsing of roxygen2 examples
-  example_one <- grep("^ *#+'", example_one, value = TRUE)
+  example_one <- example_one[example_one != ""]
 
   bare <- parse_roxygen(example_one)
   one_dont <- split(bare$text, factor(cumsum(bare$text %in% dont_keywords())))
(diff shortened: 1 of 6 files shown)


krlmlr:
Output has changed for two existing tests, but I believe the new behavior is correct.

Empty lines and comments inside examples are a hassle, would it be a big stretch if we ignored such examples entirely (except when last lines are empty)?

lorenzwalthert (approved):

lorenzwalthert:
> Empty lines and comments inside examples are a hassle, would it be a big stretch if we ignored such examples entirely (except when last lines are empty)?

I think I don't think styler should remove blank lines here. 

```r
#' @export
#' @examples
#' a + b
#'
#' # here I put a comment
#' 2 + 1
#'
NULL
```

Why should the styling of previousl lines depend on the last lines being empty?

krlmlr:
I'm referring to the following case:

```r
#' @export
#' @examples
#' a + b
#'
# here I put a comment
#' 2 + 1
#'
NULL
```

What should we do with that comment? Perhaps a warning 😱 that we can't process this, with accurate file+line information?

IndrajeetPatil:
Yeah, this is definitely a tricky case.

The clearest usage of regular comments I can see with the Roxygen blocks is to add annotations to ignore some code from being styled. For example, if I have a few examples in the docs, maybe I want some to be styled, while others not because I have custom formatting ([example](https://github.com/r-lib/lintr/blob/08098ef77adb1eccdf18f8b44a7053f36297df28/R/lint.R#L32-L34) in the wild).

For example:

``` r
"#' @export
#' @examples
#' 
#' # example code I want styled
#' x = 1
#' y = 2
#' 
#example code I want styler to skip to preserve alignment
# styler: off
#' a   = 1
#' ab  = 2
#' abc = 3
#styler: on
NULL" -> code

styler::style_text(code)
#> #' @export
#> #' @examples
#> #'
#> #' # example code I want styled
#> #' x <- 1
#> #' y <- 2
#> #'
#> # example code I want styler to skip to preserve alignment
#> # styler: off
#> #' a   = 1
#> #' ab  = 2
#> #' abc = 3
#> # styler: on
#> NULL
```

<sup>Created on 2022-12-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


As can be seen, currently, `{styler}` works as expected: 
it styles the comments (note the newly added whitespace between `#` and `example`), and
it skips the code in `@examples` that is not expected to be styled.

IINM, if we stop processing the embedded comments, we will stop supporting this use case, right?

PR #1086, opened 2022-12-26
Link rot Check in PRs and schedule only

lorenzwalthert:
Let’s save a bit of energy.


IndrajeetPatil (approved):
SGTM. 

This should work as long as we don't push directly to the `main`, which we already don't and so should be fine.

PR #1091, opened 2023-01-16
add `qmd` as supported value to `filetype` parameter

dpprdan:
"qmd" was missing as a supported value from the filetype parameter docs. 

I did not `roxygenise()`, yet.


IndrajeetPatil (approved):
Not sure how we missed updating this one. Glad you caught it.

Thanks!

PR #1103, opened 2023-04-01
Replace usage of deprecated `rlang::with_handlers()`

### LINKED ISSUE #1102: Replace deprecated `rlang::with_handlers()`
cf. https://rlang.r-lib.org/reference/with_handlers.html

This is generating a ton of warnings when tests are run.

IndrajeetPatil:
Use the suggested replacement instead.

Closes #1102


IndrajeetPatil:
Hmm, the slowdown is real, but not sure how to avoid it since it's coming from upstream.

lorenzwalthert:
Yes. That’s significant. Let’s report that upstream and see if they want to do something about it.

IndrajeetPatil:
> Yes. That’s significant. Let’s report that upstream and see if they want to do something about it.

Do you want to create an issue upstream?

lorenzwalthert:
The benchmark in the other PR suggested that moving the checkpoint forward in time did not affect the speed of the execution. Although the {rlang} > 1.0.0 requirement you add in this PR might lead to a different dependency resolution than in #1104. Hence, my suggestion is to require {rlang} > 1.0.0 in a separate PR and see if that has an effect. My benchmark posted in r-lib/rlang actually indicates that the `try_fetch()` is much faster than `with_handlers()`: https://github.com/r-lib/rlang/issues/1612

lionel-:
Is there some backtrace capture in the error handlers? `try_fetch()` preserves more stack than `tryCatch()` or `with_handers()`. This makes the backtraces more informative but it can be slowed when done with `rlang::trace_back()` or `rlang::abort()`.

If you call `abort()` from a handler, `try_fetch()` is definitely an improvement though. In addition to deeper backtraces this allows `recover()` to reach deeper (at the error site instead of the handling site).

lorenzwalthert:
cc: @IndrajeetPatil.

lorenzwalthert:
After some digging, I think this is a {touchstone} issue. 
* When running the benchmarking script, code below, locally (after `touchstone::activate()`), `with_handlers()` takes 2x as long as the new `try_fetch()`.
```r
benchmark_run(
  expr_before_benchmark = {
    library(styler)
    cache_activate(gert::git_branch())
  },
  cache_applying = style_pkg("touchstone/sources/here", filetype = c("R", "rmd")),
  n = 30
)
```
* When peeling off the {touchstone} layers and just run `system.time(style_pkg("touchstone/sources/here", filetype = c("R", "rmd")))`, the new branch is 2x faster. Not sure what's going on here since the benchmark is really just the expression of interest and setting `iterations > 1` did not make things better. Here's the relevant soure code from {touchstone}
```r
    callr::r(
      function(expr_before_benchmark, dots, branch, block, iteration, asset_dirs) {
        withr::local_namespace("touchstone")
        withr::local_options(asset_dirs)
        eval(expr_before_benchmark)
        benchmark <- bench::mark(eval(dots[[1]]), memory = FALSE, iterations = 1)
        benchmark_write(benchmark, names(dots), branch = branch, block = block, iteration = iteration)
        benchmark
      },
      args = append(args, lst(iteration)),
      libpath = c(libpath_touchstone(branch), .libPaths())
    )

```

cc: @assignUser.

lorenzwalthert:
I tried to run the exact same expression in the console that is ran by {touchstone}.

For current `main`:
```r
bench::mark(
  style_pkg("touchstone/sources/here", filetype = c("R", "rmd")), 
  iterations = 1, memory = FALSE
)
#> Warning message:
#>   Some expressions had a GC in every iteration; so filtering is disabled. 
````
By default, `filter_gc` is `TRUE`, which means, according to docs: *remove iterations that contained at least one garbage collection before summarizing. If TRUE but an expression had a garbage collection in every iteration, filtering is disabled, with a warning.*

On the other hand, `rlang-with-handlers` never seemed to trigger the above warning. 

The new branch `rlang-with-handlers` is about 2x as fast as `main` interactively, but non-interactively it's the other way round. I really think this is just an artifact related to gc, so let's not bother too much since the new version should be clearly faster and the diff is very simple.

lorenzwalthert (approved):

PR #1107, opened 2023-04-05
Check benchmarks only with rlang bump

IndrajeetPatil:


lorenzwalthert (approved):

lorenzwalthert:
Not 100% sure we also use rlang 1.1.0 in the benchmarks. Let's wait and see for the logs.

PR #1111, opened 2023-04-07
Add a script for generating many roxygen-examples-complete tests, apply it

### LINKED ISSUE #1105: [Testing] Poor parallelism of test-roxygen-examples-complete.R

MichaelChirico:
Closes #1105. Note that this includes #1110 so the diff is too large for now. I couldn't figure out a way to specify that commit as the diffbase while still creating this PR on r-lib/styler (since that commit is on my fork). Sorry for the visual noise.


.dev/gen_roxygen_tests.R
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/Rscript
+# execute from package TLD: ./.dev/gen_roxygen_tests.R
+all_examples <- list.files("tests/testthat/roxygen-examples-complete", full.names = TRUE)
+unique_tests <- unique(sub("-.*", "", basename(all_examples)))
+
+for (test in unique_tests) {
+  out_file <- file.path("tests", "testthat", paste0("test-roxygen-examples-complete-", test, ".R"))
+  extras <- if (test == "15") ', scope = "spaces"' else ""
```
IndrajeetPatil, line 8: Can we avoid magic string here? What's special about `"15"`?

.dev/gen_roxygen_tests.R
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/Rscript
+# execute from package TLD: ./.dev/gen_roxygen_tests.R
+all_examples <- list.files("tests/testthat/roxygen-examples-complete", full.names = TRUE)
+unique_tests <- unique(sub("-.*", "", basename(all_examples)))
+
```
IndrajeetPatil, line 5:
```suggestion
test_with_scope <- "15"

```

.dev/gen_roxygen_tests.R
IndrajeetPatil, line 8:
```suggestion
  extras <- if (test == test_with_scope) ', scope = "spaces"' else ""
```

lorenzwalthert:
#1110 is merged, so I am going to drop this from the title.

.dev/gen_roxygen_tests.R
MichaelChirico: Thanks, done!

.dev/gen_roxygen_tests.R
MichaelChirico: PTAL

MichaelChirico: PTAL

MichaelChirico:
Thanks for the ping, feedback handled.

lorenzwalthert:
Thanks. @IndrajeetPatil I leave the review to you, only thing I would add is that I am not a big fan of adding a top level directory (albeit hidden) to the repo. Or do we expect future scripts to live in `.dev`? Can't we just place this script in `testthat/`? IIRC, files that don't start with `test-` or `helper-` are not executed when {testthat} tests are run.

.dev/gen_roxygen_tests.R
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/Rscript
+# execute from package TLD: ./.dev/gen_roxygen_tests.R
```
IndrajeetPatil, line 2:
```suggestion
# execute from package top-level domain: ./.dev/gen_roxygen_tests.R
```

IndrajeetPatil (changes requested):
Thanks, @MichaelChirico! I like the idea of automating the generation of these test files. 

But I agree with Lorenz that it does feel a bit awkward to store the relevant scripts at the top-level of the package directory. I use and see the value of `/.dev`, but I would use it only for tasks that relate to the entire package (e.g. benchmarking, revdep, etc.). I would instead store the relevant scripts in `tests/dev` folder. 

Does that sound reasonable?

IndrajeetPatil (approved):
Thanks!

PR #1114, opened 2023-04-17
Prefer vctrs functions over slower base R equivalents

krlmlr:
for performance.


krlmlr:
And `split()` by `vec_split()` . Both operations are use a lot, the vctrs version is faster by a factor.

krlmlr:
Gone now.

krlmlr:
Finally green.

Force-pushed, with meaningful individual commits.

krlmlr:
Super-green. Greener than green.

lorenzwalthert:
Great @krlmlr. I was wondering: Do we want to replace all / more of `[` with `vec_slice()`? Because I see more cases, some of them (like in `lag()`) get called on every *nest*, so there might still be room for improvement.

krlmlr:
Good catch! `[` on a vector is harmless, but we do gain a lot with data frames. Expecting another boost with the last commit.

lorenzwalthert (approved):
Nice. 🥳

PR #1126, opened 2023-06-01
fix calculation of temp_dir in delete_if_cache_directory()

### LINKED ISSUE #1125: temp_dir in delete_if_cache_directory() does not work as intended, throws cryptic warnings
Recently when loading the `styler` package (usually through some other package importing it), I get this warning:
```
> library(styler)
Warning messages:
1: In normalizePath(Sys.getenv("TMPDIR", Sys.getenv("TMP"))) :
  path[1]="": No such file or directory
2: In normalizePath(Sys.getenv("TMPDIR", Sys.getenv("TMP"))) :
  path[1]="": No such file or directory
3: In normalizePath(Sys.getenv("TMPDIR", Sys.getenv("TMP"))) :
  path[1]="": No such file or directory
```
which comes down to the `TMPDIR` environment variable being undefined in my setup.  That is annoying.
But this code https://github.com/r-lib/styler/blob/aeaa1faef2e5ea11be26817d8cfa3774ad9d66e7/R/zzz.R#L41 has some problems:
- the result of the second `Sys.getenv()` call is bound to the `winslash` argument of `normalizePath()`, and never evaluated as a potential path (notice three warnings for `path[1]` above).
- `tempdir()` also checks the `TEMP` environment variable, which should be a viable option here too
- `normalizePath()` warns on any of the supplied paths not existing, so for any of the tested variables being unset.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
close #1125

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -38,7 +38,7 @@ delete_if_cache_directory <- function(path) {
   }
   designated_cache_path <- normalizePath(tools::R_user_dir("R.cache", which = "cache"))
   is_in_tools_cache <- startsWith(path, designated_cache_path)
-  temp_dir <- normalizePath(Sys.getenv("TMPDIR", Sys.getenv("TMP")))
+  temp_dir <- normalizePath(dirname(tempdir()))
   is_in_generic_cache <- startsWith(path, temp_dir)
   if (is_in_tools_cache || is_in_generic_cache) {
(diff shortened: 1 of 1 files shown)


lorenzwalthert (approved):

lorenzwalthert:
Thanks!

PR #1127, opened 2023-06-02
Use cli messaging for cache

olivroy:
Add a link for RStudio users

### DIFF
--- a/R/set-assert-args.R
+++ b/R/set-assert-args.R
@@ -28,22 +28,22 @@ assert_transformers <- function(transformers) {
     action <- if (utils::packageVersion("styler") >= version_cutoff) {
       "are not supported anymore"
     } else {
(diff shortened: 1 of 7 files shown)


lorenzwalthert:
Looks good, thanks. If you want, you could add links to a few more places:
* full text search for  *help(*
* full text search for *`?*

Also, if you search for *https://github.com/*, you find a few more places where we can use the `.internal` option for error and warnings.

olivroy:
I fixed a test. 
However, I didn't find more places where `.internal = TRUE` would make sense.
The only place was at 
https://github.com/r-lib/styler/blob/e8836fdf5c127848f943396def0dbc2994ed71d5/R/set-assert-args.R#L36,

but it wouldn't really fit in the context.

lorenzwalthert (approved):
LGTM

PR #1129, opened 2023-06-10
avoid warning when missing R cache dir

MichaelChirico:
We have a strange setup where `tools::R_user_dir()` won't be writeable during testing, so `normalizePath()` throws a warning here. The code works as expected after the warning (cache is just deactivated), so setting `mustWork=FALSE` here seems reasonable.

OTOH, this might be pretty unique to our testing systems, so if it's too niche an issue we'll just maintain this patch in our fork.


lorenzwalthert:
> normalizePath() throws a warning here

In my understanding, a warning is given if the path does not exist. Also, in the help file, searching for *writ* does not give any results.

So is it non-existant or non-writable?

MichaelChirico:
> So is it non-existant or non-writable?

I think non-extant because non-writable.

lorenzwalthert:
> I think non-extant because non-writable.

Really? Or the other way round?

MichaelChirico:
Sure, what else would you like to see in the PR before merge?

lorenzwalthert:
> I think non-extant because non-writable.

I don’t understand that. Can you explain?

MichaelChirico:
i.e. the process doesn't have permission to write to that directory, so it can't create the expected file(s).

lorenzwalthert:
Ok thanks. @IndrajeetPatil seems link rot check not working anymore. Can you habe a look?

PR #1131, opened 2023-06-11
Don't require dplyr anywhere

lorenzwalthert:
Superseeds #1128.

### DIFF
--- a/R/compat-dplyr.R
+++ b/R/compat-dplyr.R
@@ -57,12 +57,3 @@ map_dfr <- function(.x, .f, ...) {
   res <- map(.x, .f, ...)
   vec_rbind(!!!res)
 }
-
-mutate <- function(.data, ...) {
-  collected <- list(...)
-  for (idx in seq_along(collected)) {
-    .data[names(collected)[idx]] <- unname(collected[[idx]])
-  }
-
-  .data
-}
--- a/R/compat-dplyr.R
+++ b/R/compat-dplyr.R
@@ -57,3 +57,12 @@ map_dfr <- function(.x, .f, ...) {
   res <- map(.x, .f, ...)
   vec_rbind(!!!res)
 }
+
+mutate <- function(.data, ...) {
+  collected <- list(...)
+  for (idx in seq_along(collected)) {
+    .data[names(collected)[idx]] <- unname(collected[[idx]])
+  }
+
+  .data
+}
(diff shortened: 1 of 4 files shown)


tests/testthat/test-cache-low-level-api.R
```diff
@@ -1,16 +1,16 @@
 test_that("caching utils make right blocks with semi-colon", {
   blocks_simple_uncached <- compute_parse_data_nested(c("1 + 1", "2; 1+1")) %>%
-    dplyr::mutate(is_cached = FALSE) %>%
+    mutate(is_cached = FALSE) %>%
```
MichaelChirico, line 3: where's `%>%` coming from if not dplyr?

MichaelChirico: nvm :)

R/compat-dplyr.R
```diff
@@ -57,3 +57,12 @@ map_dfr <- function(.x, .f, ...) {
   res <- map(.x, .f, ...)
   vec_rbind(!!!res)
 }
+
+mutate <- function(.data, ...) {
```
MichaelChirico, line 61: use base::transform() instead? use case here seems like a drop-in

tests/testthat/test-cache-low-level-api.R
lorenzwalthert: {magritrr}? We'll at some  point transition to `|>` if minimal supported R version has it.

R/compat-dplyr.R
lorenzwalthert: ok sure.

vignettes/customizing_styler.Rmd
```diff
@@ -50,17 +50,17 @@ As the name says, this function removes spaces after the opening parenthesis. Bu
 string_to_format <- "call( 3)"
 pd <- styler:::compute_parse_data_nested(string_to_format) %>%
   styler:::pre_visit_one(default_style_guide_attributes)
-pd$child[[1]] %>%
-  select(token, terminal, text, newlines, spaces)
+
+cols <- c('token', 'terminal', 'text', 'newlines', 'spaces')
+pd$child[[1]][, cols]
```
MichaelChirico, line 55: maybe `subset(pd$child[[1]], select = cols)`?

MichaelChirico: or since we've magrittr:

```r
pd$child[[1]] %>%
  subset(select = c(...))
```

MichaelChirico (approved):
LGTM! Comment is optional/up to preference.

vignettes/customizing_styler.Rmd
lorenzwalthert: I don't know... the audience of the vignette is probably R developers who know `subset()` but I think it's one of these functions that are not used often anymore these days... `]` is not very pretty here I know but probably easier to understand for most.

PR #1135, opened 2023-06-28
Standalone should not be styled

### LINKED ISSUE #1134: standalone files imported from other packages should be ignored in `style_pkg()`
Hi,

I think that standalone files from core tidyverse packages should be ignored from styling.
At the moment, the default values for `exclude_files` is `c("R/RcppExports.R", "R/cpp11.R")`, but I think it could be `c("R/RcppExports.R", "R/cpp11.R", "R/import-standalone-*")`

or a new argument
`exclude_import_standalone` that would default to `TRUE`.

If you are not familiar with the new standalone files, here are some references

https://usethis.r-lib.org/reference/use_standalone.html

https://github.com/r-lib/rlang/blob/main/R/standalone-types-check.R

A PR in dplyr where they use it tidyverse/dplyr#6722. They also use it in ggplot2

In a package, if I use

```r 
usethis::use_standalone("r-lib/rlang", "types-check")
The file will be copied to `mypkg/R/important-standalone-type-check.R`
# it will be read-only.
# but when I run, `styler::style_pkg()`, it will restyle the file.
```

I think that because these files appear as read-only in RStudio. If they are to be styled, they should be styled in the original repos (in this case in rlang)

lorenzwalthert:
Closes #1134.


R/ui-styling.R
```diff
@@ -69,7 +69,7 @@ style_pkg <- function(pkg = ".",
                       style = tidyverse_style,
                       transformers = style(...),
                       filetype = c("R", "Rprofile", "Rmd", "Rmarkdown", "Rnw", "Qmd"),
-                      exclude_files = c("R/RcppExports.R", "R/cpp11.R"),
+                      exclude_files = c("R/RcppExports.R", "R/cpp11.R", "R/import-standalone.+R"),
```
lorenzwalthert, line 72:
```suggestion
                      exclude_files = c("R/RcppExports\\.R", "R/cpp11\\.R", "R/import-standalone.+\\.R"),
```

lorenzwalthert: escape dots for regex.

lorenzwalthert:
Thanks @olivroy. I will review later. `exclude_files` should be regex only (not regex and file path), otherwise it's too ambigous IMO.

lorenzwalthert:
Closed in favor of #1135.

PR #1160, opened 2023-12-03
Delete URL check workflow

IndrajeetPatil:
I am not sure what's special about `{styler}` that this workflow fails. My hunch is that including a lot of GitHub links corresponding to all contributors somehow makes GitHub API calls exceed beyond the limit for this workflow in this repo.

At any rate, this is not so important to merit further investigation. We just need to run URL check before CRAN submission and we should be fine.


IndrajeetPatil:
The precommit workflow is going to fail until https://github.com/r-lib/styler/pull/1158 is merged, and so need to wait for that PR to be merged.

PR #1165, opened 2023-12-04
Clean up unnecessary YAML front matter in README

IndrajeetPatil:
Correct me if I am wrong, but I don't think this is adding much here.


IndrajeetPatil:
> Can you toggle visual mode in RStudio? I think part of it comes from that, at least the line length stuff...

Okay, I've retained the fields relevant for Visual markdown editor.

This PR still needs to wait for #1166 to be merged.

lorenzwalthert (approved):

PR #1166, opened 2023-12-04
Fix new lints about implicit return

IndrajeetPatil:


lorenzwalthert (approved):
Thanks for this refactoring.

PR #1172, opened 2023-12-22
Adding sticker

### LINKED ISSUE #516: A {styler} sticker
Who can contribute one? How can we find candidates? Twitter?

IndrajeetPatil:
Closes #516

- [x] Finalize background color
- [x] Finalize text color
- [x] Build favicons


IndrajeetPatil:
@lorenzwalthert Do you know where the changes in `inst/WORDLIST` are coming from in the precommit workflow? 

I don't know how to prevent this.

olivroy:
I'd suggest to add favicons. 

Locally, you can run `pkgdown::build_favicons()`

IndrajeetPatil:
@olivroy Good idea! 

Need to wait before the sticker is finalized and then I can do this.

lorenzwalthert:
Thanks @IndrajeetPatil and sorry for the delay. Great to have this programatically. I changed to following:
* Adapted to the colours that were in the initial blog post mentioned in https://github.com/r-lib/styler/issues/516#issuecomment-1866645721
* Swapped the suit icon, sorry my reference was wrong.
* excluded png from the spelling hook and re-ran it to not include a gazillion characters from the png.

I am not sure we need another top level directory. Is `hextools` standard? I'd prefer to put it under an existing directory, like `man/figures/` or under `inst/` or other but add it to `.Rbuildignore`.

IndrajeetPatil:
Thanks, @lorenzwalthert! The new colors look good to me. 

Can you please also update the script `hextools/hexsticker.R` for these changes? That way, we can reproduce the sticker in future as well.

> I am not sure we need another top level directory. Is `hextools` standard? I'd prefer to put it under an existing directory, like `man/figures/` or under `inst/` or other but add it to `.Rbuildignore`.

Yes, that'd fine by me.

lorenzwalthert:
Done @IndrajeetPatil.

NAMESPACE
```diff
@@ -31,7 +31,7 @@ export(style_text)
 export(tidyverse_math_token_spacing)
 export(tidyverse_reindention)
 export(tidyverse_style)
-importFrom(magrittr,"%>%")
+importFrom(magrittr,)
```
IndrajeetPatil, line 34: This was committed by precommit, but I am pretty sure it's a bug in the new version of _roxygen2_.

lorenzwalthert: Let’s revert that then, when you file an issue or find one, We can link it to you.

IndrajeetPatil: Never mind. I adapted our code to the new API, so no need to file any issue upstream.

IndrajeetPatil:
@lorenzwalthert This is ready for a merge from my end. Let me know if you have any other comments.

lorenzwalthert (approved):
Thanks so much for keep pushing this forward, it’s the most visible change of {styler} in a while.

IndrajeetPatil:
The sticker isn't showing up on [the devel version of the website](https://styler.r-lib.org/dev/). Looking into it.

lorenzwalthert:
Maybe try build locally and see if that works?

lorenzwalthert:
Also I tried to see the website in incognito mode and I can confirm I can’t see it either.

NAMESPACE
IndrajeetPatil: For future reference, this was the original issue here: https://github.com/r-lib/roxygen2/issues/1570

PR #1176, opened 2024-01-31
Improve error message for scope

olivroy:
Also fixing some errors I introduced. When linking only to a help page, one should use topic, not help.

Also, I was wondering if it was possible to run styler, but only for quotes, i.e. changing `''` to `""`?


R/addins.R
```diff
@@ -68,7 +68,7 @@ style_active_file <- function() {
   } else if (is_r_file) {
     out <- try_transform_as_r_file(context, transformer)
   } else {
-    abort("Can only style .R, .Rmd and .Rnw files.")
+    abort("Can only style .R, .qmd, .Rmd, and .Rnw files.")
```
lorenzwalthert, line 71: Can we sort alphabetically? First thought we should also standardise casing, but seems like quarto does not encourage or even work with `.Qmd`.

R/style-guides.R
```diff
@@ -471,20 +471,19 @@ tidyverse_reindention <- function() {
...
 scope_normalize <- function(scope, name = substitute(scope)) {
   levels <- c("none", "spaces", "indention", "line_breaks", "tokens")
-  if (!all((scope %in% levels))) {
-    abort(paste(
-      "all values in", name, "must be one of the following:",
-      toString(levels)
-    ))
-  }
+  rlang::arg_match(
+    scope,
```
lorenzwalthert, line 483: Can you move `scope` one level up?

R/transform-code.R
```diff
@@ -22,7 +22,7 @@ transform_code <- function(path, fun, ..., dry) {
       ..., dry = dry
     )
   } else {
-    abort(paste(path, "is not an R, Rmd, qmd, or Rnw file"))
+    cli::cli_abort("{.path {path}} is not an R, Rmd, qmd, or Rnw file.")
```
lorenzwalthert, line 25: Same order as above I'd prefer.

lorenzwalthert (changes requested):
Thanks, just minor changes...

R/addins.R
```diff
@@ -98,10 +98,10 @@ save_after_styling_is_active <- function() {
   op_new <- getOption("styler.save_after_styling", default = "")
   if (!is.na(op_old)) {
     rlang::warn(paste(
```
IndrajeetPatil, line 100: Maybe also use `cli::cli_warn()` and the inline Markdown formatting here?

olivroy:
Thanks for your suggestion, I applied them + a couple extra things here and there!

lorenzwalthert (approved):
Thanks a lot @olivroy.

PR #1181, opened 2024-03-18
Adjust to parser failures in r devel

lorenzwalthert:


IndrajeetPatil (approved):
@lorenzwalthert Tests are fixed. PTAL.

lorenzwalthert:
Thought of this solution also. 😄 Then once we all use R 4.4, we can go back and activate the check again...

lorenzwalthert:
Do we still need `upload-results: true`? I just undid that but now not sure if that was a good idea.

IndrajeetPatil:
I find it useful to have the check results available on CI, esp. for OS that I don't have personal access to (Windows).

lorenzwalthert:
ok sorry then let's revert. But for my curiosity: You still needed to locally install R devel? Or did you manage to find the hint in the test results?

lorenzwalthert:
Thanks a lot again @IndrajeetPatil.

IndrajeetPatil:
Actually, I figured out the problem without downloading the devel version locally because the same issue was also present in one of the tests in lintr 😅

MichaelChirico:
> ok sorry then let's revert. But for my curiosity: You still needed to locally install R devel? Or did you manage to find the hint in the test results?

FWIW, in {data.table} we have an Codespace set up that we can easily log into from anywhere to check the package against r-devel:

https://github.com/Rdatatable/data.table/blob/master/.devcontainer/r-devel-gcc/Dockerfile

lorenzwalthert:
Thanks @MichaelChirico. Is that a free plan? Maybe worth looking into, after the last 3 maintenance, I am sure the next parser change requiring a {styler} release is just around the corner.

MichaelChirico:
Hmm I couldn't tell what exactly's free from a quick scan of the codespaces page:

https://github.com/features/codespaces

I get a fairly large amount of free time included with my normal $4/mo GitHub pro subscription. Even barring that, it would be free to run that Docker image locally, the advantage of Codespaces is I can do it in the browser from anywhere.

lorenzwalthert:
I am fine billing the guys who set up the r-lib org 😜. Hehe no, I don’t know if the pricing works per user or per org.

PR #1182, opened 2024-03-21
DO NOT MERGE: Compare speed improvement from initial release

lorenzwalthert:
* [styler 1.0.1](https://github.com/r-lib/styler/releases/tag/v1.0.1): double speed.
* 1.4.0: 15% faster.
* 1.8.0: 40% faster.
* 1.10.0: 2x faster.

100 -> 50 -> 42 -> 25 -> 12: -> rougly 7x faster. than at initial release.


IndrajeetPatil:
Since we have this one open, do we also need to keep #868 open?

lorenzwalthert:
No, it does not work anyways because {enc} is not on CRAN anymore

Issue #1116, opened 2023-05-05
FR: use hyperlink to indicate where styler failed.


olivroy:
I was wondering if it would be possible to have a clickable link to a styler error, using cli.

`cli::cli_text("{.file file.R:2}")`

Creates a link to file.R at line 2.
<https://cli.r-lib.org/reference/links.html?q=file%20lin#line-and-column-numbers>

lorenzwalthert:
I agree this would be handy. Also in the ouptut of `style_pkg()` and the like, where all files are enumerated. Even more generally, having clickable output might be desirable, e.g. also in `styler:::ask_to_switch_to_non_default_cache_root_impl()`'s output and others.

olivroy:
Hi,

I would like to contribute, but I don't really know how. I started looking at styler's source code and CONTRIBUTING.

meanwhile, I opened a PR with cosmetic changes to code (#1122), but nothing significant done.

PR #1186, opened 2024-03-27
Add pluralization + clickable hyperlink support

olivroy:
Addresses part of #1116.

For speed / usefulness, maybe should only hyperlink the files that changed?
Support pluralization + clickable hyperlink

Demo (since #1187 is merged):

image

styler 1.10.2 
image

### DIFF
--- a/R/transform-files.R
+++ b/R/transform-files.R
@@ -22,7 +22,7 @@ transform_files <- function(files,
   max_char <- min(max(nchar(files), 0L), getOption("width"))
   len_files <- length(files)
   if (len_files > 0L && !getOption("styler.quiet", FALSE)) {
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
I don’t get why this change would slow down the touchstone benchmark so much. One reason may be that you had an old version of the base branch. Compared to the previous release, after merging your previous PR, there is no sig officiant slow down. 

https://github.com/r-lib/styler/pull/868#issuecomment-2023830195

so I merged the main branch into your branch now. Let’s see if the issue goes away.

olivroy:
Will see. thanks for checking. Possibly that the repeated `cli::format_inline()` calls slow things down. That's why I propose that maybe only do styling for files that changed?

PR #1187, opened 2024-03-27
Add colored output

olivroy:
Add minimal colored output to log

Chose color for consistency with cli theme

(would work well with hyperlinks #1186 )
image

PR #165, opened 2017-08-28
Communication

lorenzwalthert:
Add informative messages along the way.
```r
style_pkg("tests/testthat/public-api/xyzpackage/")
# styling file R/hello-2.R.                       [DONE]
# styling file R/hello-3.R.                       [DONE]
# styling file R/hello-world.R.                   [DONE]
# styling file tests/testthat/test-package-xyz.R. [DONE]
# No files changed.
```

PR #1189, opened 2024-03-28
tweak width of output

olivroy:
follow-up to #1187 

The +1 seems to cause rendering issues 
It seems like it was added in https://github.com/r-lib/styler/pull/165

I know that cli exports `cli::ansi_nchar()`, but it will probably slow down a little bit the execution to use this.

image

while on main currently, I had
image

Sorry I didn't test this more extensively in the first place.

### DIFF
--- a/R/transform-files.R
+++ b/R/transform-files.R
@@ -54,7 +54,7 @@ transform_file <- function(path,
                            ...,
                            dry) {
   char_after_path <- nchar(message_before) + nchar(path) + 1L
-  max_char_after_message_path <- nchar(message_before) + max_char_path + 1L
+  max_char_after_message_path <- nchar(message_before) + max_char_path
   n_spaces_before_message_after <-
     max_char_after_message_path - char_after_path
   if (!getOption("styler.quiet", FALSE)) {
--- a/tests/testthat/_snaps/public_api-1.md
+++ b/tests/testthat/_snaps/public_api-1.md
@@ -45,3 +45,38 @@
       ----------------------------------------
       Please review the changes carefully!
 
+# messages have the correct width
+
+    Code
+      styled <- style_pkg(testthat_file("public-api", "xyzpackage"))
+    Output
+      Styling  3  files:
+       R/hello-world.R                  v 
+       tests/testthat.R                 v 
+       tests/testthat/test-package-xyz.R v 
+      ----------------------------------------
+      Status	Count	Legend 
+      v 	3	File unchanged.
+      i 	0	File changed.
+      x 	0	Styling threw an error.
+      ----------------------------------------
+
+---
+
+    Code
+      styled <- style_pkg(
+        testthat_file(
+          "public-api",
+          "xyzpackage"))
+    Output
+      Styling  3  files:
+       R/hello-world.R         v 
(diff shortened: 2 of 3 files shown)


lorenzwalthert:
Thanks. Seems not to slow down things. 👍 also, can you add a snapshot test for this output? Seems like we don’t have good coverage of that currently. I think there are already snapshot tests in the `test-public-api-*.R`

lorenzwalthert:
FYI @olivroy I am finish ing the release 1.10.3, so if you want this and https://github.com/r-lib/styler/pull/1187 to be released as part of that (no idea when the next release comes), please make the changes requested in the next 2 days.

olivroy:
Thanks! I will be able to make it tomorrow

olivroy:
> Thanks. Seems not to slow down things. 👍 also, can you add a snapshot test for this output? Seems like we don’t have good coverage of that currently. I think there are already snapshot tests in the `test-public-api-*.R`

Seems like width of output is more difficult to test programatically with snapshots. according to [codecov](https://app.codecov.io/gh/r-lib/styler/blob/main/R%2Ftransform-files.R), the lines I changed are already covered..

lorenzwalthert:
Can you explain what are the hurdles to add a snapshot test? We can set the width of the output programmatically with R option (`options(width = 80)` using {withr} with `local_options()` or similar to ensure consistent width.

olivroy:
> Can you explain what are the hurdles to add a snapshot test? We can set the width of the output programmatically with R option (`options(width = 80)` using {withr} with `local_options()` or similar to ensure consistent width.
> 
> Even when already covered, current tests seem insufficient to detect regressions like the one you introduced.

It is just that RStudio renders inline markup peculiarly. Before RStudio 2023.12, long output could be truncated, so they introduced a new way to render console output that seems to break lines more often than usual, but that doesn't really show in snapshot tests.. Only interactive testing seems to work for this. https://github.com/rstudio/rstudio/issues/13869

olivroy:
See the difference how cli markup is rendered vs regular one
(I am using dev usethis, which uses cli internally now)
image

lorenzwalthert:
With the release candidate,I get 
Screenshot 2024-04-03 at 16 04 45

Note that the checks are vertically aligned. Your snapshot test should show that too (but don't I think).

olivroy:
Ok. I don't really know how to make console + RStudio agree. 
I will send a PR to revert #1187

lorenzwalthert:
> Ok. I don't really know how to make console + RStudio agree.

I get this with the console, so console and RStudio  look similar tome?
Screenshot 2024-04-03 at 16 11 53

olivroy:
I agree my fix was incorrect. The problem is that if you have a path that is wider than the width, all checks marks would go to the next line.. But it is better to revert I guess, not much added value!

lorenzwalthert:
> The problem is that if you have a path that is wider than the width, all checks marks would go to the next line..

I agree, but that problem already exists currently, so I would not consider that blocking a merge of this PR. here both release candidate and your PR

enough wide: 
Screenshot 2024-04-03 at 20 24 00

not enough
Screenshot 2024-04-03 at 20 25 14

PR #1191, opened 2024-04-03
Revert "Add colored output"

olivroy:
Reverts r-lib/styler#1187

### DIFF
--- a/R/communicate.R
+++ b/R/communicate.R
@@ -27,17 +27,14 @@ communicate_summary <- function(changed, ruler_width) {
     cat("Status\tCount\tLegend \n")
     cli::cat_bullet(
       "\t", sum(!changed, na.rm = TRUE), "\tFile unchanged.",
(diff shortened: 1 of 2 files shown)


lorenzwalthert:
Thanks. Let’s wait with merge because I am not yet fully convinced we should. More later…

lorenzwalthert:
@olivroy to follow up, see my last comment https://github.com/r-lib/styler/pull/1189#issuecomment-2035311313. Hence, I'd still consider your contribution valuable. Alternatively, we can think more fundamentally about how the R CLI for {styler} could improve. If you see yourself making contributions there, let's first discuss in an issue and then you can implement them if you want. But no pressure, either merging this or #1189 would also be fine for me.

olivroy:
Ok, yeah I see, maybe #1189 would be fine. Sacrifice a single mis-alignment for the greater good may be a good idea?

PR #1193, opened 2024-04-21
docs: Document package options

### LINKED ISSUE #1185: Document options
It appears that styler has options that can be set and read. These are not documented. Consider documenting them similar to `{cli}` https://cli.r-lib.org/reference/cli-config.html

https://github.com/r-lib/styler/blob/634dce16918c4c8edcafbc2eb0083da04124ee72/R/zzz.R#L8-L16

IndrajeetPatil:
Closes #1185


R/options.R
```diff
@@ -0,0 +1,56 @@
...
+#'
+#' @section Options for the styler package:
+#'
+#' - `styler.addins_style_transformer`: character.
+#'   The name of the style transformer to use in the addins.
+#'   (default: `"styler::tidyverse_style()"`)
+#'
+#' - `styler.cache_name`: character.
+#'   The name of the styler cache to use.
+#'   (default: current styler package version)
```
lorenzwalthert, line 30: We could actually put the version here, since roxygen comments can now contain R code to evaluate.

R/options.R
```diff
@@ -0,0 +1,56 @@
...
+#'
+#' - `styler.cache_root`: character.
+#'   The directory where the cache files are stored.
+#'   (default: `NULL`)
```
lorenzwalthert, line 34: I think we could link `help("caching")` here.

R/options.R
```diff
@@ -0,0 +1,56 @@
...
+#'
+#' - `styler.colored_print.vertical`: logical.
+#'   It decides whether or not the output should be colored with `prettycode::highlight()`.
+#'   (default: `TRUE`)
+#'
+#' - `styler.ignore_alignment`: logical.
+#'   If `TRUE`, alignment, when detected, is ignored.
+#'   (default: `FALSE`)
```
lorenzwalthert, line 42: I think we could link `vignette("detect-alignment")` here.

R/options.R
```diff
@@ -0,0 +1,56 @@
...
+#'
+#' - `styler.ignore_start`, `styler.ignore_stop`: character.
+#'   Regular expressions to ignore lines that match them.
+#'   (default: `"styler: off"` and `"styler: on"`, respectively)
```
lorenzwalthert, line 46: I think we can link `help("stylerignore")` here.

lorenzwalthert (changes requested):
Good start, a few suggestions to improve things further. I think {pkgdown} should pick up these if you stick to the suggested syntax, according to the [docs](https://pkgdown.r-lib.org/articles/linking.html).

lorenzwalthert (approved):

lorenzwalthert:
Thanks. Great addition to our docs.

PR #1195, opened 2024-05-01
Remove blank lines after opening and before closing braces

### LINKED ISSUE #1032: Blank lines in `if` condition expression not removed
I am using the latest GitHub version of `{styler}` (https://github.com/r-lib/styler/commit/38e4b1f5261a49ede102d68dda7c3d7ebb9b5e2b).

## Actual

``` r
styler::style_text('if (
  
  require("logspline") &&
  require("rstanarm")
) {
  print("Hi")
}')
#> if (
#> 
#>   require("logspline") &&
#>     require("rstanarm")
#> ) {
#>   print("Hi")
#> }
```

<sup>Created on 2022-10-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

## Expected

``` r
#> if (require("logspline") &&
#>   require("rstanarm")) {
#>   print("Hi")
#> }
```

IndrajeetPatil:
Closes #1032 

This is only one example. Please check newly added tests to get a feel for this change:

``` r
styler::style_text(text = '(
  
  require("logspline") &&
  require("rstanarm")
  
  
)')
#> (
#>   require("logspline") &&
#>     require("rstanarm")
#> )
```

<sup>Created on 2024-05-01 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


R/rules-spaces.R
```diff
@@ -140,20 +140,34 @@ remove_space_before_opening_paren <- function(pd_flat) {
...
-remove_space_after_opening_paren <- function(pd_flat) {
-  paren_after <- pd_flat$token %in% c("'('", "'['", "LBB")
+remove_space_after_opening_paren <- function(pd_flat, strict) {
+  braces <- c("'('", "'['", "LBB")
+  paren_after <- pd_flat$token %in% braces
   if (!any(paren_after)) {
     return(pd_flat)
   }
+  # remove blank lines after opening braces
+  if (strict) {
```
IndrajeetPatil, line 150: TBH, I don't think we should have `strict` parameter here. Such blank lines should always be removed. But I wanted to keep them in to align with other functions in this family. 

If you disagree, and we do need to support this parameter, then there need to be some additional tests to cover strict set to `FALSE`.

tests/testthat/line_breaks_and_other/assignment-in.R
```diff
@@ -4,8 +4,7 @@ x <-
 
 x <- 3
 
-# FIXME: edge case not working for R < 3.6: Problem: most likely, comment is
```
IndrajeetPatil, line 7: We no longer support these R versions.

lorenzwalthert:
Thanks. Out of experience (and because I produced bugs before), I know that removing line breaks is problematic when there are comments, also see https://styler.r-lib.org/articles/customizing_styler.html#showcasing-the-development-of-a-styling-rule. Hence: Can you modify some tests to cover situations with comments (and adapt the source code if necessary)? Also, the other thing we have to keep in mind is cache invalidation and transformer dropping. Here it seems not that they play a big role. Maybe we can also create developer documentation for that in `CONTRIBUTING.md` elsewhere, because some of it bit me in the foot already in the past multiple times (and there are other things I can't recall here).

IndrajeetPatil:
> Can you modify some tests to cover situations with comments (and adapt the source code if necessary)?

I have added some tests with comments. PTAL.

As for updating contributing guidelines, I think that can be handled in a separate PR, although it's not completely clear to me what needs to be added, but I can give it a go.

R/rules-spaces.R
```diff
@@ -141,19 +141,29 @@ remove_space_before_opening_paren <- function(pd_flat) {
 }
 
 remove_space_after_opening_paren <- function(pd_flat) {
-  paren_after <- pd_flat$token %in% c("'('", "'['", "LBB")
+  braces <- c("'('", "'['", "LBB")
```
lorenzwalthert, line 144: Sorry I did not see that before but you are modifying line breaks in a rule for the scope `spaces`. I don't think that function is the right one to implement that functionality. Mostly since it will break styling for the scope `spaces` and also for readability, this is not optimal. Can you put that rule in a related rule in `R/rules-line-breaks.R` or create a new transformer? The latter makes {styler} slower so it's the fallback if we can't accommodate the rule in an existing transformer.

lorenzwalthert (changes requested):
see comment in `R/rules-spaces.R`.

R/rules-spaces.R
IndrajeetPatil: Thanks. Makes sense. 

I have created two new transformers now, since I don't see any other transformer that easily incorporate these changes.

IndrajeetPatil: The performance impact due to new transformers also doesn't look that dire.

R/rules-spaces.R
```diff
@@ -150,7 +152,8 @@ remove_space_after_opening_paren <- function(pd_flat) {
 }
 
 remove_space_before_closing_paren <- function(pd_flat) {
-  paren_after <- pd_flat$token %in% c("')'", "']'")
+  closing_braces <- c("')'", "']'")
```
IndrajeetPatil, line 155: @lorenzwalthert Is it deliberate to not include `"RBB"` here?

lorenzwalthert: What is `"RBB"` again?

IndrajeetPatil: There is no RBB (corresponding to LBB); I was hallucinating about having seen it 🙈

lorenzwalthert:
Looks good in general, I don't know if we need to boost the testing suite so much though (you are adding 500 lines of tests here)... One concern is that testing time increases (on CRAN, but also during the local development cycle) and also it becomes harder to navigate a big test suite for developers. Can you cut down on a test or two?

R/rules-spaces.R
IndrajeetPatil: For whatever reason, `[[` is LBB, but `]]` is two right brackets 🤷 

``` r
styler:::get_parse_data("x[[1]]")
#>   line1 col1 line2 col2 id parent     token terminal   text pos_id short
#> 1     1    1     1    6 11      0      expr    FALSE x[[1]]      1 x[[1]
#> 2     1    1     1    1  1      3    SYMBOL     TRUE      x      2     x
#> 3     1    1     1    1  3     11      expr    FALSE      x      3     x
#> 4     1    2     1    3  2     11       LBB     TRUE     [[      4    [[
#> 5     1    4     1    4  4      5 NUM_CONST     TRUE      1      5     1
#> 6     1    4     1    4  5     11      expr    FALSE      1      6     1
#> 7     1    5     1    5  6     11       ']'     TRUE      ]      7     ]
#> 8     1    6     1    6 10     11       ']'     TRUE      ]      8     ]
```

<sup>Created on 2024-05-11 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

lorenzwalthert: Sounds like a case for the parser crew to fix...

lorenzwalthert:
Also, can you explain the rationale for making two different functions? I initially was scared it would be a performance problem, but seems not... In general I think we prefer separation of concern (as you did with the implementation of separate functions), but on the other hand having an abstract rule (line breaks around braces) should also correspond to one transformer, so I am not sure we should split it up here.

IndrajeetPatil:
> Looks good in general, I don't know if we need to boost the testing suite so much though (you are adding 500 lines of tests here)... One concern is that testing time increases (on CRAN, but also during the local development cycle) and also it becomes harder to navigate a big test suite for developers. Can you cut down on a test or two?

I am literally reusing the existing tests. How can that be increasing test time? Note that most of the diffs are from data tree. I am not sure why we need them, but they introduce big diffs.

Some quantitative data on the time required to run tests (unit: seconds):

| OS  | `main`  | PR  | 
|---|---|---|
| windows-latest-R  |  49 |  50 | 
| ubuntu-devel-R  | 66  | 65  | 
| mac-latest-R  |  48 | 49  |

IndrajeetPatil:
> Also, can you explain the rationale for making two different functions?

The guard clauses and the braces involved are completely different, and so it doesn't make sense to club them together. Doing so couples two separate concerns: removing leading blank lines and terminal blank lines.

lorenzwalthert:
> I am literally reusing the existing tests. How can that be increasing test time? Note that most of the diffs are from data tree. I am not sure why we need them, but they introduce big diffs.

My bad sorry. Since the tree files are not rendered by default in this repo, I initially did not see they caused a large part of the diff. Let's discuss those in https://github.com/r-lib/styler/issues/1206.

lorenzwalthert:
> The guard clauses and the braces involved are completely different, and so it doesn't make sense to club them together. Doing so couples two separate concerns: removing leading blank lines and terminal blank lines.

I see that they are different, but different is relative, and if you look at the code base, you can see that there are rules like `set_line_break_before_curly_opening` or `set_line_break_around_curly_curly`, which have a rather large scope. The rule you add seems to be about removing redundant blank lines between an opening and closing brace. Many other line break rules are about whether or not there should be a line break, e.g. at the opening of a call, `set_line_break_after_opening_if_call_is_multi_line` governs that:
```r
call(
  many = args
)
# vs 
call(many, 
  args = 2
)
```
By setting `pd$lag_newlines` to `1L` or `0L`, we implicitly remove extra blank lines after the opening `(` and hence this rule handles extra blank lines for that context already. 

In other situations, e.g. for `style_line_break_around_curly`, you will see that `strict` determines whether to set the number of line breaks between tokens to exactly one or if we should keep the number of blank lines if it is already more than one.

These things make me wonder how extra blank lines should be removed. Should there be only one fallback rule (which would be even greater in scope than those two that we have now) to set blank lines in general to at most one consecutive (exceptions to be defined) and other rules should maybe only deal with zero vs one blank line distinction? 

In case you did not notice, the order of transformers can also play a role. You put yours last. To me, we could also put this rule first (hopefully tests would all pass).

I do appreciate this discussion with you, as I think it makes me think and explain the status quo (which is by no means perfect) and hopefully, this ultimately leeds to better code quality.

IndrajeetPatil:
@lorenzwalthert Thanks for the detailed explanation!

Yes, makes sense now, and so I have switched to use only one transformer and I have also put it at the start, and all tests still pass. So this should be ready for another round of review.

Blocked by #1208

vignettes/strict.Rmd
```diff
@@ -58,12 +58,11 @@ function()
 1 ~  more()   #   comment
 ```
 
--   More than one line break is tolerated before closing curly brace and line breaks between curly and round braces are not removed.
+-   Line breaks between curly and round braces are not removed.
```
IndrajeetPatil, line 61: I don't think we should be tolerating such stray empty lines in any mode, strict or non-strict. 

Such line breaks are rarely intentional, and are often the result of code being removed via "search and remove".

tests/testthat/line_breaks_and_other/base-pipe-line-breaks-out.R
```diff
@@ -88,6 +88,14 @@ blew(y = 2, x |>
   a |> c() + 1
 }
 
+{
+  (
+    # some comment
+
```
IndrajeetPatil, line 94: This is expected. The new transformer shouldn't remove line breaks in between lines of code.

.github/workflows/check-full.yaml
```diff
@@ -25,9 +25,9 @@ jobs:
           # use 4.1 to check with rtools40's older compiler
           - { os: windows-latest, r: "4.1" }
 
-          - { os: ubuntu-latest, r: "devel", http-user-agent: "release" }
-          - { os: ubuntu-latest, r: "devel", locale: "en_US" }
+          - { os: ubuntu-latest, r: "devel", locale: "en_US", http-user-agent: "release" }
           #- { os: ubuntu-latest, r: "release", locale: "zh_CN" }
+          - { os: ubuntu-latest, r: "release" }
```
IndrajeetPatil, line 30: This was unintentionally removed in the PR that updated to support difference locales. So adding it back.

.github/workflows/pre-commit.yaml
```diff
@@ -24,19 +24,14 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Install system dependencies
-        if: runner.os == 'Linux'
-        run: |
-          # your system installation code here
-          # sudo apt-get install -y libcurl4-openssl-dev
```
IndrajeetPatil, line 31: This is redundant.

vignettes/strict.Rmd
lorenzwalthert: Maybe yes. Can always open an issue for it, but I wont' work on it 😀

lorenzwalthert (approved):
Great, thanks for incorporating the requested changes.

Issue #1199, opened 2024-05-06
Regression of re-styling of unicode characters


cicdguy:
Hello,

I believe we are seeing a regression of https://github.com/r-lib/styler/issues/847 in R 4.4.0 on Linux. MacOS does not have this issue, and I haven't tested it on Windows.

### Start a shell using the rocker/verse:4.4.0 image

```shell
docker run -it --rm --platform=linux/amd64 rocker/verse:4.4.0 sh
```

### Observe the OS version

```shell
cat /etc/os-release
```

### Install styler from CRAN
```shell
R -s -e 'install.packages("styler", repos = "https://cloud.r-project.org", quiet = T, Ncpus = 8)'
```

### Create a simple file containing unicode characters

```shell
echo 'a <- "R² μ ≥"' > ex.R
```

### Style the file

```shell
R -s -e 'styler::style_file("ex.R")'
```

### Observe the re-styled file

```shell
cat ex.R
```

## Supplemental Information

Running `utils::getParseData(parse(text = 'suit <- "♠"'))` in the container gives me:

```R
  line1 col1 line2 col2 id parent       token terminal           text
7     1    1     1   13  7      0        expr    FALSE
1     1    1     1    4  1      3      SYMBOL     TRUE           suit
3     1    1     1    4  3      7        expr    FALSE
2     1    6     1    7  2      7 LEFT_ASSIGN     TRUE             <-
4     1    9     1   13  4      6   STR_CONST     TRUE "\342\231\240"
6     1    9     1   13  6      7        expr    FALSE
```

But running on my MacOS laptop gives me:

```R
  line1 col1 line2 col2 id parent       token terminal text
7     1    1     1   11  7      0        expr    FALSE
1     1    1     1    4  1      3      SYMBOL     TRUE suit
3     1    1     1    4  3      7        expr    FALSE
2     1    6     1    7  2      7 LEFT_ASSIGN     TRUE   <-
4     1    9     1   11  4      6   STR_CONST     TRUE  "♠"
6     1    9     1   11  6      7        expr    FALSE
```

lorenzwalthert:
Thanks for the good repro. If I am not mistaken, `utils::getParseData()` is the problem here and styler has nothing to do with it? @IndrajeetPatil maybe you can jump in.

IndrajeetPatil:
[...]

PR #1200, opened 2024-05-06
Add non-US locale and use pak to upgrade packages in GHA

IndrajeetPatil:
To check #1199

### DIFF
--- a/.github/workflows/check-full.yaml
+++ b/.github/workflows/check-full.yaml
@@ -12,31 +12,44 @@ jobs:
   R-CMD-check:
     runs-on: ${{ matrix.config.os }}
 
-    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
+    name: ${{ matrix.config.os }} (${{ matrix.config.r }}) ${{ matrix.config.locale }}
 
     strategy:
       fail-fast: false
       matrix:
         config:
-          - {os: macos-latest,   r: 'release'}
+          - { os: macos-latest, r: "release" }
 
-          - {os: windows-latest, r: 'devel'}
-          - {os: windows-latest, r: 'release'}
+          - { os: windows-latest, r: "devel" }
+          - { os: windows-latest, r: "release" }
           # use 4.1 to check with rtools40's older compiler
-          - {os: windows-latest, r: '4.1'}
+          - { os: windows-latest, r: "4.1" }
 
-          - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
-          - {os: ubuntu-latest,   r: 'release'}
-          - {os: ubuntu-latest,   r: 'oldrel-1'}
-          - {os: ubuntu-latest,   r: 'oldrel-2'}
-          - {os: ubuntu-latest,   r: 'oldrel-3'}
-          - {os: ubuntu-latest,   r: 'oldrel-4'}
+          - { os: ubuntu-latest, r: "devel", http-user-agent: "release" }
+          - { os: ubuntu-latest, r: "devel", locale: "en_US" }
+          - { os: ubuntu-latest, r: "release", locale: "zh_CN" }
+          - { os: ubuntu-latest, r: "oldrel-1" }
+          - { os: ubuntu-latest, r: "oldrel-2" }
+          - { os: ubuntu-latest, r: "oldrel-3" }
+          - { os: ubuntu-latest, r: "oldrel-4" }
 
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
       R_KEEP_PKG_SOURCE: yes
 
     steps:
+      - name: Set locale
+        if: matrix.config.locale == 'en_US'
+        run: |
+          sudo locale-gen en_US
+          echo "LC_ALL=en_US" >> $GITHUB_ENV
+
+      - name: Set locale
+        if: matrix.config.locale == 'zh_CN'
+        run: |
+          sudo locale-gen zh_CN
+          echo "LC_ALL=zh_CN" >> $GITHUB_ENV
+
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


.github/workflows/check-full.yaml
```diff
@@ -12,31 +12,44 @@ jobs:
...
       fail-fast: false
       matrix:
         config:
-          - {os: macos-latest,   r: 'release'}
+          - { os: macos-latest, r: "release" }
 
-          - {os: windows-latest, r: 'devel'}
-          - {os: windows-latest, r: 'release'}
+          - { os: windows-latest, r: "devel" }
+          - { os: windows-latest, r: "release" }
```
IndrajeetPatil, line 24: The formatting changes are coming from Prettier formatter in VS code. They are all desirable.

.github/workflows/check-full.yaml
```diff
@@ -49,6 +62,8 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
+          pak-version: devel
+          upgrade: "TRUE"
```
IndrajeetPatil, line 66: This fixes the windows-devel build because pak fetches the latest-stringi.

tests/testthat/test-parsing.R
```diff
@@ -36,10 +36,17 @@ test_that("mixed CRLF / LF EOLs fail", {
   )
 })
 
-test_that("unicode can't be propprely handled on Windows for R < 4.2", {
-  msg <- ifelse(getRversion() < "4.2" && is_windows(),
-    "Can't parse input due to unicode restriction in base R\\.",
-    NA
+test_that("unicode can't be properly handled on Windows for R < 4.2", {
```
IndrajeetPatil, line 39: I think splitting this test into two makes both of them a bit more readable, but I don't feel too strongly about this change. WDYT?

lorenzwalthert:
Still wanna merge this or not necessary anymore?

IndrajeetPatil:
> Still wanna merge this or not necessary anymore?

Yes, this will be still helpful to us in the long term.

.github/workflows/check-full.yaml
```diff
@@ -12,31 +12,44 @@ jobs:
...
 
-          - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
-          - {os: ubuntu-latest,   r: 'release'}
-          - {os: ubuntu-latest,   r: 'oldrel-1'}
-          - {os: ubuntu-latest,   r: 'oldrel-2'}
-          - {os: ubuntu-latest,   r: 'oldrel-3'}
-          - {os: ubuntu-latest,   r: 'oldrel-4'}
+          - { os: ubuntu-latest, r: "devel", http-user-agent: "release" }
+          - { os: ubuntu-latest, r: "devel", locale: "en_US" }
+          #- { os: ubuntu-latest, r: "release", locale: "zh_CN" }
```
IndrajeetPatil, line 30: Some tests fail in this locale, but we can deal with it in a different PR.

.github/workflows/check-full.yaml
lorenzwalthert: Sure.

.github/workflows/check-full.yaml
lorenzwalthert: Ok but I don’t see r devel for windows failing for `main` currently. Is that still necessary?

tests/testthat/test-parsing.R
lorenzwalthert: Good idea yes.

.github/workflows/check-full.yaml
lorenzwalthert: Do we need `http-user-agent: "release"` here?

IndrajeetPatil: This is not something I added; it is present in the example actions itself: https://github.com/r-lib/actions/blob/bc8320839a8d0af1e88cadfa533048e49eeee845/examples/check-full.yaml#L33

.github/workflows/check-full.yaml
```diff
@@ -12,31 +12,44 @@ jobs:
...
+          - { os: ubuntu-latest, r: "oldrel-2" }
+          - { os: ubuntu-latest, r: "oldrel-3" }
+          - { os: ubuntu-latest, r: "oldrel-4" }
 
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
       R_KEEP_PKG_SOURCE: yes
 
     steps:
+      - name: Set locale
```
lorenzwalthert, line 41: Can we have just one step for setting the locale and reference the value from the matrix as an env variable? If not directly, turn the context into a env variable first and then access it? Like this? 
```yaml
      - name: Set locale
        env:
          OVERRIDE_LOCALE: '${{ matrix.config.locale }}'
        run: echo $OVERRIDE_LOCALE >> $GITHUB_ENV

```

.github/workflows/check-full.yaml
lorenzwalthert: I tried to figure out an explanation of what `upgrade:` does and I found 
> upgrade: Whether to install the latest available versions of the dependencies. Must be an R expression. See the README for details if you need quoting. Defaults to FALSE.

So does that mean that we might install source packages but that's better than outdated binaries because of incompatibilities between the latest binary of package A and an outdated binary of package B?

lorenzwalthert (approved):

.github/workflows/check-full.yaml
IndrajeetPatil: Yes, exactly, provided they are available for that particular R version.

.github/workflows/check-full.yaml
IndrajeetPatil: The solutions to DRYing this up was leading to more complicate code, so I am going to retain the more readable individual steps for now.

PR #1207, opened 2024-05-12
Use `all_linters()` instead of using tags

IndrajeetPatil:


.lintr
```diff
@@ -1,13 +1,11 @@
...
-    tags = NULL,
+linters: all_linters(
     commented_code_linter = NULL,
     cyclocomp_linter = cyclocomp_linter(40),
     fixed_regex_linter = NULL,
     function_argument_linter = NULL,
     indentation_linter = NULL,
     line_length_linter(120L),
     namespace_linter = NULL,
-    nested_ifelse_linter = NULL,
```
IndrajeetPatil, line 10: We should be linting this.

.lintr
```diff
@@ -1,13 +1,11 @@
-linters: linters_with_tags(
-    tags = NULL,
+linters: all_linters(
```
IndrajeetPatil, line 1: Just a more intuitive API to access all offered linters.

.lintr
lorenzwalthert: Sure. Wanna add it and fix the lints in this PR?

IndrajeetPatil: The fact that it's not `NULL` means it's already included. There are no new lints to fix. :)

lorenzwalthert: Great 😊. But then I don’t get why you said we should be listing this in the first place 🤔

IndrajeetPatil: When a linter is set to NULL, it is excluded from _all_ linters, and thus not used. So I am just removing its exclusion.

lorenzwalthert (approved):
Thanks.

PR #1209, opened 2024-05-14
Use Bootstrap 5 template for pkgdown website

IndrajeetPatil:
BS3 is being [deprecated](https://github.com/r-lib/pkgdown/pull/2545) in pkgdown.

But, even if that were not the case, the website has a lot to gain from this switch, and not much to lose (I recall you not being happy with the search feature). The BS3 look is quite cumbersome (e.g. try reading [this vignette](https://styler.r-lib.org/dev/articles/customizing_styler.html) on mobile browser).

### DIFF
--- a/_pkgdown.yaml
+++ b/_pkgdown.yaml
@@ -1,9 +1,6 @@
 home:
   strip_header: true
 
-template:
-  bootstrap: 5
-
 reference:
   - title: "Styling API"
     desc: >
@@ -45,11 +42,9 @@ reference:
       - print.vertical
 
 template:
+  bootstrap: 5
   params:
     bootswatch: flatly # https://bootswatch.com/flatly/
-    docsearch:
-      api_key: 13580d327d8a7159f83a7cff178d2141
-      index_name: r-lib_styler
 
 authors:
   Kirill Müller:
--- a/_pkgdown.yaml
+++ b/_pkgdown.yaml
@@ -1,6 +1,9 @@
 home:
   strip_header: true
 
+template:
+  bootstrap: 5
+
 reference:
   - title: "Styling API"
     desc: >


jayhesselberth:
Is docsearch somehow incompatible with BS5 in pkgdown? I don't think that is intended. But maybe you wanted to use the built-in search anyway.

lorenzwalthert:
For reference: https://github.com/r-lib/pkgdown/issues/1960

Indeed I don’t like the default search. Is there another way to do search? What is docsearch?

jayhesselberth:
Got it. Docsearch is Algolia's product: https://docsearch.algolia.com/

Looks like you'll have to choose between BS5 and docsearch. I recommend the former as BS5 will only continue to improve. 

Not sure what it is you don't like about the default search, but the search results from [fuse.js](https://www.fusejs.io/) — the default pkgdown search — could likely be spruced up with some CSS modifications.

lorenzwalthert:
> Looks like you'll have to choose between BS5 and docsearch. I recommend the former as BS5 support will only continue to improve.

Yeah, I guess that's what we have to do. 

> Not sure what it is you don't like about the default search, but the search results from [fuse.js](https://www.fusejs.io/) — the default pkgdown search — could likely be spruced up with some CSS modifications.

Oh, great, are you interested to contribute those for {styler} or upstream in {pgkdown}? :D

lorenzwalthert:
> Not sure what it is you don't like about the default search, but the search results from [fuse.js](https://www.fusejs.io/) — the default pkgdown search — could likely be spruced up with some CSS modifications.

@jayhesselberth, any interest? Otherwise I think we'll merge as is.

jayhesselberth:
I suggest merging as-is. The [next release](https://github.com/r-lib/pkgdown/commit/69f7e6b8db1b9d13234b3ac28bb779bebc332610) of pkgdown will have some major improvements for BS5 sites.

lorenzwalthert (approved):
Ok, then let's merge as is and keep an eye on {pkgdown} releases.

lorenzwalthert:
Also, @IndrajeetPatil, can you please make sure that after the merge, the dev docs display as expected on mobile and desktop?

IndrajeetPatil:
> Also, @IndrajeetPatil, can you please make sure that after the merge, the dev docs display as expected on mobile and desktop?

Yes, will check once the websites are built.

PR #1210, opened 2024-05-14
Update spelling wordlist

IndrajeetPatil:
via `spelling::update_wordlist()`


vignettes/customizing_styler.Rmd
```diff
@@ -44,7 +44,7 @@ We note that there are different types of transformer functions. `initialize` in
 tidyverse_style()$space$remove_space_after_opening_paren
 ```
 
-As the name says, this function removes spaces after the opening parenthesis. But how? Its input is a *nest*. Since the visitor will go through all levels of nesting, we just need a function that can be applied to a *nest*, that is, to a parse table at one level of nesting. We can compute the nested parse table and look at one of the levels of nesting that is interesting for us (more on the data structure in the vignettes "Data structures" and "Manipulating the parse table"):
```
IndrajeetPatil, line 47: AFAICT, these vignettes don't exist.

lorenzwalthert: Agree, these were deleted at some point.

lorenzwalthert:
I did not know about `spelling::update_wordlist()`. Maybe the pre-commit hook should use it instead of a custom logic that only adds words and never removes words when the `--read-only` flag is absent. Interested in a PR there? 😀

lorenzwalthert (approved):
LGTM, thanks.

IndrajeetPatil:
Yes, I can update precommit to use  spelling package in a separate PR.

IndrajeetPatil:
> I did not know about `spelling::update_wordlist()`. Maybe the pre-commit hook should use it instead of a custom logic that only adds words and never removes words when the `--read-only` flag is absent. Interested in a PR there? 😀

I created an issue upstream so that everyone can benefit from these changes, and not just `{styler}`:

https://github.com/lorenzwalthert/precommit/issues/572

Either I or someone else can work on it if you can't.

PR #1212, opened 2024-05-15
Remove unused testing helper: `generate_test_samples()`

IndrajeetPatil:


lorenzwalthert:
Hmmmm... Not sure we actually want to remove these functions. I think Kirill deliberately committed these generators, not sure maybe he thought we'd later change them later to re-generate tests. Hasn't happened though... I guess we can remove them, and come back here in case we need them.

lorenzwalthert (approved):

PR #1216, opened 2024-06-13
Cli links

olivroy:
Progress towards #1116 

## For `style_text()` and the addins
Before
image

This PR:
image

## for `style_file()` and friends

before
image

This PR
image

## for style_dir()

On purpose, I malformed 2 files to have a feel of how many malformed files feel.
It may be too verbose as many lines are there, but I would argue it is unlikely someone has multiple malformed files they want to style

image


IndrajeetPatil (approved):
The trace reads much better. Thanks! :)

PR #1219, opened 2024-06-29
Check for warnings in examples and tests

IndrajeetPatil:
@lorenzwalthert We can either do this and/or add a [new workflow from lintr](https://github.com/r-lib/lintr/blob/main/.github/workflows/test-package-vigilant.yaml) that runs tests in strict mode.

I prefer doing both, but wanted to first check with you.


lorenzwalthert:
I also prefer both 👍.

lorenzwalthert:
Do we really need a separate workflow or can we incorporate it an existing one?

IndrajeetPatil:
> Do we really need a separate workflow or can we incorporate it an existing one? I try to limit compute time.

Not really. I have consolidated two workflows into one to check for warnings in both examples and tests.

lorenzwalthert (approved):
Thanks @IndrajeetPatil.

PR #1220, opened 2024-06-30
[WIP] Use raw strings (from R 4.0)

IndrajeetPatil:


R/roxygen-examples-add-remove.R
```diff
@@ -29,7 +25,7 @@ remove_roxygen_mask <- function(text) {
 #' #' @examples c(1, 2)
 #' @keywords internal
 remove_roxygen_header <- function(text) {
-  gsub("^[\\s\t]*@examples(If)?(\\s|\t)*", "", text, perl = TRUE)
+  gsub(R"(^[\s\t]*@examples(If)?(\s|\t)*)", "", text, perl = TRUE)
```
IndrajeetPatil, line 28: Was this a bug? That is, should this actually have been `\\s|\\t`?

lorenzwalthert: `\s` stands for space, no? So it should capture spaces and tabs.

lorenzwalthert: @IndrajeetPatil are you confident in these topics?

lorenzwalthert:
Why close? Is it not a a good idea?

IndrajeetPatil:
When the tools are ripe enough, I'd much rather an AI agent take care of this than do it manually, since the latter is a more painstaking and error-prone approach.

Issue #1231, opened 2024-10-01
Pick up new implicit and explicit rules from the style guide


krlmlr:
- [ ] https://github.com/tidyverse/style/issues/209

More to come.

IndrajeetPatil:
There are three parts to the "vertical space" item:

- [x] Avoid empty lines at the start or end of functions.
- [ ] Only use a single empty line when needed to separate functions or pipes.
- [ ] It often makes sense to put an empty line before a comment block, to help visually connect the explanation with the code that it applies to.

The third item is really not clear to me. What counts as a comment block? Anything that spans more than a single line?

Issue #1232, opened 2024-10-02
Remove spaces and/or line-breaks from empty curly braces


IndrajeetPatil:
According to cf. https://style.tidyverse.org/syntax.html?q=empty%20curly%20braces#indenting, following 

``` r
library(styler)

style_text(
  "while (waiting_for_something()) {  }"
)
#> while (waiting_for_something()) {  }


style_text(
  "while (waiting_for_something()) {

}"
)
#> while (waiting_for_something()) {
#> 
#> }
```

<sup>Created on 2024-10-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>|

should instead be styled to:

```r
while (waiting_for_something()) {}
```

PR #1234, opened 2024-10-06
empty curly have no spaces around

lorenzwalthert:
Towards https://github.com/r-lib/styler/issues/1231.


tests/testthat/line_breaks_and_other/curly-in.R
```diff
@@ -41,3 +41,9 @@ test_that(
     # comment
     expect_equal(1 + 1, 2)
   })
+
+
+while (TRUE) { }
+
+while (TRUE)
+  { }
```
IndrajeetPatil, line 49: We also need to get rid of line-breaks, if present (cf. https://github.com/r-lib/styler/issues/1232):

``` r
"while (TRUE) { 

}" -> code

styler::style_text(code)
#> while (TRUE) {
#> 
#> }
```

<sup>Created on 2024-10-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

IndrajeetPatil (changes requested):
Thanks. This is a good start. Just one case remains to be covered.

tests/testthat/line_breaks_and_other/curly-in.R
lorenzwalthert: Right…

lorenzwalthert: Ok, but that seems to imply that we always remove the line breaks inside an empy `{}`, which _breaks_ a bunch of other test cases, see e1455c66c50bb0e69ddb08e0ae0290c59df3ffd3... Do you think that's inline with the style guide?

IndrajeetPatil: Yes, I think this is what the style guide dictates, and all the new changes look desirable as per the style guide.

IndrajeetPatil (approved):

lorenzwalthert:
Thanks for the review. 🎉

Issue #1238, opened 2024-11-28
Only use a single empty line when needed to separate functions or pipes


IndrajeetPatil:
cf. https://style.tidyverse.org/syntax.html#vertical-space

## Actual

``` r
library(styler)

"f <- function() NULL


g <- function() f()" -> code

style_text(text = code)
#> f <- function() NULL
#> 
#> 
#> g <- function() f()
```

<sup>Created on 2024-11-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

## Expected

``` r
library(styler)

"f <- function() NULL


g <- function() f()" -> code

style_text(text = code)
#> f <- function() NULL
#> 
#> g <- function() f()
```

PR #1239, opened 2024-11-28
Remove excessive line breaks between top-level expressions

IndrajeetPatil:
Partly related to #1238, but kind of veered away from it during implementation. Opening for an early feedback.

``` r
library(styler)
styler::cache_deactivate()
#> Deactivated cache.

# Example-1 --------------------------------

"f <- function() NULL






g <- function() NULL" -> code

style_text(text = code)
#> f <- function() NULL
#> 
#> 
#> g <- function() NULL

# Example-2 --------------------------------

"while (TRUE) {
  print('hello')
}







for (i in 1:10) {
  print(i)
}" -> code

style_text(text = code)
#> while (TRUE) {
#>   print("hello")
#> }
#> 
#> 
#> for (i in 1:10) {
#>   print(i)
#> }

# Example-3 --------------------------------

"#' @param x A numeric vector
#' @return bla bla bla
f <- function(x) NULL







#' @param y A numeric vector
#' @return bla bla bla
g <- function(y) NULL" -> code

style_text(text = code)
#> #' @param x A numeric vector
#> #' @return bla bla bla
#> f <- function(x) NULL
#> 
#> 
#> #' @param y A numeric vector
#> #' @return bla bla bla
#> g <- function(y) NULL
```

<sup>Created on 2024-11-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/rules-line-breaks.R
```diff
@@ -447,3 +447,28 @@ remove_empty_lines_after_opening_and_before_closing_braces <- function(pd) {
 
   pd
 }
+
+
+#' Reduce multiple blank lines to a maximum number of allowed blank lines
+#' @param pd_flat A flat parse table.
+#' @param allowed_blank_lines The maximum number of allowed blank lines between code elements. Default is `2L`.
+#' @keywords internal
+reduce_extra_blank_lines_between_scopes <- function(pd_flat, allowed_blank_lines = 2L) {
```
IndrajeetPatil, line 456: Style guide recommends one line breaks, but only for function calls. But that itself felt quite aggressive, given the huge numbers of changes in test outputs.

`2L` nicely aligns with Python and feels "natural" as well.

R/rules-line-breaks.R
```diff
@@ -447,3 +447,28 @@ remove_empty_lines_after_opening_and_before_closing_braces <- function(pd) {
...
+  # Calculate the maximum allowed lag_newlines
+  max_lag_newlines <- allowed_blank_lines + 1L # +1 accounts for the line with the previous token
+
+  # Create a copy of lag_newlines to track modifications
+  modified_lag_newlines <- pd_flat$lag_newlines
+
+  # Iterate through the dataframe to reduce consecutive blank lines
+  for (i in seq_along(modified_lag_newlines)) {
```
lorenzwalthert, line 464: I think you can just use `pmax()` here to avoid both loop and if condition.

R/rules-line-breaks.R
```diff
@@ -447,3 +447,28 @@ remove_empty_lines_after_opening_and_before_closing_braces <- function(pd) {
...
+    if (modified_lag_newlines[i] > max_lag_newlines) {
+      modified_lag_newlines[i] <- max_lag_newlines
+    }
+  }
+
+  # Update the original data frame
+  pd_flat$lag_newlines <- modified_lag_newlines
```
lorenzwalthert, line 471: Why not directly assign to `pd_flat`? I think it you use `pmax()` you can assign the whole vector to `pd_flat$lag_newlines` in one assignment.

R/rules-line-breaks.R
```diff
@@ -447,3 +447,28 @@ remove_empty_lines_after_opening_and_before_closing_braces <- function(pd) {
...
+
+  return(pd_flat)
```
lorenzwalthert, line 473: Most rules don't use explicit retrurn (unless early), i.e. just `x` instead of `return(x)` at the end of the function body. Can you adapt?

R/rules-line-breaks.R
```diff
@@ -447,3 +447,14 @@ remove_empty_lines_after_opening_and_before_closing_braces <- function(pd) {
 
   pd
 }
+
+
+reduce_extra_blank_lines_between_scopes <- function(pd, allowed_blank_lines = 2L) {
```
lorenzwalthert, line 452: I believe what you call _scope_ is called top level expressions in this repo (just full text search for it). Can you re-name this function and PR to that effect?

R/rules-line-breaks.R
```diff
@@ -447,3 +447,14 @@ remove_empty_lines_after_opening_and_before_closing_braces <- function(pd) {
...
+  # Calculate the maximum allowed lag_newlines
```
lorenzwalthert, line 453: sory to be so particular but can you 
```
  pd$lag_newlines <- pmin(pd$lag_newlines, allowed_blank_lines + 1L)
```
and remove the comments? If you look at other transformers (i.e. the rule functions), they rarely contain comments, except where something truly _weird_ needs explanation.

lorenzwalthert (changes requested):

R/rules-line-breaks.R
IndrajeetPatil: That's quite clean. Thanks!

IndrajeetPatil:
@lorenzwalthert I am not fully sold on the testing strategy here. Instead of tweaking the existing tests, I think I should create a new folder for "line breaks and top-level expressions" and add all new test cases there. 


WDYT?

lorenzwalthert:
> @lorenzwalthert I am not fully sold on the testing strategy here. Instead of tweaking the existing tests, I think I should create a new folder for "line breaks and top-level expressions" and add all new test cases there.

I am ok with that. Also, I renamed the transformer (again) to make its name more consistent with existing names of line break related transformes.

lorenzwalthert (approved):
Thanks a lot @IndrajeetPatil 🥳

PR #1240, opened 2024-11-30
Update WORDLIST

IndrajeetPatil:
Automated changes via `spelling::update_wordlist()`


IndrajeetPatil:
I am wondering if we should remove `/inst` folder from touchstone receive workflow. I don't see why benchmarking should be run for PRs like these.

lorenzwalthert (approved):

Issue #247, opened 2017-10-22
Enforcing a 80-character line length rule?


jarodmeng:
Is it possible to enforce a rule to limit code to 80 character per line? `styler` currently doesn't seem to do that. The closest issue I can find is #118, but it's not clear if that option is available.

lorenzwalthert:
Thanks for having consulted open issues first. #118 is an edge case as far as indention goes and not quite what you are referring to. The task of splitting lines of code with more than 80 characters is non-trivial as it is not a-priori clear at which position lines should be split. I think this may also not follow a general and abstract rule (as the other styling rules) but is best done manually and decided on a case-to-case basis. Therefore, this feature is currently no supported in styler and will probably no be so in the near feature. We suggest to use the packages lintr to detect lines with more than 80 characters and rearrange manually.

krlmlr:
I'm working on (finally!) releasing utf8, this means we could release styler soon. After this I think we could start thinking about that 80-character rule, because editing code from "wide" to "long" takes quite a lot of time (measured in seconds, but still). I'd like to suggest to do it at the level of function calls and arithmetic expressions first:

- If a single-line function call hits the 80-character boundary, add a line break after the opening paren, and proceed with usual rules.
- If an arithmetic expression is too wide, add line breaks after lowest-prio operators (i.e., the `+` in `1 * 2 + 3 * 4`)

Maybe start with only function calls -- if we add a set of rules after all other rules (because we need to be sure about the width), and then do only reindention of function call arguments?

jarodmeng:
Thanks for the reply! I agree that it's not easy to implement such a rule. @krlmlr 's suggestion looks good to me.

lorenzwalthert:
So do you @krlmlr suggest to take the 80 character limit from the (raw) unstyled expression, right?

krlmlr:
No, that would be after styling, when we know the final width -- we'd just add linebreaks but not change much otherwise.

lorenzwalthert:
Issues that can't be resolved soonish will be closed and labelled "Status: Postponed".

lorenzwalthert:
Reference: #65.

krivit:
@lorenzwalthert , I've implemented a version of this, by adding a loop to `parse_transform_serialize` to do the following:

1. Run through the formatting and flattening process to obtain `flattened_pd`.
2. See if any of the tokens have an end column greater than the limit. If any do, note the first token with this property and add its `pos_id` to a list. If none do, render the text and exit.
[...]

PR #1244, opened 2024-12-06
Introduce line-breaks in multiline arguments

IndrajeetPatil:
cf. https://github.com/tidyverse/style/issues/209

``` r
styler::style_text("
list(
  x, y,
  c(
    a
  ), c(
    b
  )
)
")
#> list(
#>   x, y,
#>   c(
#>     a
#>   ),
#>   c(
#>     b
#>   )
#> )
```

<sup>Created on 2024-12-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


IndrajeetPatil:
Wow, that's quite the slowdown. Need to figure out what's causing it once the scope is clear.

IndrajeetPatil:
@lorenzwalthert Any thoughts? 

I am blocked here without any further feedback from the team.

lorenzwalthert:
Also important: in the example you gave, all line breaks should simply be removed if we had https://github.com/r-lib/styler/issues/247.
Maybe we should talk about the abstract rule first. What is it exactly that we try to achieve?
As for the code, I just wanted to say that you can easily figure out if there are line-breaks in a nest by looking at the respective column. To figure out if there are line breaks _hidden_ in one of the children, consier the `is_multi_line` column. That will save you a lot of computations I think.

IndrajeetPatil:
> What is it exactly that we try to achieve?

This is where I need some clarity. 

I am interpreting the discussion in https://github.com/tidyverse/style/issues/209 to mean that the style guide recommends "one line per argument" rule.

Elsewhere, the style guide carves out an exception to put multiple arguments per line, but the intent of clubbing similar arguments is not something `{styler}` can understand:

Screenshot 2024-12-18 at 08 58 13

Therefore, I opted to just always follow the stated rule and if users wish to retain multiple arguments per like, they will just have to add ignore directives.

lorenzwalthert:
as it is evident from the test cases, enforcing one argument per line is quite a change to the current behaviour I'd discourage this. Also, sometimes you want more than one argument per line [when named arguments are involved](https://github.com/tidyverse/style/issues/39#issuecomment-324352762): 
```
map(x, f
  y = 2
)
```
So I suggest to just focus on the smallest scope first and then maybe extend it gradually. And in my view, a small improvement involves the focus of multi-line arguments and it is to enforce a line-break between two multi-line arguments using the `is_multi_line` argument. 
```r
c(
 c(
  'b'
 ), fun(
  f = 2
 )
)
# becomes 
c(
 c(
  'b'
 ), 
 fun(
  f = 2
 )
)

```
Or even slightly more rigid, each multi-line expression has to be on its own line:
```r
c(
 a, c(
  'b'
 ), fun(
  f = 2
 )
)
# becomes 
c(
 a, 
 c(
  'b'
 ), fun(
  f = 2
 )
)

```
Comments should probably be ignored in that logic, i.e. token after or before comma should be defined as in _skip the comments_. Do you want to adapt your PR to that effect? I can also tell you how I'd solve the puzzle in abstract terms first, but I don't want to take away the fun of figuring it out, so you can go ahead if you want.

IndrajeetPatil:
@lorenzwalthert I am not sure how to fix the performance issues here, and I no longer have time to work on this, at least not in the forseeable future. 

I can either close the PR, or let you take it over from here. Up to you.

lorenzwalthert:
@IndrajeetPatil ok, let's leave it open. I have a different approach in mind, who knows when I'll have time 😇

IndrajeetPatil:
Given the performance regressions, this is not a good approach. I think this requires a new and fresh take on this issue.

PR #1249, opened 2024-12-09
Improve a few function names

IndrajeetPatil:
We are being unnecessarily terse in some places where we don't need to be. I always feel like I am hitting a little speed bump every time I need to decode these names mentally.


lorenzwalthert (approved):
It's not a mental burden for me, but I also wrote it 😄

PR #1290, opened 2026-06-08
Better handling of nested multi-line function definitions, e.g. R6

### LINKED ISSUE #1136: R6 methods don't allow double-indented formals?
The following should be allowed (ref: https://style.tidyverse.org/functions.html#long-lines-1 doesn't mention R6):

```r
R6Class("MyClass",
  public = list(
    param = NULL,
    initialize = function(
      my_long_parameter = getOption("default_long_parameter", 1)) {
      self$param <- my_long_parameter
    }
  )
)
```

But it's currently re-styled (`style_text()` output):

```r
R6Class("MyClass",
  public = list(
    param = NULL,
    initialize = function(my_long_parameter = getOption("default_long_parameter", 1)) {
      self$param <- my_long_parameter
    }
  )
)
```

Which of course creates an over-full line.

MichaelChirico:
Closes #1136.

Written by Gemini.

One ambiguous case was the existing test case:

https://github.com/r-lib/styler/blob/290d9934643ba1d7ecd5d0f23ba455c5b3c85858/tests/testthat/fun_dec/line_break_fun_dec-in.R#L38-L42

Here, it's a bit tough to decide whether to style this as hanging or "open". I nudged the rule to select "hanging" here because (1) `x` and `y` are in the columns they would be for "hanging" and (2) the `)` comes on the same line as `y`. See new test cases for more similar situations.

### DIFF
--- a/R/rules-indention.R
+++ b/R/rules-indention.R
@@ -58,12 +58,20 @@ is_single_indent_function_declaration <- function(pd, indent_by = 2L) {
   if (length(formals) == 0L) return(FALSE)
 
   first_formal_idx <- formals[1L]
-  if (!any(pd$lag_newlines[seq2(idx_paren_open + 1L, first_formal_idx)] > 0L)) {
-    return(FALSE)
+  if (any(pd$lag_newlines[seq2(idx_paren_open + 1L, first_formal_idx)] > 0L)) {
+    return(pd$lag_newlines[idx_paren_close] > 0L || pd$spaces[first_formal_idx - 1L] <= 2L * indent_by)
   }
 
-  pd$lag_newlines[idx_paren_close] > 0L ||
-    pd$spaces[first_formal_idx - 1L] <= 2L * indent_by
+  formals_nl <- which(
+    pd$token %in% c("SYMBOL_FORMALS", "SYMBOL_SUB") &
+      pd$lag_newlines > 0L &
+      row_idx > idx_paren_open &
+      row_idx < idx_paren_close
+  )
+  if (length(formals_nl) == 0L) return(FALSE)
+
+  first_nl_formal <- formals_nl[1L]
+  pd$spaces[first_nl_formal - 1L] <= 2L * indent_by
 }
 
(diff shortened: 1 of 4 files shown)


lorenzwalthert:
Thanks @MichaelChirico. I feel like what's confusing now is the notion of standard / single / double indent vs hanging / open indent. In the [style guide](https://style.tidyverse.org/functions.html#multi-line-function-definitions), there is only ever talked about single and handing indent now it seems. Can you please re-name the rule you modified as well as the helper function and test comment to that effect?

lorenzwalthert:
Also, if you rebase on `main`, you should get the pre-commit hooks passing.

MichaelChirico:
OK, I had a pass at clarifying that, especially in the Rd write-up. pre-commit still stubbornly refuses to pass :)

lorenzwalthert:
Can you have a look at my changes @MichaelChirico ? I don't think we actually needed `SYMBOL_SUB`.

MichaelChirico:
that's right, I wasn't reading it carefully enough myself. At a glance, my brain morphed it to be EQ_FORMALS but that doesn't make sense.

FWIW, we miss cases like function("a"=1) unless styler has already normalized those away by the time we get to this step. I wouldn't stress about it barring a user report.

lorenzwalthert:
Thanks. I guess that won't be a problem.
```
styler:::parse_text('function("a"=1) NULL')
Error:
✖ Styling failed
Caused by error in `parse()`:
! <text>:1:10: unexpected string constant
1: function("a"
             ^

Show Traceback
```

lorenzwalthert (approved):

MichaelChirico:
Nice! I assumed it would work since it does in calls:

```
foo = function("a" = 1) NULL
# Error: unexpected string constant in "foo = function("a""
foo = function(a = 1) NULL
foo("a" = 2)
# NULL
```

Yet another corner case / inconsistency of the parser to mentally earmark 📝

PR #1291, opened 2026-06-09
Compatibility with air + housekeeping

lorenzwalthert:
~There are surprisingly few conflicts.~ Now, people can format this repo with air and {styler}, with the same output. {styler} is far less invasive than air, so formatting with air once means a lot of code is not changed by {styler} on subsequent styling. 
* Bug in air: https://github.com/posit-dev/air/issues/499 caused a re-naming for convenience here. 
* multi-line assignment is handled differently by {styler} than air, this can be circumvented by making sure the first line break in the statement does not occur directly after `<-`.
* {styler} forced no line break after first argument in `switch()`. This was relaxed for compatibility with air.
* We used `fmt: skip` in `switch()` statements to make the pre-commit hook pass.
* ~air does not respect the tidyverse rules which allow to keep unnamed argument on the first line~ This rule seems removed. it turns: 
  ```r
  map(x, f, 
    arg = 1, 
    arg = 2
  )
  ``` 
  Into
  ```
  map(
    x, 
    f, 
    arg = 1, 
    arg = 2
  )
  ```
  Which I think is not optimal.

* Another case is
  ```r
    other_trigger_on_same_line <- (
      pd$token[remaining_row_idx_between_trigger_and_line_break] %in%
        other_trigger_tokens
    )
  ```
  becomes (with air)
  ```r
    other_trigger_on_same_line <- (pd$token[
      remaining_row_idx_between_trigger_and_line_break
    ] %in%
      other_trigger_tokens)
  ```
  Where I also think {styler} styles superior.

* It could not preserve alignment in the blow snippet
  ```r
  style_guide_name <- "styler::tidyverse_style@https://github.com/r-lib"
      create_style_guide(
        # transformer functions
        initialize             = default_style_guide_attributes,
        line_break             =        line_break_manipulators,
        space                  =             space_manipulators,
        token                  =             token_manipulators,
        indention              =         indention_manipulators,
        ...
      )
  ```
* it broke a lot of long lines, which I think is desirable. This increased the line count of the source code.
* Only single lines can be ignored, no ranges: https://github.com/posit-dev/air/issues/456.


MichaelChirico:
@randy3k recently ran a similar comparison internally -- care to share if anything further was found there?

DavisVaughan:
For 

```r
map(x, f, 
  arg = 1, 
  arg = 2
)

# -> 
map(
  x, 
  f, 
  arg = 1, 
  arg = 2
)
```

We really try to limit the forms a function call can be in. There are really only 2:

```r
# "Flat" (all args on same line)
map(x, f, arg = 1, arg = 2)

# "Expanded" (all args broken across lines)
map(
  x, 
  f, 
  arg = 1, 
  arg = 2
)
```

With one exception for trailing anonymous functions, i.e. common `map()` syntax of

```r
# Allowed
map(x, function(x) {
  # do the thing
})
```

This has generally proved to be pretty useful in keeping complexity down, especially across intermingled Air features

This also comes into play with `switch()`, which is "just another function call" to Air and gets no special treatment whatsoever. So there is no support for "first arg on same line" in Air, which we have mostly found to be fine

```r
# air will expand this
switch(x,
  a = 2,
  y = 3
)
```

---

https://github.com/posit-dev/air/issues/199 would fix part of

```r
  other_trigger_on_same_line <- (pd$token[
    remaining_row_idx_between_trigger_and_line_break
  ] %in%
    other_trigger_tokens)
```

---

https://github.com/posit-dev/air/issues/114 would be opt in but would probably help with

```r
style_guide_name <- "styler::tidyverse_style@https://github.com/r-lib"
    create_style_guide(
      # transformer functions
      initialize             = default_style_guide_attributes,
      line_break             =        line_break_manipulators,
      space                  =             space_manipulators,
      token                  =             token_manipulators,
      indention              =         indention_manipulators,
      ...
    )
```

lorenzwalthert:
Thanks for following up on this @DavisVaughan 
* For the _map_ case, I think {styler} still follows the implementation predating https://github.com/tidyverse/style/pull/226. Maybe we should follow up on that at some point, although I still think {styler}'s implementation is better (but I previously clarified that I see {styler} [bound](https://github.com/tidyverse/style/pull/174#issuecomment-909212890) to the style guide, otherwise things become messy). The distinction with named / unnamed arguments indeed added a bit of complexity 😀. 
* As mentioned, the `switch()` special handling gets removed in this very PR.

lorenzwalthert:
Some more interesting cases: 

```r
  mask <- c(
    1L, 2L, if (roxygen[3L] == "\n") 3L, last(which(roxygen == "}"))
  ) %>% sort()

  call(
    1, 2, 3, 4
  )
```
Both are expanded, despite not maxing out the character width limit. In my opinion this consumes too much vertical space and is not optimal: 
```r
mask <- c(
  1L,
  2L,
  if (roxygen[3L] == "\n") 3L,
  last(which(roxygen == "}"))
) %>%
  sort()

call(
  1,
  2,
  3,
  4
)
```
In particular in the second case, one could argue this matches @DavisVaughan's description `# "Flat" (all args on same line)`, so expansion should not happen.

lorenzwalthert:
Another difference is that ignoring happens per line in {styler}, air ignores by expression. Since air lacks ignoring ranges, this comes in handy if the range you want to ignore coincides with a multi-line expression like a function call.  Also, to ignore the subsequent expression, air places the marker above the expression to ignore, {styler} on the same line, and to exclude a range in {styler}, you put the marker above the first line to ignore, while air interprets that marker position as a marker to just ignore one expression. 

I was initially inclined to set the marker text options to include both {styler}'s and airs marker text, but given that they work differently, it would probably just create more confusion.

lorenzwalthert:
Given the open issues + deviances from how {styler} formats the code, I am not willing to accept these changes and introduce Air compatibility here. I merged #1294, which is much smaller in scope instead.

lorenzwalthert:
@MichaelChirico I wondered if we should write up a comparison vignette regarding Air and {styler} (I guess I am quite late to the party but anyways...). My concerns are not only code output differences, but also others, such as speed, support for file types like qmd, Rmd, Rnw, configurability, third-party integrations etc.

